PR #3, opened 2015-08-14
Adding initDeviceAndGetDefaultIME, caps and pushStrings

moizjv:
@Jonahss please review

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -266,5 +266,25 @@ helpers.unlock = async function (adb) {
   });
 };
 
+helpers.initDeviceAndGetDefaultIME = async function (adb, opts) {
+  let defaultIME = null;
+  if (opts.avd !== null) {
+    await helpers.prepareEmulator(adb, opts);
+  }
+  let [deviceId, emPort] = await helpers.getActiveDevice(adb, opts.udid);
+  adb.setDeviceId(deviceId);
+  if (emPort) {
+    adb.setEmulatorPort(emPort);
+  }
+  await adb.waitForDevice();
+  await helpers.ensureDeviceLocale(adb, opts.language, opts.locale);
+  await adb.startLogcat();
+  await adb.forwardPort(opts.systemPort, DEVICE_PORT);
+  await helpers.pushSettings(this.adb);
+  await helpers.pushUnlock(this.adb);
+  if (opts.unicodeKeyboard) {
+    return await helpers.initUnicode(this.adb);
+  }
+}
 
 export default helpers;
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -13,7 +13,6 @@ const DEVICE_PORT = 4724;
 class AndroidDriver extends BaseDriver {
   constructor (opts = {}, shouldValidateCaps = true) {
     super(opts, shouldValidateCaps);
-
     this.desiredConstraints = desiredConstraints;
   }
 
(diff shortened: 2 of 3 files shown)


lib/driver.js
```diff
@@ -83,20 +74,54 @@ class AndroidDriver extends BaseDriver {
...
+            this.opts.language, this.opts.app, stringsTmpDir);
+      this.apkStrings[this.opts.language] = apkStrings;
+      await this.adb.push(jsonFile, remotePath);
+    } catch (err) {
+      if (!(await fs.exists(this.opts.app))) {
+        // delete remote string.json if present
+        await this.adb.rimraf(`${remotePath}/${stringsJson}`);
+      } else {
+        log.warn("Could not get strings, continuing anyway");
+        let remoteFile = remotePath + '/' + stringsJson;
```
Jonahss: can't you use string interpolation for this? then it will be the same as line 117.

lib/driver.js
```diff
@@ -83,20 +74,54 @@ class AndroidDriver extends BaseDriver {
...
   }
+
+  async checkAppPresent () {
+    logger.debug("Checking whether app is actually present");
+    if (!(await fs.exists(this.opts.app))) {
+      logger.errorAndThrow(`Could not find app apk at ${this.opts.app}`);
+    }
+  }
+
+  async pushStrings (language) {
```
Jonahss: Can you add a comment for this function? I'm not really sure how it works, or why it's important.

lib/driver.js
```diff
@@ -22,13 +21,20 @@ class AndroidDriver extends BaseDriver {
...
+                       category: "android.intent.category.LAUNCHER",
+                       flags: "0x10200000",
+                       disableAndroidWatchers: false};
+    _.defaults(this.opts, defaultOpts);
     if (!this.opts.javaVersion) {
       this.opts.javaVersion = await helpers.getJavaVersion();
     }
-
+    // Add condition for chrome driver
+    await this.checkAppPresent();
```
Jonahss: I think there's an issue here.
Here we are checking for the presence of the app file, but at this time, driver.opts.app can be a url. We don't download or unzip apps until the `configureApp` function is called, in this case on line 63.
Maybe you can create an "check valid app cap" function which calls a `checkValidUrl` function if driver.opts.app is a valid url, otherwise call `checkAppPresent()`.

lib/android-helpers.js
```diff
@@ -266,5 +266,25 @@ helpers.unlock = async function (adb) {
   });
 };
 
+helpers.initDeviceAndGetDefaultIME = async function (adb, opts) {
```
Jonahss: I think having the word "and" in a function name is an indicator that the function is doing too much.
What does getting the default IME have to do with initializing the device?

Can we break this into two functions? one that gets the default IME and another that initializes the device?

lib/driver.js
moizjv: yes definitely i spent quiet some time understanding this process

lib/driver.js
moizjv: yeah good point

lib/android-helpers.js
imurchie: :+1:

lib/driver.js
```diff
@@ -13,7 +13,6 @@ const DEVICE_PORT = 4724;
 class AndroidDriver extends BaseDriver {
   constructor (opts = {}, shouldValidateCaps = true) {
     super(opts, shouldValidateCaps);
-
     this.desiredConstraints = desiredConstraints;
```
imurchie: I think this should be `this.desiredCapConstraints = desiredContraints;`

https://github.com/appium/appium-base-driver/blob/master/lib/driver.js#L45

moizjv: :+1:

lib/driver.js
moizjv: i think it would make more sense to call `configureApp` here before we set up lot of things. As if configure app fails that would mean we have malformed url or incorrect zip file or incorrect apk. And hence we can save lot of time that would be other wise wasted in initializing device

PR #4, opened 2015-08-18
MJ merge into master

jlipps:
making this so I can add comments @moizjv @Jonahss

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -150,7 +150,7 @@ helpers.reinstallRemoteApk = async function (adb, localApkPath, pkg,
       logger.warn("Uninstalling remote APK failed, maybe it wasn't installed");
     }
     try {
-      await adb.installRemote(remotePath);
+      await adb.installFromDevicePath(remotePath);
     } catch (e) {
       logger.warn("Installing remote APK failed, going to uninstall and try " +
                   "again");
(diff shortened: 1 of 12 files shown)


lib/android-helpers.js
```diff
@@ -118,7 +119,7 @@ helpers.getLaunchInfoFromManifest = async function (adb, opts) {
   return {appPackage, appWaitPackage, appActivity, appWaitActivity};
 };
 
-helpers.getRemoteApkPath = async function (localApkMd5) {
+helpers.getRemoteApkPath = function (localApkMd5) {
```
jlipps: this needs to be `async`

lib/driver.js
```diff
@@ -1,9 +1,102 @@
...
+    let launchInfo = await helpers.getLaunchInfo(this.adb, this.opts);
+    Object.assign(this.opts, launchInfo);
+
+    // install app
+    await helpers.installApkRemotely(this.adb, this.opts.app, this.opts.appPackage, this.opts.fastReset);
+
+    return launchInfo;
+  }
+
+  async initAndroidBootstrap () {
```
jlipps: do we use this more than once? if not, do we need a 2-line function?

lib/driver.js
```diff
@@ -1,9 +1,102 @@
...
+    this.bootstrap = new Bootstrap(DEVICE_PORT);
+
+    await this.bootstrap.start(this.opts.appPackage, false);
+  }
 
+  async startAUT () {
```
jlipps: do we need a 1-line function? do we use this more than once?

Jonahss: moiz made this function larger i think. lets wait till the end to refactor the flow.

PR #5, opened 2015-08-18
Android driver initialization flow and fixing defects

moizjv:


lib/desired-caps.js
```diff
@@ -1,8 +1,108 @@
-// TODO all the various contraints for Android driver
 const desiredCapConstraints = {
+  platformName: {
+    isString: true,
+    inclusion: ['Android']
```
jlipps: think we also want `presence: true` here

lib/driver.js
```diff
@@ -13,8 +15,8 @@ const DEVICE_PORT = 4724;
 class AndroidDriver extends BaseDriver {
   constructor (opts = {}, shouldValidateCaps = true) {
     super(opts, shouldValidateCaps);
-
-    this.desiredConstraints = desiredConstraints;
+    this.desiredCapConstraints = desiredConstraints;
+    this.settings = new DeviceSettings({ignoreUnimportantViews:false});
```
jlipps: space

lib/driver.js
```diff
@@ -22,29 +24,43 @@ class AndroidDriver extends BaseDriver {
...
+      // download and unzip
+      this.opts.app = await this.helpers.configureApp(this.opts.app, APP_EXTENSION);
+    }
+    await this.checkAppPresent();
     await this.startAndroidSession(this.opts);
-
     return [sessionId, caps];
   }
 
+  async onSettingsUpdate (update) {
```
jlipps: looking at this now, and since we're not using an event emitter, i'm wondering if this should work more like:

```
async onSettingsUpdate (key, value) {
```

lib/driver.js
```diff
@@ -53,33 +69,41 @@ class AndroidDriver extends BaseDriver {
...
+                                     this.opts.locale);
+    await this.adb.startLogcat();
+    if (this.opts.unicodeKeyboard) {
+      this.defaultIME = await helpers.initUnicode(this.adb);
+    }
     await helpers.pushSettingsApp(this.adb);
+    await helpers.pushUnlock(this.adb);
+    // Set CompressedLayoutHierarchy on the device based on current settings object
+    if (this.opts.ignoreUnimportantViews) {
+      this.settings.update({ignoreUnimportantViews: this.opts.ignoreUnimportantViews});
```
jlipps: we don't want to `await` this?

lib/driver.js
```diff
@@ -87,11 +111,62 @@ class AndroidDriver extends BaseDriver {
...
   }
+
+  async checkAppPresent () {
+    log.debug("Checking whether app is actually present");
+    if (!(await fs.exists(this.opts.app))) {
+      log.errorAndThrow(`Could not find app apk at ${this.opts.app}`);
+    }
+  }
+
+  // pushStings method extracts string.xml converts it to string.json and pushes
```
jlipps: pushStrings, also you need an `and` before `converts`

lib/driver.js
```diff
@@ -87,11 +111,62 @@ class AndroidDriver extends BaseDriver {
...
+      this.apkStrings[this.opts.language] = apkStrings;
+      let jsonFile = path.join(localPath, stringsJson);
+      await this.adb.push(jsonFile, remotePath);
+    } catch (err) {
+      if (!(await fs.exists(this.opts.app))) {
+        // delete remote string.json if present
+        await this.adb.rimraf(`${remotePath}/${stringsJson}`);
+      } else {
+        log.warn("Could not get strings, continuing anyway");
+        let remoteFile = remotePath + '/' + stringsJson;
```
jlipps: interpolated string?

lib/driver.js
```diff
@@ -87,11 +111,62 @@ class AndroidDriver extends BaseDriver {
...
+        await this.adb.shell(`echo '{}' > ${remoteFile}`);
```
jlipps: is this how we're supposed to call adb.shell?

lib/driver.js
```diff
@@ -22,29 +24,44 @@ class AndroidDriver extends BaseDriver {
...
+
   async startAndroidSession () {
     log.info(`Starting Android session`);
-
     // set up an instance of ADB
     this.adb = await ADB.createADB();
-
     // set up the device to run on (real or emulator, etc)
-    await this.initDevice();
+    await this.initDevice(this.adb, this.opts);
```
Jonahss: it has 'this'

lib/driver.js
```diff
@@ -87,11 +112,61 @@ class AndroidDriver extends BaseDriver {
...
+        await this.adb.rimraf(`${remotePath}/${stringsJson}`);
+      } else {
+        log.warn("Could not get strings, continuing anyway");
+        let remoteFile = `${remotePath}/${stringsJson}`;
+        await this.adb.shell('echo', [`'{}' > ${remoteFile}`]);
+      }
+    }
+  }
+
+  async setProcessFromManifest () {
```
Jonahss: getPRocess, so we don't set in a function.

Jonahss: do we need it?

lib/driver.js
```diff
@@ -87,11 +112,61 @@ class AndroidDriver extends BaseDriver {
...
+    if (this.opts.app) {
+      let appProcess = await this.adb.processFromManifest(this.opts.app);
+      this.appProcess = appProcess || this.opts.appPackage;
+      log.debug(`Set app process to: ${this.appProcess}`);
+    }
+  }
+
+  // Set CompressedLayoutHierarchy on the device
+  async setCompressedLayoutHierarchy (compress) {
+    await this.proxy(["compressedLayoutHierarchy", {compressLayout: compress}]);
```
Jonahss: this.bootstrap.sendCommand

PR #6, opened 2015-08-18
first stab at android webview support

jlipps:


lib/commands/context.js
```diff
@@ -0,0 +1,174 @@
+import _ from 'lodash';
+import logger from '../logger';
+import webviewHelpers from '../webview-helpers';
+import Chromedriver from 'appium-chromedriver';
+import { errors } from 'mobile-json-wire-protocol';
+import { NATIVE_WIN, WEBVIEW_BASE, WEBVIEW_WIN, CHROMIUM_WIN } from '../webview-helpers';
+
+let commands = {}, helpers = {}, extensions = {};
+
+helpers.defaultContextName = function () {
```
Jonahss: why isn't this just a constant?

lib/webview-helpers.js
```diff
@@ -0,0 +1,111 @@
...
+  logger.debug(`returning process name: ${pkg}`);
+  return pkg;
+}
+
+helpers.getWebviews = async function (adb, deviceSocket) {
+  logger.debug("Getting a list of available webviews");
+  let webviews = await webviewsFromProcs(adb, deviceSocket);
+
+  if (deviceSocket) {
+    return webviews;
```
Jonahss: ?
Can we have a comment for this function, what's the 'deviceSocket' parameter for, and why do we just return here if we supplied it?

lib/commands/context.js
jlipps: because Selendroid implements its own version which returns something different

lib/webview-helpers.js
jlipps: yeah, good idea

lib/commands/context.js
Jonahss: ahh. Maybe a comment to that effect?

jlipps: yep!

lib/commands/context.js
```diff
@@ -0,0 +1,174 @@
...
+helpers.isWebContext = function () {
+  return this.curContext !== null && this.curContext !== NATIVE_WIN;
+};
+
+function isChromedriverContext (viewName) {
+  return viewName.indexOf(WEBVIEW_WIN) !== -1 || viewName === CHROMIUM_WIN;
+}
+
+helpers.startChromedriverProxy = async function (context) {
+  logger.debug("Connecting to chrome-backed webview");
```
imurchie: It seems like `context` is readable (not a random udid or something) so it might be nice to log it, as we do in `setupExistingChromedriver`.

lib/commands/context.js
```diff
@@ -0,0 +1,174 @@
...
+
+helpers.stopChromedriverProxies = async function () {
+  for (let context of _.keys(this.sessionChromedrivers)) {
+    logger.debug(`Stopping chromedriver for context ${context}`);
+    // stop listening for the stopped state event
+    this.sessionChromedrivers[context].removeAllListeners(Chromedriver.EVENT_CHANGED);
+    try {
+      await this.sessionChromedrivers[context].stop();
+    } catch (err) {
+      logger.warn("Error stopping Chromedriver: " + err.message);
```
imurchie: String interpolation.

lib/webview-helpers.js
```diff
@@ -0,0 +1,111 @@
+import _ from 'lodash';
+import logger from './logger';
+import { asyncMap } from 'asyncbox';
+
+const NATIVE_WIN = "NATIVE_APP";
+const WEBVIEW_WIN = "WEBVIEW";
+const WEBVIEW_BASE = WEBVIEW_WIN + "_";
```
imurchie: String interpolation.

lib/webview-helpers.js
```diff
@@ -0,0 +1,111 @@
...
+const CHROMIUM_WIN = "CHROMIUM";
+
+let helpers = {};
+
+// This function gets a list of android system processes and returns ones
+// that look like webviews, with the appropriate webview prefix and their PID
+async function webviewsFromProcs (adb, deviceSocket) {
```
imurchie: It might just be me, but this function, and `procFromWebview`, seem like they ought to be part of `appium-adb`. They seem to expose internals to the Android OS which are out of place here.

jlipps: good point. hmm.

jlipps: @imurchie added some TODOs, moving to adb would require a bit of refactor that I don't have time for at this point, but I agree it should be done in our cleanup phase

jlipps:
@Jonahss @imurchie I cleaned this up, reorganized it, fixed a few bugs, and added more comments. Please re-review!

imurchie:
Looks good to me.

PR #9, opened 2015-08-19
element attribute commands

Jonahss:
8 more passing tests

### DIFF
--- a/lib/commands/element.js
+++ b/lib/commands/element.js
@@ -26,27 +26,31 @@ commands.elementSelected = async function (elementId) {
   let el = this.getElement(elementId);
   return el.isSelected();
 };
-
-commands.setValue = async function (keys, elementId) {
-  let value = keys;
-  if (keys instanceof Array) {
-    value = keys.join("");
-  }
-  let el = this.getElement(elementId);
-  if (el.type !== "MockInputField") {
-    throw new errors.InvalidElementStateError();
-  }
-  el.setAttr('value', value);
-};
 */
+commands.setValue = async function (keys, elementId) {
+  let text = keys;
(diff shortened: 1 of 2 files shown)


lib/commands/element.js
```diff
@@ -46,32 +46,19 @@ commands.getText = async function (elementId) {
 commands.clear = async function (elementId) {
   await this.setValue('', elementId);
 };
-
+*/
 commands.click = async function (elementId) {
```
moizjv: remove async

lib/commands/general.js
```diff
@@ -46,6 +46,10 @@ commands.getScreenshot = async function () {
 
 */
 
+commands.back = async function () {
```
moizjv: no async

lib/commands/element.js
```diff
@@ -26,52 +26,43 @@ commands.elementSelected = async function (elementId) {
...
-    value = keys.join("");
+    text = keys.join("");
   }
-  let el = this.getElement(elementId);
-  if (el.type !== "MockInputField") {
-    throw new errors.InvalidElementStateError();
-  }
-  el.setAttr('value', value);
+
+  let p = {
```
moizjv: p?

jlipps:
I disagree that those `async`s should have been removed. In this case, while it's functionally equivalent not to have `async`, `async` serves as a kind of annotation that this is a function that returns a promise. Otherwise, I need to go and look at some _other_ function to know that I really need to `await` _this_ function if I want to use it correctly.

So my suggestion would be: if we have an async function, decorate it with `async`! It's a form of documentation as well.

PR #10, opened 2015-08-19
Fixing onSettingsUpdate and shutdown

moizjv:


jlipps:
why is my commit in here? it's already in master, right?

lib/driver.js
```diff
@@ -83,6 +84,10 @@ class AndroidDriver extends BaseDriver {
     // start UiAutomator
     this.bootstrap = new Bootstrap(DEVICE_PORT, this.opts.websocket);
     await this.bootstrap.start(this.opts.appPackage, this.opts.disableAndroidWatchers);
+    // handling unexpected shutdown
+    this.bootstrap.onUnexpectedShutdown.catch((err) => {
+      this.startUnexpectedShutdown(err);
```
jlipps: technically this should be `await`ed as part of an async function

lib/driver.js
```diff
@@ -177,11 +182,30 @@ class AndroidDriver extends BaseDriver {
...
   // Set CompressedLayoutHierarchy on the device
   async setCompressedLayoutHierarchy (compress) {
-    await this.bootstrap.sendCommand("compressedLayoutHierarchy", {compressLayout: compress});
+    await this.bootstrap.sendAction("compressedLayoutHierarchy", {compressLayout: compress});
   }
 
-  // TODO: add shutdown flow, unexpected uiautomator exit handling,
-  // call to stopChromedriverProxies, etc...
+  async deleteSession () {
+    if (this.bootstrap) {
```
jlipps: i think this check should just be around the bootstrap shutdown line.

lib/driver.js
jlipps: don't you need to set `this.bootstrap` to null here as well, since it's shut down at this point?

moizjv:
closing this in favor of #11

PR #13, opened 2015-08-20
getLog, isAppInstalled, removeApp, installApp

Jonahss:


lib/commands/general.js
```diff
@@ -82,6 +83,30 @@ commands.getLogTypes = function () {
...
+
+commands.isAppInstalled = function (appPackage) {
+  return this.adb.isAppInstalled(appPackage);
+};
+
+commands.removeApp = function (appPackage) {
+  return this.adb.uninstallApk(appPackage);
+};
+
+commands.installApp = async function (appPath) {
```
moizjv: do we want to check if appPath is valid?

Jonahss: hm. i think we do, but that logic should be extracted somewhere. will work on it

PR #14, opened 2015-08-20
Adding lock,unlock and islocked

moizjv:


lib/commands/actions.js
```diff
@@ -88,6 +90,18 @@ commands.drag = function (startX, startY, endX, endY, duration, touchCount, elem
   }
 };
 
+commands.lock = async function () {
```
Jonahss: I think it should be `return this.adb.lock()`?

lib/commands/actions.js
```diff
@@ -88,6 +90,18 @@ commands.drag = function (startX, startY, endX, endY, duration, touchCount, elem
...
+  await this.adb.lock();
+};
+
+commands.isLocked = function () {
+  return await this.adb.isScreenLocked();
+};
+
+commands.unlock = function () {
```
Jonahss: `helpers` is already defined in this file. maybe call it `androidHelpers`

lib/commands/actions.js
```diff
@@ -88,6 +90,18 @@ commands.drag = function (startX, startY, endX, endY, duration, touchCount, elem
   }
 };
 
+commands.lock = async function () {
+  await this.adb.lock();
+};
+
+commands.isLocked = function () {
```
Jonahss: dont't need to `await` since we should return the promise.

PR #16, opened 2015-08-21
Adding setLocation, pullFile, pushFile, pullFolder and getScreenShot

moizjv:


lib/commands/actions.js
```diff
@@ -102,6 +107,75 @@ commands.unlock = function () {
...
+    await fs.unlink(localFile);
+  }
+};
+
+commands.pullFolder = async function (remotePath) {
+  let localFolder = temp.path({prefix: 'appium'});
+
+  let bufferOnSuccess = function (buffer) {
+    logger.debug("Converting in-memory zip file to base64 encoded string");
+    return buffer.toString('base64');
```
Jonahss: This callback returns the buffer, but the `pullFolder` function returns nothing. You'll have to create a promise and return it.

lib/commands/actions.js
```diff
@@ -102,6 +107,75 @@ commands.unlock = function () {
...
+  await this.adb.pull(remotePath, localFolder);
+  // TODO: find better alternative.
+  let zip = new AdmZip();
+  zip.addLocalFolder(localFolder);
+  zip.toBuffer(bufferOnSuccess, bufferOnFail);
+};
+
+commands.getScreenshot = async function () {
+  let localFile = temp.path({prefix: 'appium', suffix: '.png'});
+  let b64data = '';
```
Jonahss: why declare this here? can just declare on line 174

PR #17, opened 2015-08-24
Commands

Jonahss:


Jonahss:
Do not merge yet.
Seems like somehow my emulator crash more often after this change...... will investigate.

lib/commands/general.js
```diff
@@ -148,6 +148,29 @@ commands.getStrings = async function (language) {
   return this.apkStrings[language];
 };
 
+commands.startActivity = function (appPackage, appActivity) {
```
jlipps: this isn't async?

lib/commands/general.js
```diff
@@ -148,6 +148,29 @@ commands.getStrings = async function (language) {
...
+
+commands.reset = async function () {
+  if (this.opts.fastReset) {
+    log.info("Running fast reset (stop and clear)");
+    return this.adb.stopAndClear(this.opts.appPackage);
+  }
+
+  log.info("Running old fashion reset (reinstall)");
+  await this.adb.uninstallApk(this.opts.appPackage);
+  return androidHelpers.installApkRemotely(this.adb, this.opts.app, this.opts.appPackage, this.opts.fastReset);
```
jlipps: we're doing `return await` now

jlipps:
one comment, then LGTM (pending resolution of your crash comment)

lib/commands/general.js
```diff
@@ -148,6 +148,29 @@ commands.getStrings = async function (language) {
...
+    activity: appActivity,
+    action: this.opts.intentAction,
+    category: this.opts.intentCategory,
+    flags: this.opts.intentFlags,
+    optionalIntentArguments: this.opts.optionalIntentArguments,
+    stopApp: this.opts.stopAppOnReset || !this.opts.dontStopAppOnReset,
+  });
+};
+
+commands.reset = async function () {
```
moizjv: https://github.com/appium/appium/blob/master/lib/appium.js#L380 we need this reset

lib/commands/general.js
```diff
@@ -148,6 +148,32 @@ commands.getStrings = async function (language) {
...
+  if (this.opts.fastReset) {
+    log.info("Running fast reset (stop and clear)");
+    return this.adb.stopAndClear(this.opts.appPackage);
```
moizjv: in this case dont we need to `await this.startAUT()` ?

lib/commands/general.js
```diff
@@ -148,6 +148,32 @@ commands.getStrings = async function (language) {
   return this.apkStrings[language];
 };
 
+commands.startActivity = async function (appPackage, appActivity) {
+  return this.adb.startApp({
```
jlipps: what does this return? if nothing, we don't need it.

also, we need an `await` here

lib/commands/general.js
jlipps: correct; we also need to `await this.adb.stopAndClear`

jlipps: sounds like we should just do away with the `return` here and `await` instead

lib/driver.js
```diff
@@ -43,29 +43,39 @@ class AndroidDriver extends BaseDriver {
...
+                         category: "android.intent.category.LAUNCHER",
+                         flags: "0x10200000",
+                         disableAndroidWatchers: false,
+                         tmpDir: await tempDir.staticDir(),
+                         fullReset: false};
+      _.defaults(this.opts, defaultOpts);
+      if (!this.opts.javaVersion) {
+        this.opts.javaVersion = await helpers.getJavaVersion();
+      }
+      // Add condition for chrome driver
```
jlipps: fyi, this is going to conflict with my PR as well, guess we'll get to see whose is merged first :-)

lib/commands/general.js
Jonahss: ok

lib/driver.js
Jonahss: I'll handle conflicts :)

jlipps: you mean by merging this that I'll handle conflicts I guess :-)

jlipps: conflicts handled!

Jonahss: I meant I'd be fine merging the other PR manually for you :)

jlipps: ah ok thanks :-)

PR #18, opened 2015-08-24
Adding element gesture commands

moizjv:


lib/commands/element.js
```diff
@@ -40,28 +41,57 @@ commands.setValue = function (keys, elementId) {
...
+  return await this.bootstrap.sendAction("element:getLocation", {elementId});
 };
 
-commands.getSize = function (elementId) {
-  return this.bootstrap.sendAction("element:getSize", {elementId});
+commands.getSize = async function (elementId) {
+  return await this.bootstrap.sendAction("element:getSize", {elementId});
 };
+
+commands.touchLongClick = async function (elementId, x, y, duration) {
```
Jonahss: can we name these "params" instead of "opts" to be more consistent with the other commands?

lib/android-helpers.js
```diff
@@ -264,4 +264,12 @@ helpers.unlock = async function (adb) {
   });
 };
 
+helpers.ignoreNullProperties = function (obj) {
```
jlipps: would rather call this `removeNullProperties`. `ignore` isn't really an action.

test/unit/android-helper-specs.js
```diff
@@ -273,4 +274,21 @@ describe('Android Helpers', () => {
...
     });
   }));
+  describe('ignoreNullProperties', () => {
+    it('should ignore null properties', async () => {
+      let test = {foo:null, bar: true};
+      helpers.ignoreNullProperties(test);
+      _.keys(test).length.should.equal(1);
+    });
+    it('should ignore undefined properties', async () => {
+      let test = {foo:undefined, bar: true};
```
jlipps: space after foo

jlipps:
@moizjv 2 small comments then LGTM

PR #19, opened 2015-08-26
add initial chrome browser support

jlipps:
@Jonahss @moizjv here's the initial work. I got a chrome session started, yay! I think there's still some kinks to be worked out, particularly around how we want to handle contexts in chrome sessions (to be honest I'm not sure this works at all in current appium).

but anyway ready for review and for someone to take over!


jlipps:
ok I actually fixed up the way contexts are handled so we're using the same paradigm for both chrome and hybrid. i got a working chrome session yay!

Jonahss:
looks good on first pass. I'll merge and testy test some stuff.

moizjv:
:+1:

jlipps:
@Jonahss sounds good let me know when testy testing is done

jlipps:
@Jonahss rebased off your merge

jlipps:
still awaiting some @Jonahss love

jlipps:
k tested with Chrome and Browser and it's working. going to merge

PR #20, opened 2015-08-27
Adding touch commands

moizjv:
@imurchie @sebv @Jonahss @jlipps please take a look. 

I have tested them with our functional tests and they work fine.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -278,4 +278,12 @@ helpers.removeNullProperties = function (obj) {
   }
 };
 
+helpers.truncateDecimals = function (number, digits) {
+  let multiplier = Math.pow(10, digits),
+      adjustedNum = number * multiplier,
+      truncatedNum = Math[adjustedNum < 0 ? 'ceil' : 'floor'](adjustedNum);
+
+  return truncatedNum / multiplier;
+};
+
 export default helpers;
--- a/lib/commands/element.js
+++ b/lib/commands/element.js
@@ -56,7 +56,7 @@ commands.click = async function (elementId) {
   return await this.bootstrap.sendAction("element:click", {elementId});
 };
 
-commands.getLocation = async function (elementId) {
+commands.getLocationInView = async function (elementId) {
   return await this.bootstrap.sendAction("element:getLocation", {elementId});
 };
 
@@ -92,6 +92,21 @@ commands.complexTap = async function (tapCount, touchCount, duration, x, y) {
   return await this.bootstrap.sendAction("click", {x: x, y: y});
 };
 
+commands.tap = async function (elementId, x = 0, y = 0, count = 1) {
+  for (let i = 0; i < count; i++) {
+    if (elementId) {
+      // we are either tapping on the default location of the element
+      // or an offset from the top left corner
+      if (x !== 0 || y !== 0) {
+        await this.bootstrap.sendAction("element:click", {elementId, x, y});
+      } else {
+        await this.bootstrap.sendAction("element:click", {elementId});
+      }
+    } else {
+      await this.bootstrap.sendAction("click", {x: x, y: y});
+    }
+  }
+};
 /*
 commands.equalsElement = function (el1Id, el2Id) {
   let el1 = this.getElement(el1Id);
--- a/lib/commands/touch.js
+++ b/lib/commands/touch.js
@@ -1,0 +1,282 @@
+import log from '../logger';
+import _ from 'lodash';
+import androidHelpers from '../android-helpers';
+import B from 'bluebird';
+import { errors, isErrorType } from 'mobile-json-wire-protocol';
+import { asyncmap } from 'asyncbox';
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.doTouchAction = async function (action, opts) {
+  switch (action) {
+    case 'tap':
+      return await this.tap(opts.element, opts.x, opts.y, opts.count);
+    case 'press':
+      return await this.touchDown(opts.element, opts.x, opts.y);
+    case 'release':
+      return await this.touchUp(opts.element, opts.x, opts.y);
+    case 'moveTo':
+      return await this.touchMove(opts.element, opts.x, opts.y);
+    case 'wait':
+      return await B.delay(opts.ms);
+    case 'longPress':
+      if (typeof opts.duration === 'undefined' || !opts.duration) {
+        opts.duration = 1000;
+      }
+      return await this.touchLongClick(opts.element, opts.x, opts.y, opts.duration);
+    case 'cancel':
+      // TODO: clarify behavior of 'cancel' action and fix this
+      log.warn("Cancel action currently has no effect");
+      break;
+    default:
+      log.errorAndThrow(`unknown action ${action}`);
+  }
+};
+
+
+// drag is *not* press-move-release, so we need to translate
+// drag works fine for scroll, as well
+commands.doTouchDrag = async function (gestures) {
+  let longPress = gestures[0];
+  let moveTo = gestures[1];
+  let startX = longPress.options.x || 0,
+      startY = longPress.options.y || 0,
+      endX = moveTo.options.x || 0,
+      endY = moveTo.options.y || 0;
+  if (longPress.options.element) {
+    let loc = await this.getLocationInView(longPress.options.element);
+    startX += loc.x || 0;
+    startY += loc.y || 0;
+  }
+  if (moveTo.options.element) {
+    let loc = await this.getLocationInView(moveTo.options.element);
+    endX += loc.x || 0;
+    endY += loc.y || 0;
+  }
+  let apiLevel =  await this.adb.getApiLevel();
+  // lollipop takes a little longer to get things rolling
+  let duration = apiLevel >= 5 ? 2 : 1;
+  // `drag` will take care of whether there is an element or not at that level
+  return await this.drag(startX, startY, endX, endY, duration, 1, longPress.options.element, moveTo.options.element);
+};
+
+// Release gesture needs element or co-ordinates to release it from that position
+// or else release gesture is performed from center of the screen, so to fix it
+// This method sets co-ordinates/element to release gesture if it has no options set already.
+helpers.fixRelease = async function (gestures, release) {
+  // sometimes there are no options
+  release.options = release.options || {};
+  // nothing to do if release options are already set
+  if (release.options.element || (release.options.x && release.options.y)) {
+    return;
+  }
+  // without coordinates, `release` uses the center of the screen, which,
+  // generally speaking, is not what we want
+  // therefore: loop backwards and use the last command with an element and/or
+  // offset coordinates
+  gestures = _.clone(gestures);
+  let ref = null;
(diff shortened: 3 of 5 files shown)


lib/android-helpers.js
```diff
@@ -272,4 +272,12 @@ helpers.removeNullProperties = function (obj) {
   }
 };
 
+helpers.truncateDecimals = function (number, digits) {
```
jlipps: let's be prepared to put this in appium-support if we need it anywhere else

lib/commands/element.js
```diff
@@ -92,6 +92,22 @@ commands.complexTap = async function (tapCount, touchCount, duration, x, y) {
   return await this.bootstrap.sendAction("click", {x: x, y: y});
 };
 
+commands.tap = async function (elementId, x = 0, y = 0, count = 1) {
+
```
jlipps: minor nitpick, but i don't think there's any reason for this blank line

lib/commands/element.js
```diff
@@ -92,6 +92,22 @@ commands.complexTap = async function (tapCount, touchCount, duration, x, y) {
...
+  for(let i=1; i <= count; i++) {
```
jlipps: spacing issues here. also it's more idiomatic to do `i = 0; i < count;`

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+      if (typeof opts.duration === 'undefined' || !opts.duration) {
+        opts.duration = 1000;
+      }
+      return await this.touchLongClick(opts.element, opts.x, opts.y, opts.duration);
+    case 'cancel':
+      // TODO: clarify behavior of 'cancel' action and fix this
+      log.warn("Cancel action currently has no effect");
+      break;
+    default:
+      return `unknown action ${action}`;
```
jlipps: why would we return a string? do we not want to throw an error?

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+    endY += loc.y || 0;
+  }
+  let apiLevel =  await this.adb.getApiLevel();
+  // lollipop takes a little longer to get things rolling
+  let duration = apiLevel >= 5 ? 2 : 1;
+  // `drag` will take care of whether there is an element or not at that level
+  return await this.drag(startX, startY, endX, endY, duration, 1, longPress.options.element, moveTo.options.element);
+};
+
+// Fix last release action
```
jlipps: might want to explain _why_ we need to fix the last release action?

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+helpers.fixRelease = async function (gestures, release) {
+  //  let release = gestures[actions.length - 1];
```
jlipps: what's this line doing

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+  release.options = release.options || {};
+  // nothing to do if release options are already set
+  if (release.options.element || (release.options.x && release.options.y)) {
+    return;
+  }
+  // without coordinates, `release` uses the center of the screen, which,
+  // generally speaking, is not what we want
+  // therefore: loop backwards and use the last command with an element and/or
+  // offset coordinates
+  let ref = _(gestures).chain().initial().filter(function (gesture) {
```
jlipps: can this be implemented using ES6 map/filter rather than underscore?

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+    if (opts.x && opts.y) {
+      release.options = _.pick(opts, 'x', 'y');
+    }
+  }
+  return release;
+};
+
+// Perform one gesture
+helpers.performGesture = async function (gesture) {
+  let res = await this.doTouchAction(gesture.action, gesture.options || {});
```
jlipps: why is this getting a jsonwp response object? i thought only the mjsonwp knows about that? `this.doTouchAction` should be returning the actual result value; the fact that it's not tells me something is wrong somewhere (or this code is not actually doing what you think it does in your testing, but perhaps without bad side effect). This is the same thing I was commenting about on @imurchie's code yesterday.

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+
+    let fixedGestures = await this.parseTouch(gestures, false);
+    for (let g of fixedGestures) {
+      await this.performGesture(g);
+    }
+  }
+};
+
+commands.parseTouch = async function (gestures, multi) {
+  if (multi && _.last(gestures).action === 'release') {
```
jlipps: comment about why we discard the last gesture if it's a release. is it because the automation backend does this for us automatically, or what?

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+        options.y = (gesture.options.y || 0);
+
+        let touchStateObject = {
+          action: gesture.action,
+          options: options,
+          timeOffset: 0.005,
+        };
+        touchStateObjects.push(touchStateObject);
+      }
+    }
```
jlipps: `} else {`

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+          options.x =  pos.x + (size.width / 2);
+          options.y = pos.y + (size.height / 2);
+        }
+        let touchStateObject = {
+          action: gesture.action,
+          options: options,
+          timeOffset: 0.005,
+        };
+        touchStateObjects.push(touchStateObject);
+      }
```
jlipps: `} else {`

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+      };
+      touchStateObjects.push(touchStateObject);
+    }
+  }
+  // we need to change the time (which is now an offset)
+  // and the position (which may be an offset)
+  let prevPos = null,
+      time = 0;
+  for (let state of touchStateObjects) {
+    if (typeof state.options.x === 'undefined' && typeof state.options.y === 'undefined') {
```
jlipps: `_.isUndefined(state.options.x)` etc

lib/commands/touch.js
moizjv: :+1:  good point

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+
+
+commands.performMultiAction = async function (actions, elementId) {
+  // Android needs at least two actions to be able to perform a multi pointer gesture
+  if (actions.length === 1) {
+    throw new Error("Multi Pointer Gestures need at least two actions. " +
+                    "Use Touch Actions for a single action.");
+  }
+
+  let states = [];
```
jlipps: `let states = asyncMap(actions, async (a) => { return await this.parseTouch(action, true); }, false);`

test/unit/touch-specs.js
```diff
@@ -0,0 +1,41 @@
...
+
+/*const should = */chai.should();
+chai.use(chaiAsPromised);
+
+describe('android-controller', function () {
+  describe('#parseTouch', function () {
+    describe('given a touch sequence with absolute coordinates', function () {
+      it('should use offsets for moveTo', async () => {
+        let driver = new AndroidDriver({foo: 'bar'});
+        let actions = [ { action: 'press', options: { x: 100, y: 101 } },
```
jlipps: no need for extra spaces everywhere between brackets and braces

test/unit/touch-specs.js
```diff
@@ -0,0 +1,41 @@
...
+                        { action: 'moveTo', options: { x: 50, y: 51 } },
+                        { action: 'wait', options: { ms: 5000 } },
+                        { action: 'moveTo', options: { x: -40, y: -41 } },
+                        { action: 'release', options: {} } ];
+        let touchStates = await driver.parseTouch(actions, false);
+        touchStates.length.should.equal(5);
+        let parsedActions = [  {action: 'press', x: 100, y: 101},
```
jlipps: alignment

lib/commands/touch.js
imurchie: `await asyncmap(...)`

lib/commands/touch.js
moizjv: yes that is good catch. AndroidBootstrap does not send a JSONWP object, this code would not have worked :+1: copying code curse

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    let fixedGestures = await this.parseTouch(gestures, false);
+    for (let g of fixedGestures) {
+      await this.performGesture(g);
+    }
+  }
+};
+
+commands.parseTouch = async function (gestures, multi) {
```
Jonahss: is this really a command? seems like a 'helper' function, not a command for the driver itself.

Does it need to be async then? Looks sync. If it's sync, you can do a regular .map() on line 253

lib/commands/touch.js
jlipps: yes, it does need to be awaited doesn't it

lib/commands/touch.js
jlipps: @Jonahss it is a helper function however it does need to be `async`, notice that it gets info from the bootstrap

lib/commands/touch.js
```diff
@@ -0,0 +1,274 @@
...
+  if (multi && _.last(gestures).action === 'release') {
+    gestures.pop();
+  }
+
+  let touchStateObjects = [];
+  for (let gesture of gestures) {
```
Jonahss: could this loop be done with a `.map()` function as well?

lib/commands/element.js
```diff
@@ -92,6 +92,21 @@ commands.complexTap = async function (tapCount, touchCount, duration, x, y) {
   return await this.bootstrap.sendAction("click", {x: x, y: y});
 };
 
+commands.tap = async function (elementId, x = 0, y = 0, count = 1) {
+  for (let i=0; i < count; i++) {
```
jlipps: `let i = 0;`

lib/commands/touch.js
```diff
@@ -0,0 +1,282 @@
...
+  return release;
+};
+
+// Perform one gesture
+helpers.performGesture = async function (gesture) {
+  try {
+    return await this.doTouchAction(gesture.action, gesture.options || {});
+  } catch (e) {
+    // sometime the element is not available when releasing, retry without it
+    if (e instanceof errors.NoSuchElementError && gesture.action === 'release' &&
```
jlipps: we should not check like this; instanceof will fail if packages are using different versions of mjsonwp. Instead import the instance checking function from mjsonwp and use that

jlipps: this is an important point that @Jonahss @sebv @imurchie @scottdixon should all be aware of

test/unit/touch-specs.js
```diff
@@ -0,0 +1,41 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import 'mochawait';
+//import * as helpers from '../../lib/android-helpers';
```
jlipps: let's not commit commented code?

lib/commands/touch.js
Jonahss: Good to know, thanks for note.

PR #21, opened 2015-08-28
split out and export common android capability constraints

jlipps:
so selendroid can use them.

ping @Jonahss @moizjv for a quick review

### DIFF
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -1,12 +1,8 @@
-const desiredCapConstraints = {
+let commonCapConstraints = {
   platformName: {
     isString: true,
     inclusion: ['Android'],
     presence: true
-  },
-  browserName: {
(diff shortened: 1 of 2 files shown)


Jonahss:
:+1:

lib/desired-caps.js
```diff
@@ -102,8 +107,13 @@ const desiredCapConstraints = {
...
   chromeOptions: {
-    isString: true
-  }
+    isObject: true
+  },
 };
 
+let desiredCapConstraints = {};
+
+_.merge(desiredCapConstraints, commonCapConstraints, uiautomatorCapConstraints);
```
moizjv: y not `Object.assign` pattern?

jlipps: yeah i can do that, good point

PR #24, opened 2015-09-01
get/set network connections

scottdixon:
@moizjv, @imurchie, @jlipps

### DIFF
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -1,0 +1,55 @@
+import log from '../logger';
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.getNetworkConnection = async function () {
+  log.info('Getting network connection');
+  let airplaneModeOn = await this.adb.isAirplaneModeOn();
+  let connection = airplaneModeOn ? 1 : 0;
+
+  // no need to check anything else if we are in airplane mode
+  if (!airplaneModeOn) {
+    let wifiOn = await this.adb.isWifiOn();
+    connection += (wifiOn ? 2 : 0);
+    let dataOn = await this.adb.isDataOn();
+    connection += (dataOn ? 4 : 0);
+  }
+
+  return connection;
+};
+
+commands.setNetworkConnection = async function (type) {
+  log.info('Setting network connection');
+  // decode the input
+  let airplaneMode = type % 2;
+  type >>= 1;
+  let wifi = type % 2;
+  type >>= 1;
+  let data = type % 2;
+
+  await this.wrapBootstrapDisconnect(async () => {
+    await this.adb.setAirplaneMode(airplaneMode);
+  });
+  await this.wrapBootstrapDisconnect(async () => {
+    await this.adb.broadcastAirplaneMode(airplaneMode);
+  });
+  if (!airplaneMode) {
+    await this.wrapBootstrapDisconnect(async () => {
+      await this.adb.setWifiAndData({wifi, data});
+    });    
+  }
+
+  return await this.getNetworkConnection();
+};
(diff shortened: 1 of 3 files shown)


lib/commands/network.js
```diff
@@ -0,0 +1,50 @@
...
+  type >>= 1;
+  let wifi = type % 2;
+  type >>= 1;
+  let data = type % 2;
+
+  await this.handleADBDisconnectAndPerformAction('setAirplaneMode', airplaneMode);
+  await this.handleADBDisconnectAndPerformAction('broadcastAirplaneMode', airplaneMode);
+
+  if (!airplaneMode){
+    await this.handleADBDisconnectAndPerformAction('setWifiAndData', {wifi, data});
```
scottdixon: Should {wifi, data} be broken into two lines?

imurchie: I would say not.

But there ought to be a space before the `{`.

lib/commands/network.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  return await this.getNetworkConnection();
+};
+
+helpers.handleADBDisconnectAndPerformAction = async function(action, params){
```
imurchie: Space after `function`, and another before `{`

jlipps: spacing should be: `async function (action, params) {`

lib/commands/network.js
```diff
@@ -0,0 +1,50 @@
...
+  this.ignoreUnexpectedShutdown = true;
+  await this.adb[action](params);
+  await this.adb.restart();
+  await this.bootstrap.start(this.opts.appPackage, this.opts.disableAndroidWatchers);
```
jlipps: the fact that the uiautomator bootstrap shuts down makes me think that this function is really `handleUiAutomatorDisconnectAndPerformAction`. The long name also makes me wonder if we should be architecting this differently. What if it was:

```js
// how we call it
  await this.adb.setAirplaneMode(airplaneMode);
});

// how it's defined
helpers.wrapBootstrapDisconnect = async function (wrapped) {
  this.ignoreUnexpectedShutdown = true;
  await wrapped();
  await this.adb.restart();
  await this.bootstrap.start(this.opts.appPackage, this.opts.disableAndroidWatchers);
};
```

This seems a lot cleaner and has the advantage of having clearer names :-)

lib/commands/network.js
```diff
@@ -0,0 +1,55 @@
...
+  type >>= 1;
+  let data = type % 2;
+
+  await this.wrapBootstrapDisconnect(async () => {
+    await this.adb.setAirplaneMode(airplaneMode);
+  });
+  await this.wrapBootstrapDisconnect(async () => {
+    await this.adb.broadcastAirplaneMode(airplaneMode);
+  });
+  if (!airplaneMode){
```
jlipps: I think @imurchie meant that there should be a space before this brace, so it looks like:

```js
if (!airplaneMode) {
```

lib/commands/network.js
```diff
@@ -0,0 +1,55 @@
...
+    await this.wrapBootstrapDisconnect(async () => {
+      await this.adb.setWifiAndData({ wifi, data });
```
jlipps: and then this one should be as you had it:

```js
{wifi, data}
```

the general rule is that JS objects don't have spaces around braces, but functions/control structures/imports/exports do (so as to distinguish them among other things)

jlipps:
also, looks like some rebasing is required. i pushed some stuff so might have gotten in your way. let me know if it's not obvious how to resolve the conflicts

PR #28, opened 2015-09-04
fixing unlock command

moizjv:


moizjv:
merging it as it is minor change

PR #35, opened 2015-09-09
1.5 cleanup: style fixes

scottdixon:


scottdixon:
@moizjv @imurchie

package.json
```diff
@@ -29,17 +29,17 @@
   },
   "dependencies": {
     "adm-zip": "^0.4.7",
-    "appium-adb": "2.0.0-beta14",
+    "appium-adb": "2.0.0-beta15",
```
imurchie: We ought to be able to use `beta` and `latest` for our packages, while they are up in the air. The packages need to be published with the tags, though (at least, the beta ones do).

package.json
```diff
@@ -29,17 +29,17 @@
...
     "appium-base-driver": "^0.4.1",
     "appium-chromedriver": "^2.3.0",
-    "appium-express": "^0.0.5",
+    "appium-express": "^0.0.6",
     "appium-logger": "^1.1.7",
     "appium-support": "2.0.0-beta21",
     "appium-unlock": "0.0.1",
     "asyncbox": "^2.0.4",
-    "babel-runtime": "=5.5.5",
+    "babel-runtime": "5.8.20",
```
imurchie: Is this the one we want to use? If so, we also need `=` here or we sometimes run into problems with babel

lib/android-helpers.js
```diff
@@ -75,7 +75,7 @@ helpers.ensureDeviceLocale = async function (adb, language, locale) {
 };
 
 helpers.getActiveDevice = async function (adb, udid) {
-  logger.info('Retrieving device list');
+  logger.info("Retrieving device list");
```
imurchie: Why the single quote to double?

scottdixon: The code was a mixture of both, predominately doubles. I can make all singles?

imurchie: Sure. But going forward this is pretty low priority. :)

package.json
sebv: Yeah anything:
-beta -> change it to: beta
0.x.x -> change it  to latest or upgrade the deps to 1.0.0 (SemVer is not enable for 0.x.x packages.)

scottdixon:
Reverted package.json changes until packages are tagged correctly. Only changes in this PR are style fixes.

PR #43, opened 2015-09-18
make sure to start chromedrivers with reference to the correct adb server port

jlipps:
to avoid regressing from current appium behavior.

cc @aluedeke

ping @moizjv @Jonahss for review


Jonahss:
looks good to me.

aluedeke:
nice thx @jlipps

PR #53, opened 2015-10-12
Defaulting findElorEls to single element

scottdixon:
@imurchie on second thoughts, I think the benefit from defaulting to false (i.e. requesting a single element) will be outweighed by potential confusion for the user.

The old API had `elementByClassName` and `elementsByClassName`. If these are removed, it might make sense to train the user from the beginning to specify single or multiple elements.

I'd be curious to get your thoughts.


jlipps:
users won't be using this API directly so I don't think it matters much. they'll be hitting the session via an actual client which this change doesn't affect.

imurchie:
I agree. If it makes the test code feel cleaner, I'm fine with the change.

PR #59, opened 2015-11-20
add test to check if java version parsing works for openjdk

christian-bromann:
Original PR: appium/appium#5532
Helper method is already fully ported.

@imurchie is this test even valid anymore?


imurchie:
Closing. #61 fixed the problem with the build, and adds the test.

PR #64, opened 2015-12-01
getStrings fix, setting ADB java version to 2.0

scottdixon:
@imurchie

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -82,7 +82,7 @@ class AndroidDriver extends BaseDriver {
       }
 
       // set up an instance of ADB
-      this.adb = await ADB.createADB();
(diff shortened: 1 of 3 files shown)


lib/driver.js
```diff
@@ -82,7 +82,7 @@ class AndroidDriver extends BaseDriver {
       }
 
       // set up an instance of ADB
-      this.adb = await ADB.createADB();
+      this.adb = await ADB.createADB({javaVersion: 2.0});
```
imurchie: Should this not be `this.opts.javaVersion`?

scottdixon: Good catch. Updated.

imurchie:
LGTM

jlipps:
:+1:

PR #66, opened 2015-12-01
allow user to provide app package under 'app' capability

scottdixon:
Backward compatibility - our docs don't suggest that we support this feature but we do in current Appium and our existing e2e tests rely on it.

@imurchie

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -84,6 +84,12 @@ class AndroidDriver extends BaseDriver {
       // set up an instance of ADB
       this.adb = await ADB.createADB({javaVersion: this.opts.javaVersion});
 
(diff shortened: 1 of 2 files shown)


lib/driver.js
```diff
@@ -84,6 +84,12 @@ class AndroidDriver extends BaseDriver {
       // set up an instance of ADB
       this.adb = await ADB.createADB({javaVersion: this.opts.javaVersion});
 
+      if (this.helpers.isPackageOrBundle(this.opts.app)){
```
imurchie: Does `this.helpers.isPackageOrBundle` already exist?

scottdixon: yep

imurchie: Cool. Looks good to me, then.

PR #67, opened 2015-12-02
Sanitize orientation

imurchie:
Also make sure all the commands are async.


imurchie:
Port of https://github.com/appium/appium/pull/5890

PR #72, opened 2015-12-15
getSession fix: return actual capabilities

scottdixon:
## getSession():

With this change we return the same as 1.4.

### 1.4 returns:
image

### 1.5 was returning:
image

@imurchie


scottdixon:
Whoops. Adding tests.

scottdixon:
Please review @imurchie

imurchie:
LGTM

PR #74, opened 2015-12-17
add setUrl command

jlipps:
calls adb startUri under the hood so users can deep link to a dynamic place in their AUT.

also change the name of the commands tests in the functional directory since it shouldn't be tied to which app we are using to test the commands with.

review please @imurchie @scottdixon


scottdixon:
Nice one :+1:

imurchie:
+1 for the test renaming.

imurchie:
For `setUrl`, is the app already launched? And they are just going to an activity through a uri?

jlipps:
@imurchie `setUrl` is a way of providing access to `startUri` for the AUT in question. we could call it with the contacts app for example by doing:

```
await driver.get("content://contacts/people");
```

in my testcase it just so happens that the data uri i'm using is an actual url, and the app i'm automating is a web browser. does that make sense?

technically, the app need not already be launched but i envision people doing this as part of the startup flow then using `setUrl` to navigate somewhere specific in their app.

this gets around the necessity of adding startUri to MJSONWP, android-driver, and the clients (since we can reuse, in a semantically decent way, setUrl)

imurchie:
Ok. That was my understanding. My only hesitation is that if it _does_ start the app the semantic decency begins to degrade. But as it is, thumbs up!

jlipps:
Yes good point. I try to maintain the semantic decency by not allowing them to specify an app package different than the AUT. so since they already have a session, the app is already launched! the putative start_uri endpoint would not have this restriction (but hopefully we don't need it in practice)

jlipps:
published as 1.3.0

PR #75, opened 2015-12-22
return merged server + desired caps on createSession

scottdixon:
Resolves https://github.com/appium/appium/issues/5946


scottdixon:
@jlipps - please confirm createSession() should return the same value as getSession(), not just the user's desired caps.

jlipps:
@scottdixon yeah i believe it should!

scottdixon:
published as v1.3.2

PR #76, opened 2016-01-04
getLocationInView/getLocation mapping

scottdixon:
- Replaced getLocationInView with getLocation.
- Mapped getLocationInView to getLocation.

(these changes are consistent with iOS driver - https://github.com/appium/appium-ios-driver/blob/c84609bb7e50c20fbd602e080d1536d762568313/lib/commands/general.js#L58-L60)

Resolves confusion such as https://github.com/appium/appium/issues/5960

### DIFF
--- a/lib/commands/element.js
+++ b/lib/commands/element.js
@@ -54,8 +54,12 @@ commands.click = async function (elementId) {
   return await this.bootstrap.sendAction("element:click", {elementId});
 };
 
-commands.getLocationInView = async function (elementId) {
+commands.getLocation = async function (elementId) {
   return await this.bootstrap.sendAction("element:getLocation", {elementId});
+};
+
+commands.getLocationInView = async function (elementId) {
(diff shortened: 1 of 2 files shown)


scottdixon:
Please review @imurchie

lib/commands/general.js
```diff
@@ -64,6 +64,10 @@ commands.getLogTypes = function () {
   return _.keys(logTypesSupported);
 };
 
+commands.getLocationInView = async function (element) {
```
imurchie: Why is this in `general.js` while the other is in `element.js`? I would prefer both of them in `element.js`, which is the logical place for them.

Also, for consistency's sake it should be `elementId`.

scottdixon:
@imurchie - updated.

Also updated ios-driver https://github.com/appium/appium-ios-driver/pull/58

imurchie:
LGTM

PR #77, opened 2016-01-05
adjust ADB setup. fixes #5949

Jonahss:


lib/android-helpers.js
```diff
@@ -76,23 +77,41 @@ helpers.ensureDeviceLocale = async function (adb, language, locale) {
...
+    udid = devices[0].udid;
+    emPort = adb.getPortFromEmulatorString(udid);
+  }
+  logger.info(`Using device: ${udid}`);
+  return {udid, emPort};
+};
+
+// returns a new adb instance with deviceId set
+helpers.createADB = async function (javaVersion, udid, emPort) {
+  let adb = await ADB.createADB({ javaVersion });
```
jlipps: we don't use spaces in regular object like this

lib/driver.js
```diff
@@ -92,8 +91,13 @@ class AndroidDriver extends BaseDriver {
         log.info(`Chrome-type package and activity are ${pkg} and ${activity}`);
       }
 
+      // get device udid for this session
+      let { udid, emPort } = await helpers.getDeviceInfoFromCaps(this.opts);
```
jlipps: same here

jlipps:
couple style comments. otherwise LGTM and \o/ for tests

lib/android-helpers.js
Jonahss: oh gothca. Was wondering if we treated objects differently when we used the "key is same as value" shorthand.

Jonahss: can we have this added to our linters?

jlipps:
@Jonahss i'm not aware of a lint setting for this style rule

jlipps:
LGTM now

PR #78, opened 2016-01-08
added a new command for getting device date and time

SrinivasanTarget:
@Jonahss Do we have any method existing to get current device id for android? Also please review.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -25,6 +25,16 @@ commands.keys = async function (keys) {
   return await this.bootstrap.sendAction('setText', params);
 };
 
+commands.getDeviceTime = async function() {
+  log.info('Attempting to capture android device date and time');
+  try {
+    let out = await this.adb.shell(['-s', this.opts.udid], 'date');
+    return out.trim();
+  } catch (err) {
+    log.warn('Could not capture device date and time');
+  }
+};
+
 commands.getPageSource = function () {
   return this.bootstrap.sendAction('source');
 };
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ console.log(await driver.getOrientation()); // -> 'LANDSCAPE'
 | `getContexts`              |
 | `getCurrentActivity`       |
 | `getCurrentContext`        |
+| `getDeviceTime`            |
 | `getLocationInView`        |
 | `getLog`                   |
 | `getLogTypes`              |


Jonahss:
Ah, you can use `this.adb`. It's a reference to an ADB object from [appium-adb](https://github.com/appium/appium-adb). By the time the session is created, it already has the udid set on the adb object and it adds the udid arguments automatically.

SrinivasanTarget:
@Jonahss Is this looks fine?

lib/commands/general.js
```diff
@@ -4,6 +4,7 @@ import { fs } from 'appium-support';
 import B from 'bluebird';
 //import { errors } from 'mobile-json-wire-protocol';
 import log from '../logger';
+import { exec } from 'teen_process';
```
imurchie: No need for this.

lib/commands/general.js
```diff
@@ -25,6 +26,16 @@ commands.keys = async function (keys) {
   return await this.bootstrap.sendAction('setText', params);
 };
 
+commands.getDeviceTime = async function() {
+  log.info('Attempting to capture android device date and time');
+  try {
+    await this.adb.shell(['-s', this.opts.udid],'date')
```
imurchie: Need to get the standard output from this call, to use in the next line. Something like we find here https://github.com/appium/appium-android-driver/blob/6e6478c6acab054b1966fdd2090412090726f3ff/lib/webview-helpers.js#L20

imurchie: Also need a semicolon at the end.

imurchie:
A few comments to address the CI failure.

lib/commands/general.js
```diff
@@ -25,6 +25,16 @@ commands.keys = async function (keys) {
   return await this.bootstrap.sendAction('setText', params);
 };
 
+commands.getDeviceTime = async function() {
+  log.info('Attempting to capture android device date and time');
+  try {
+    let out = await this.adb.shell(['-s', this.opts.udid],'date');
```
imurchie: Can you add a space before `'date'`? We are then on-style.

imurchie:
One more quick change, then squash and we can commit! Thanks for this!

SrinivasanTarget:
Yup Its squashed @imurchie

imurchie:
Thanks for this!

PR #83, opened 2016-01-15
ported PR #6007 - Android 6 procFromWebview fix

scottdixon:
Ports PR https://github.com/appium/appium/pull/6007 to 1.5 which fixes issue https://github.com/appium/appium/issues/5689

cc) @imurchie


imurchie:
Looks good to me!

PR #89, opened 2016-02-01
Add default export

imurchie:
So we don't need to do the whole import thing, if we just want the driver.


imurchie:
Expand to include fixing e2e tests after turning CI back on.

lib/commands/find.js
```diff
@@ -58,9 +60,16 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
     }
   }
 
-  return element;
+  if (mult) {
```
jlipps: should we double-check that ios has this logic? and really it seems like common logic we could extend from basedriver somehow. but that can wait for another time.

imurchie: It does. I basically copied this from there. It should definitely be abstracted out (perhaps having the actual drivers just implement the actual find code, and the base driver doing the call implicit wait magics).

PR #90, opened 2016-02-02
Add optional arguments to startActivity

imurchie:
There is much more that can be sent to the function. See https://github.com/appium/appium/issues/6097


jlipps:
:+1:

TikhomirovSergey:
:+1:

PR #92, opened 2016-02-04
Fix replaceValue method

imurchie:
Parameters were switched around.

Fixes https://github.com/appium/appium/issues/6101


jlipps:
:+1:

PR #93, opened 2016-02-08
Add tests for startActivity with dontStopAppOnReset

imurchie:
Addresses https://github.com/appium/appium/issues/6114


imurchie:
All of the attempts to use the intents, etc., from the capabilities failed. Remove.

PR #95, opened 2016-02-09
setting the mock location on the device and android emulator.

anandsadu:
#### Fix: ####
The code fix addresses the issue [https://github.com/appium/appium/issues/4309] on **appium 1.5** to set the Location on the Android Devices and Android Emulators.It is linked with the [appium/io.appium.settings#6] .This feature will work once the settings apk is published to npm.

#### Commit summary: ####

set Location uses Android Service (in the settings apk )is invoked in the case of device.
Need to enable **Allow mock locations** option in the Developer Options in Device/Emulator

### DIFF
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -76,13 +76,10 @@ commands.toggleFlightMode = async function () {
   });
 };
 
-commands.setGeoLocation = async function (location) {
-  /*
-   * TODO: Implement isRealDevice(). This method fails on
-   * real devices, it should throw a NotYetImplementedError
-   */
-  let cmd = `geo fix ${location.longitude} ${location.latitude}`;
-  return await this.adb.sendTelnetCommand(cmd);
+commands.setGeoLocation = async function (location) {    
+   let cmd = ['am', 'startservice', '-e', 'longitude', location.longitude,
+              '-e', 'latitude', location.latitude, 'io.appium.settings/.LocationService'];
+   return await this.adb.shell(cmd);
 };
 
 commands.toggleLocationServices = async function () {
--- a/test/functional/commands/geo-location-e2e-specs.js
+++ b/test/functional/commands/geo-location-e2e-specs.js
@@ -1,0 +1,53 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import { retryInterval } from 'asyncbox';
+import { AndroidDriver } from '../../../..';
+import sampleApps from 'sample-apps';
+
+
+chai.should();
+chai.use(chaiAsPromised);
+
+let driver;
+let caps = {
+  app: sampleApps('gpsDemo-debug'),
+  deviceName: 'Android',
+  platformName: 'Android'
+};
+
+describe.skip("geo-location", function () {
+  before(async () => {
+    driver = new AndroidDriver();
+    await driver.createSession(caps);
+  });
+  after(async () => {
+    await driver.deleteSession();
+  });
+
+  it('should set geo location @skip-ci', async () => {
+    let getText = async () => {
+      let els = await driver.findElOrEls('class name', 'android.widget.TextView', true);
+      return await driver.getText(els[1].ELEMENT);
+    };
+
+    let latitude = '27.17';
+    let longitude = '78.04';
+
+    let text = await getText();
+    text.should.not.include(`Latitude: ${latitude}`);
+    text.should.not.include(`Longitude: ${longitude}`);
+
+    await driver.setGeoLocation({latitude, longitude});
+
+    // wait for the text to change
+    await retryInterval(6, 1000, async () => {
+      if (await getText() === 'GPS Tutorial') {
+        throw new Error('Location not set yet. Retry.');
+      }
+    });
+
+    text = await getText();
+    text.should.include(`Latitude: ${latitude}`);
+    text.should.include(`Longitude: ${longitude}`);
+  });
+});
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "asyncbox": "^2.0.4",
     "babel-runtime": "=5.8.24",
     "bluebird": "^2.9.32",
-    "io.appium.settings": "^2.0.0",
+    "io.appium.settings": "^2.1.0",
     "lodash": "^3.10.0",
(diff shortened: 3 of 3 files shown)


imurchie:
`io.appium.settings` v2.1.0 published with the required changes.

lib/commands/network.js
```diff
@@ -77,12 +77,8 @@ commands.toggleFlightMode = async function () {
 };
 
 commands.setGeoLocation = async function (location) {
-  /*
-   * TODO: Implement isRealDevice(). This method fails on
-   * real devices, it should throw a NotYetImplementedError
-   */
-  let cmd = `geo fix ${location.longitude} ${location.latitude}`;
-  return await this.adb.sendTelnetCommand(cmd);
+  let cmd = 'am startservice -e longitude ' + '"' + location.longitude + '"' + ' -e latitude ' + '"' + location.latitude + '"' + ' io.appium.settings/.LocationService';
```
imurchie: I don't think this is how `adb.shell` works anymore. It should be taking an array. See https://github.com/appium/appium-android-driver/blob/master/lib/commands/general.js#L59-L63

anandsadu: Can you please review the changes.

lib/commands/network.js
```diff
@@ -76,13 +76,9 @@ commands.toggleFlightMode = async function () {
...
 };
 
-commands.setGeoLocation = async function (location) {
-  /*
-   * TODO: Implement isRealDevice(). This method fails on
-   * real devices, it should throw a NotYetImplementedError
-   */
-  let cmd = `geo fix ${location.longitude} ${location.latitude}`;
-  return await this.adb.sendTelnetCommand(cmd);
+commands.setGeoLocation = async function(location) {
```
imurchie: A space is needed after `function` and before the opening parenthesis (see https://github.com/appium/appium/blob/1.5/docs/en/contributing-to-appium/style-guide-2.0.md#style-notes).

imurchie:
One small style change, then it looks good!

lib/commands/network.js
```diff
@@ -76,13 +76,9 @@ commands.toggleFlightMode = async function () {
...
 
-commands.setGeoLocation = async function (location) {
-  /*
-   * TODO: Implement isRealDevice(). This method fails on
-   * real devices, it should throw a NotYetImplementedError
-   */
-  let cmd = `geo fix ${location.longitude} ${location.latitude}`;
-  return await this.adb.sendTelnetCommand(cmd);
+commands.setGeoLocation = async function(location) {    
+    let cmd = ['am', 'startservice', '-e longitude', location.longitude, '-e latitude', location.latitude, 'io.appium.settings/.LocationService'];
```
imurchie: I spoke too soon. This needs to separate out all the elements. So `-e` and `longitude` need to be separated, for instance.

imurchie:
```js
let cmd = ['am', 'startservice', '-e', 'longitude', location.longitude, '-e', 'latitude', location.latitude, 'io.appium.settings/.LocationService'];
```

lib/commands/network.js
```diff
@@ -76,13 +76,9 @@ commands.toggleFlightMode = async function () {
...
+    return await this.adb.shell(cmd);
```
imurchie: Can you also bring the indent back to 2 spaces?

imurchie:
Two other things to do:

1. Update the version of `io.appium.settings` in `package.json` to 2.1.0, so the settings app has everything that is needed.

2. Add a test. I've taken the liberty to write one, which you can put into this PR: https://gist.github.com/imurchie/6fdfa7b90717916e23d1

anandsadu:
@imurchie Thanks for your review. I have updated the commit.

test/functional/commands/geo-location-e2e-specs.js
```diff
@@ -0,0 +1,53 @@
...
+    let longitude = '78.04';
+
+    let text = await getText();
+    text.should.not.include(`Latitude: ${latitude}`);
+    text.should.not.include(`Longitude: ${longitude}`);
+
+    await driver.setGeoLocation({latitude, longitude});
+
+    // wait for the text to change
+    await retryInterval(3, 1000, async () => {
```
imurchie: The failure in CI seems to be because we don't retry this enough times. Try bumping from `3` to `6` and rerunning. If it still fails, we can regroup.

lib/commands/network.js
```diff
@@ -76,13 +76,9 @@ commands.toggleFlightMode = async function () {
...
 
-commands.setGeoLocation = async function (location) {
-  /*
-   * TODO: Implement isRealDevice(). This method fails on
-   * real devices, it should throw a NotYetImplementedError
-   */
-  let cmd = `geo fix ${location.longitude} ${location.latitude}`;
-  return await this.adb.sendTelnetCommand(cmd);
+commands.setGeoLocation = async function (location) {    
+  let cmd = ['am', 'startservice', '-e', 'longitude', location.longitude, '-e', 'latitude', location.latitude, 'io.appium.settings/.LocationService'];
```
jlipps: this should really be broken up into separate lines according to our style guide so that it's not longer than 80 chars:

```js
let cmd = [foo, bar, baz,  // <-- no more than 80 chars
           abc, qed, xyz];
```

anandsadu:
@jlipps @imurchie 

Ideally to execute this test we need to enable **Allow mock locations** option in the Developer Options. Do you think we have to check this option in the Device/Emulator against which the test is getting executing ?

imurchie:
Ah! Hadn't thought of that.

imurchie:
I've spent a bit trying to automate turning on mock location privileges on the CI machines, to no avail. I will try to get the image changed so it is just on.

In the meantime, I think you ought to just mark this to skip (i.e., `describe.skip("geo-location", function () {`) and we'll revisit when the CI has been fixed.

jlipps:
@imurchie or we could put "@skip-ci" in the text as we have done in other places

imurchie:
I don't think `gulp e2e-test` respects `@skip-ci`.

jlipps:
ahhh, good point

rompic:
+1

aluedeke:
@imurchie can we proceed with this?

imurchie:
Revisiting.

PR #101, opened 2016-02-10
Attempt to de-flakify the network setting tests

imurchie:
They run fine locally. Trying to get them to work reliably in CI.

### DIFF
--- a/test/functional/commands/network-e2e-specs.js
+++ b/test/functional/commands/network-e2e-specs.js
@@ -1,42 +1,52 @@
-// This currently does not work reliably in CI
-// Further, our CI does not respect skip or @skip-ci
-// investigate and reinstate
-
-// import chai from 'chai';
-// import chaiAsPromised from 'chai-as-promised';
-// import { AndroidDriver } from '../../../..';
-// import sampleApps from 'sample-apps';
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import { AndroidDriver } from '../../../..';
+import sampleApps from 'sample-apps';
 
 
-// chai.should();
-// chai.use(chaiAsPromised);
+chai.should();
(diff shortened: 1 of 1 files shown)


test/functional/commands/network-e2e-specs.js
```diff
@@ -1,42 +1,52 @@
...
+    try {
+      await driver.deleteSession();
+    } catch (err) {
+      console.log('Unable to delete session. Continueing anyway.');
+    }
+  });
+  describe('setNetworkConnection @skip-ci', () => {
+    it('should be able to set to 1', async () => {
+      await driver.setNetworkConnection(1);
+      await driver.getNetworkConnection().should.eventually.equal(1);
```
scottdixon: setNetworkConnection calls getNetworkConnection when it's done. You could avoid the additional call. Although, I like the explicit nature of the test.

imurchie: Yeah. The [spec](https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#device-modes) says it needs to return the connection type.

But the flakiness is in the setting (which means lots of adb fails and restarts, along with restarting the device).

imurchie: Yeah. The [spec](https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#device-modes) says it needs to return the connection type.

But the flakiness is in the setting (which means lots of adb fails and restarts, along with restarting the device).

imurchie: Yeah. The [spec](https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#device-modes) says it needs to return the connection type.

But the flakiness is in the setting (which means lots of adb fails and restarts, along with restarting the device).

PR #103, opened 2016-02-12
Udpate chromedriver

imurchie:
Move to 2.21.


jlipps:
:+1: want to publish a patch version so we get it into next appium?

imurchie:
Sure!

PR #104, opened 2016-02-13
removed cancel touch action

SrinivasanTarget:
@imurchie please review this.


imurchie:
The `cancel` action is still valid, so throwing an error is the wrong behavior. Logging it and doing nothing is what we want to do at this time.

SrinivasanTarget:
Thanks @imurchie for your clarification.

PR #106, opened 2016-02-17
Keep track of ignoreUnexpectedShutdown in Bootstrap

scottdixon:
This way Bootstrap can access `ignoreUnexpectedShutdown` when it errors on Windows.

cc) @imurchie

### DIFF
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -162,7 +162,6 @@ helpers.doKey = async function (key) {
 };
 
 helpers.wrapBootstrapDisconnect = async function (wrapped) {
-  log.debug('Ignoring bootstrap disconnect while operation completes');
   this.bootstrap.ignoreUnexpectedShutdown = true;
   try {
     await wrapped();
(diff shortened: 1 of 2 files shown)


lib/commands/network.js
```diff
@@ -163,13 +163,13 @@ helpers.doKey = async function (key) {
 
 helpers.wrapBootstrapDisconnect = async function (wrapped) {
   log.debug('Ignoring bootstrap disconnect while operation completes');
```
imurchie: Perhaps this bit of logging ought to be in bootstrap as well?

lib/commands/network.js
```diff
@@ -163,13 +163,13 @@ helpers.doKey = async function (key) {
...
-  this.ignoreUnexpectedShutdown = true;
+  this.bootstrap.ignoreUnexpectedShutdown = true;
   try {
     await wrapped();
     await this.adb.restart();
     await this.bootstrap.start(this.opts.appPackage, this.opts.disableAndroidWatchers, this.acceptSslCerts);
   } finally {
-    this.ignoreUnexpectedShutdown = false;
+    this.bootstrap.ignoreUnexpectedShutdown = false;
     log.debug('Operation complete. Re-watching for bootstrap disconnect');
```
imurchie: Ditto.

imurchie:
I like this!

PR #109, opened 2016-02-18
use androidAdbPort cap to specify a custom adb port

jlipps:


imurchie:
Is the `android` in `androidAdbPort` not redundant?

jlipps:
updated!

lib/driver.js
```diff
@@ -11,6 +11,7 @@ import { fs, tempDir } from 'appium-support';
 
 const APP_EXTENSION = '.apk';
 const DEVICE_PORT = 4724;
+const DEFAULT_ADB_PORT = 5037;
```
imurchie: Should this be exposed from `appium-adb` and just used here? Seems like info only it should know.

jlipps:
ran the failing functional locally and it passed, going ahead and merging

jlipps:
published as 1.6.12

jlipps:
i mean, 1.6.11 not counting the fat finger

PR #111, opened 2016-02-23
Enable support for xpath find commands on elements

guysmoilov:
Will work in combination with: https://github.com/appium/appium-android-bootstrap/pull/17


imurchie:
Looks good. I'll merge this and then update `package.json` to make sure that the version of `appium-android-bootstrap` is high enough to support the context.

imurchie:
Thanks for the work that went into this!

PR #112, opened 2016-02-24
Changed type of keystorePassword from Boolean to String

arudenko:
I changed type of keystorePassword from Boolean to String


imurchie:
My apologies. This change has been made elsewhere. Thanks for the work, though!

PR #118, opened 2016-03-08
Fix ensureDeviceLocale to work with API 23

rgonalo:
Android 6.0 (API 23) uses 'persist.sys.locale' instead of 'persist.sys.language' and 'persist.sys.country' properties to store device language settings.
I have modified ensureDeviceLocale method to use setDeviceLanguage/Country or setDeviceLocale, depending on the API level.

This PR requires the new methods added in this appium-adb PR: https://github.com/appium/appium-adb/pull/143

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -55,22 +55,39 @@ helpers.prepareEmulator = async function (adb, opts) {
                       avdReadyTimeout);
 };
 
-helpers.ensureDeviceLocale = async function (adb, language, locale) {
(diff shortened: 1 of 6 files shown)


imurchie:
I published your changes in `appium-adb` 2.5.0, and the tests here pass now.

Can you update `package.json` to make the minimum version of `appium-adb` be 2.5.0, and re-push? 

Thanks for this work!

rgonalo:
Updated!

imurchie:
Nice! Thanks!

PR #120, opened 2016-03-09
Launch AVD before getting devices

rgonalo:
In Appium 1.5, getDeviceInfoFromCaps() fails when avd capability is configured and no device is connected.
I have moved prepareEmulator call from initDevice to getDeviceInfoFromCaps to ensure that the emulator is up before trying to access it.

This PR and https://github.com/appium/appium-adb/pull/145 fix the issue https://github.com/appium/appium/issues/6233


imurchie:
appium/appium-adb#145 published in 2.5.1.

Thanks for this. Looks good.

PR #121, opened 2016-03-10
Fix the android reset command https://github.com/appium/appium/issues…

JadeDu:
As the fastReset option is not available, changing the if check to use fullReset.


imurchie:
Published in 1.7.2.

PR #139, opened 2016-04-28
Remove restrictions on browserName and combining with app capability for Grid

jessehudl:
This is the android implementation of appium/appium-ios-driver#127, which has already been merged.

#19 looks to have added these particular restrictions. `browserName` needs to be open since that's how the Selenium Grid matches nodes. See appium/appium-ios-driver#127 for more details.

This PR also had to remove the specific check for providing `browserName` and `app` together, which was not necessary for the iOS PR. I'm not positive what the rationale for this was, I'm assuming to try and catch mistakenly providing both. However providing both still prefers `app` with a custom `browserName` (which is necessary for the Grid). Another alternative would be to check if this server is being run as a Grid node, to do this conditionally. Perhaps @jlipps can comment since he added these changes?

I've left `CHROME_BROWSERS` in `android-helpers.js` since it appears to be used by other things.

I've tested these changes with Appium v1.5 locally and with Selenium Grid 2.52.0.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -317,10 +317,11 @@ class AndroidDriver extends BaseDriver {
       let msg = 'The desired capabilities must include either an app, package or browser';
       log.errorAndThrow(msg);
     }
-    // make sure that the capabilities don't have both `app` and `browser`
+    // warn if the capabilities have both `app` and `browser, although this
+    // is common with selenium grid
     if (caps.browserName && caps.app) {
-      let msg = 'The desired capabilities should not include both an app and a browser';
-      log.errorAndThrow(msg);
+      let msg = 'The desired capabilities should generally not include both an app and a browser';
+      log.warn(msg);
     }
   }
 
--- a/test/unit/driver-specs.js
+++ b/test/unit/driver-specs.js
@@ -244,7 +244,7 @@ describe('driver', () => {
       }).to.throw(/must include/);
       expect(() => {
         driver.validateDesiredCaps({platformName: 'Android', deviceName: 'device', browserName: 'Netscape Navigator'});
-      }).to.throw(/Netscape Navigator/);
+      }).to.throw(/must include/);
     });
     it('should not throw an error if caps contain an app, package or valid browser', () => {
       expect(() => {
@@ -257,15 +257,15 @@ describe('driver', () => {
         driver.validateDesiredCaps({platformName: 'Android', deviceName: 'device', appPackage: 'some.app.package'});
       }).to.not.throw(/must include/);
     });
-    it('should not be sentivie to platform name casing', () => {
+    it('should not be sensitive to platform name casing', () => {
       expect(() => {
         driver.validateDesiredCaps({platformName: 'AnDrOiD', deviceName: 'device', app: '/path/to/some.apk'});
       }).to.not.throw(Error);
     });
-    it('should throw an error if caps contain both an app and browser', () => {
+    it('should not throw an error if caps contain both an app and browser, for grid compatibility', () => {
       expect(() => {
-        driver.validateDesiredCaps({platformName: 'Android', deviceName: 'device', app: '/path/to/some.apk', browserName: 'Chrome'});
-      }).to.throw(/should not include both/);
+        driver.validateDesiredCaps({platformName: 'Android', deviceName: 'device', app: '/path/to/some.apk', browserName: 'iPhone'});
+      }).to.not.throw(Error);
     });
   });
   describe('proxying', () => {
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -1,5 +1,3 @@
-import { CHROME_BROWSERS } from './android-helpers';
-
 let commonCapConstraints = {
   platformName: {
     isString: true,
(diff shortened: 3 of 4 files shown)


jessehudl:
Looks like I'll need to update tests to go along with this, since they explicitly test for these cases. I'd like to get some feedback if possible, before spending the time to do that.

jlipps:
This makes sense to me. Though I'd like to keep the check and emit a warning so people know why there might be odd behavior if they use both caps. Does that sound good to you?

imurchie:
I agree with having a log message that both are set. Otherwise this is fine.

lib/driver.js
```diff
@@ -317,10 +317,11 @@ class AndroidDriver extends BaseDriver {
...
       log.errorAndThrow(msg);
     }
-    // make sure that the capabilities don't have both `app` and `browser`
+    // warn if the capabilities have both `app` and `browser, although this
+    // is common with selenium grid
     if (caps.browserName && caps.app) {
-      let msg = 'The desired capabilities should not include both an app and a browser';
-      log.errorAndThrow(msg);
+      let msg = 'The desired capabilities should generally not include both an app and a browser';
+      log.warn(msg);
```
jessehudl: Changed message to include the word "generally" since it's not necessarily bad, also make a not in the comment that it's common when using Appium with Selenium Grid. Also, no longer throw error.

test/unit/driver-specs.js
```diff
@@ -244,7 +244,7 @@ describe('driver', () => {
       }).to.throw(/must include/);
       expect(() => {
         driver.validateDesiredCaps({platformName: 'Android', deviceName: 'device', browserName: 'Netscape Navigator'});
-      }).to.throw(/Netscape Navigator/);
+      }).to.throw(/must include/);
```
jessehudl: Since `validateDesiredCaps` when passing in an unknown browser no longer fails with:
```
The desiredCapabilities object was not valid for the following reason(s): browserName Netscape Navigator is not included in the list.
```

Change the test to check for the error:
```
The desired capabilities must include either an app, package or browser
```

test/unit/driver-specs.js
```diff
@@ -257,16 +257,11 @@ describe('driver', () => {
         driver.validateDesiredCaps({platformName: 'Android', deviceName: 'device', appPackage: 'some.app.package'});
       }).to.not.throw(/must include/);
     });
-    it('should not be sentivie to platform name casing', () => {
+    it('should not be sensitive to platform name casing', () => {
```
jessehudl: Noticed this typo while I was in here, so fixed it in a separate commit.

test/unit/driver-specs.js
```diff
@@ -257,16 +257,11 @@ describe('driver', () => {
...
       expect(() => {
         driver.validateDesiredCaps({platformName: 'AnDrOiD', deviceName: 'device', app: '/path/to/some.apk'});
       }).to.not.throw(Error);
     });
-    it('should throw an error if caps contain both an app and browser', () => {
-      expect(() => {
-        driver.validateDesiredCaps({platformName: 'Android', deviceName: 'device', app: '/path/to/some.apk', browserName: 'Chrome'});
-      }).to.throw(/should not include both/);
-    });
```
jessehudl: Remove this test since this is a valid combination when used as a Grid node.

jlipps:
@jessehudl LGTM now; can you please squash into one commit?

lib/driver.js
```diff
@@ -317,10 +317,11 @@ class AndroidDriver extends BaseDriver {
       let msg = 'The desired capabilities must include either an app, package or browser';
       log.errorAndThrow(msg);
     }
-    // make sure that the capabilities don't have both `app` and `browser`
+    // warn if the capabilities have both `app` and `browser, although this
+    // is common with selenium grid
     if (caps.browserName && caps.app) {
-      let msg = 'The desired capabilities should not include both an app and a browser';
-      log.errorAndThrow(msg);
+      let msg = 'The desired capabilities should generally not include both an app and a browser';
```
jessehudl: Changed message to include the word "generally" since it's not necessarily bad, also make a note in the comment that it's common when using Appium with Selenium Grid.

jessehudl:
@jlipps I did 2 commits since one of the changes is unrelated to this work and just fixes a typo I noticed. Do you still want me to squash?

jessehudl:
Looks like there's an e2e test for this too, I'll squash and remove that one as well.

jlipps:
ok, everything looks good now. thanks! i'll let @imurchie do a pass and merge

jessehudl:
Sorry guys, I just pushed again. I decided to change the existing unit test to specifically check that this case **does** pass, so it doesn't break in the future.

Just curious, is SauceLabs using Selenium Grid? Since most of the Appium maintainers seem to work for Sauce, I'm wondering if it's actually getting tested in this way.

jlipps:
@jessehudl no, Sauce does not use Appium grid :-)

jessehudl:
@jlipps ah, secret sauce huh? 😬

Anyways, everything's passing and should be ready for code review, or merge 😀

imurchie:
Looks good to me! Thanks for this! And thanks for taking the time to fix the typo!

PR #141, opened 2016-05-06
Auto-select devices based on platform version

midopa:
Here's a draft implementation for appium/appium#6473.

Tested with an Android 6.0 device and a 4.0 emulator running.

Spun up an Appium instance from the Java client with PLATFORM_VERSION = 4 and got the emulator. 6 got my phone. 5 got nothing and the server returned an error response and remained online.

Let me know what you think. Once we have quorum I'll add tests.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -107,24 +107,63 @@ helpers.getDeviceInfoFromCaps = async function (opts = {}) {
   let udid = opts.udid;
   let emPort = null;
 
+  // a specific avd name was given. try to initialize with that
   if (opts.avd) {
     await helpers.prepareEmulator(adb, opts);
     udid = adb.curDeviceId;
     emPort = adb.emulatorPort;
   } else {
+    // no avd given. lets try whatever's plugged in devices/emulators
     logger.info("Retrieving device list");
     let devices = await adb.getDevicesWithRetry();
+
+    // udid was given, lets try to init with that device
     if (udid) {
       if (!_.contains(_.pluck(devices, 'udid'), udid)) {
         logger.errorAndThrow(`Device ${udid} was not in the list ` +
                              `of connected devices`);
       }
       emPort = adb.getPortFromEmulatorString(udid);
+    } else if (!!opts.platformVersion) {
+      // a platform version was given. lets try to find a device with the same os
+      logger.info(`Looking for a device with Android ${opts.platformVersion}`);
+
+      // in case we fail to find something, give the user a useful log that has
+      // the device udids and os versions so they know what's available
+      let availDevicesStr = [];
+
+      // first try started devices/emulators
+      for (let device of devices) {
+        // direct adb calls to the specific device
+        await adb.setDeviceId(device.udid);
+        let deviceOS = await adb.getPlatformVersion();
+
+        // build up our info string of available devices as we iterate
+        availDevicesStr.push(`${device.udid} (${deviceOS})`);
+
+        // we do a begins with check for implied wildcard matching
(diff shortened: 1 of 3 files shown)


lib/android-helpers.js
```diff
@@ -107,24 +107,67 @@ helpers.getDeviceInfoFromCaps = async function (opts = {}) {
   let udid = opts.udid;
   let emPort = null;
 
+  // a specific avd name was given. try to initialize with that
   if (opts.avd) {
     await helpers.prepareEmulator(adb, opts);
     udid = adb.curDeviceId;
     emPort = adb.emulatorPort;
-  } else {
+  }
```
jlipps: code style nitpick; we'd do this as:

```js
} else {
  // no avd given. lets try whatever's plugged in devices/emulators
```

lib/android-helpers.js
```diff
@@ -107,24 +107,67 @@ helpers.getDeviceInfoFromCaps = async function (opts = {}) {
...
       emPort = adb.getPortFromEmulatorString(udid);
-    } else {
+    }
+    // a platform version was given. lets try to find a device with the same os
+    else if (!!opts.platformVersion) {
+      logger.info(`Looking for a device with Android ${opts.platformVersion}`);
+
+      // in case we fail to find something, give the user a useful log that has
+      // the device udids and their os versions so they know what's available
+      let device_strs = [];
```
jlipps: we use `camelCase`

jlipps:
@midopa I like this approach! It actually seems solid to me apart from the few style comments I made. so I think adding unit and functional tests is a go from my perspective. What do you think @imurchie?

I guess we also need to be aware that this could potentially be a breaking change for users who are currently using platformVersion but for whatever reason relying on appium just giving them the first device.

imurchie:
This looks like a good approach. I don't see much of a problem with branching off `platformVersion`. If people start complaining we can always add a capability to specify that condition. add tests amd we're good to go!

midopa:
Tests added. Will update docs now.

midopa:
Docs added.

test/unit/android-helper-specs.js
```diff
@@ -129,21 +129,45 @@ describe('Android Helpers', () => {
...
+
     before(() => {
       sinon.stub(ADB, 'createADB', async () => {
         return {
           getDevicesWithRetry: async () => {
-            return [
-              {udid: 'emulator-1234'},
-              {udid: 'rotalume-1337'}
-            ];
+            let udids = [];
```
imurchie: lodash has some handy functions for this sort of thing.

The equivalent would be

```js
return _.map(devices, (device) => { return {udid: device.udid}; });
```

test/unit/android-helper-specs.js
```diff
@@ -129,21 +129,45 @@ describe('Android Helpers', () => {
...
             return 1234;
           },
           getRunningAVD: () => {
             return {'udid': 'emulator-1234', 'port': 1234};
           },
+          setDeviceId: (udid) => {
+            curDeviceId = udid;
+          },
+          getPlatformVersion: () => {
+            for (let device of devices) {
```
imurchie: Similarly, this can be

```js
return _.filter(devices, {udid: curDeviceId})[0].os;
```

imurchie:
This is great! A couple of comments with lodash equivalents.

imurchie:
Looking into why Jenkins isn't responding. But here is the console from the above run: https://team-appium.ci.cloudbees.com/job/Appium-android-driver/316/console

The error does not seem to be from your code changes, I don't think. Feels like a flake.

imurchie:
Thanks for this PR! It's a good development for Appium!

PR #142, opened 2016-05-10
Make sure ignoreUnimportantViews cap gets passed through

imurchie:
If we try to set the cap before initializing bootstrap then we get an error when trying to call a method on undefined.


jlipps:
👍

imurchie:
I'm still trying to get cloudbees to report back about the build status beyond it being triggered or running. For reasons that are not yet apparent it doesn't report passed/failed/errored.

Passed: https://team-appium.ci.cloudbees.com/job/Appium-android-driver/322/

PR #148, opened 2016-05-19
Fix for setNetworkConnection API

SrinivasanTarget:
Fixes: appium/appium#5256

@imurchie @jlipps @TikhomirovSergey Please review. This fixes the API to work on all the Android API till API Level 23 (Tested from 19 to 23)

### DIFF
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -38,7 +38,8 @@ commands.setNetworkConnection = async function (type) {
   });
   if (!airplaneMode) {
     await this.wrapBootstrapDisconnect(async () => {
-      await this.adb.setWifiAndData({wifi, data});
+      await this.adb.setWifiState(wifi);
+      await this.adb.setDataState(data);
     });
   }
 
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -38,8 +38,8 @@ commands.setNetworkConnection = async function (type) {
   });
   if (!airplaneMode) {
     await this.wrapBootstrapDisconnect(async () => {
-      await this.adb.setWifiState({wifi});
-      await this.adb.setDataState({data});
+      await this.adb.setWifiState(wifi);
+      await this.adb.setDataState(data);
     });
   }
 
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -38,8 +38,8 @@ commands.setNetworkConnection = async function (type) {
   });
   if (!airplaneMode) {
     await this.wrapBootstrapDisconnect(async () => {
-      await this.adb.setWifiState(wifi);
-      await this.adb.setDataState(data);
+      await this.adb.setWifiState({wifi});
+      await this.adb.setDataState({data});
     });
   }
 
--- a/test/unit/commands/network-specs.js
+++ b/test/unit/commands/network-specs.js
@@ -76,8 +76,8 @@ describe('Network', () => {
       await driver.setNetworkConnection(1);
       driver.adb.setAirplaneMode.calledWithExactly(1).should.be.true;
       driver.adb.broadcastAirplaneMode.calledWithExactly(1).should.be.true;
-      driver.adb.setWifiState.calledWithExactly(1).should.be.true;
-      driver.adb.setDataState.calledWithExactly(1).should.be.true;
+      driver.adb.setWifiState.called.should.be.false;
+      driver.adb.setDataState.called.should.be.false;
     });
     it('should turn on wifi', async () => {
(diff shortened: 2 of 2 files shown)


lib/commands/network.js
```diff
@@ -38,7 +38,8 @@ commands.setNetworkConnection = async function (type) {
   });
   if (!airplaneMode) {
     await this.wrapBootstrapDisconnect(async () => {
-      await this.adb.setWifiAndData({wifi, data});
+      await this.adb.setWifiState(wifi);
+      await this.adb.setDataState(data);
```
imurchie: I think these need to take an object, still. Like in the toggle methods.

```
await this.adb.setWifiAndData({wifi});
await this.adb.setWifiAndData({data});
```

SrinivasanTarget: Fixed now.

imurchie:
Unit tests need to be fixed to allow for multiple calls to `setWifiAndData`.

You can run those tests locally with `gulp once` (or, while you work, you can use `gulp watch` to run the tests whenever you save.

SrinivasanTarget:
@imurchie Sure will fix it

SrinivasanTarget:
What does this mean?
```
Network SetNetworkConnection should turn off wifi and data:
     AssertionError: expected false to be true
      at doAsserterAsyncAndAddThen (node_modules/chai-as-promised/lib/chai-as-promised.js:296:33)
      at [object Object].<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:286:25)
      at [object Object].Object.defineProperty.get (node_modules/chai/lib/chai/utils/overwriteProperty.js:50:37)
      at Object.callee$2$0$ (test/unit/commands/network-specs.js:72:61)
      at tryCatch (node_modules/babel-runtime/regenerator/runtime.js:67:40)
      at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/babel-runtime/regenerator/runtime.js:315:22)
      at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (node_modules/babel-runtime/regenerator/runtime.js:100:21)
      at GeneratorFunctionPrototype.invoke (node_modules/babel-runtime/regenerator/runtime.js:136:37)
      at bound (domain.js:287:14)
      at GeneratorFunctionPrototype.runBound (domain.js:300:12)
      at run (node_modules/core-js/library/modules/es6.promise.js:104:47)
      at node_modules/core-js/library/modules/es6.promise.js:115:28
      at flush (node_modules/core-js/library/modules/$.microtask.js:19:5)
```

imurchie:
What is the test code that is producing this? It means that the line `test/unit/commands/network-specs.js:72` is returning `false` when you expect `true`.

SrinivasanTarget:
@imurchie It's thrown from here: https://github.com/appium/appium-android-driver/pull/148/files#diff-ba240c2e81294a2c5bffa5ebed970081R72

imurchie:
Ah! But they are not called with '1' but with an object, like "{wifi: 1}"
or something, no?

SrinivasanTarget:
@imurchie you mean like this,
```
driver.adb.setWifiState.calledWithExactly({wifi:0}).should.be.false;
```

imurchie:
Something like that, yes. And you can positively check that what you send
in is what is there.

SrinivasanTarget:
Have tried the same but still failed

lib/commands/network.js
SrinivasanTarget: @imurchie I think we need to revert this change you have suggested as it should be the same way as,
```
await this.adb.setAirplaneMode(airplaneMode);
```
Airplane mode and Wifi state are defined the same way in appium-adb. Toggle methods take objects whereas these are not taking object as is.

SrinivasanTarget: @imurchie Im reverting this as these methods doesnt take objects as inputs here,https://github.com/appium/appium-adb/pull/158/files#diff-83add4eba9c23315341d94aec97ce76bR291

SrinivasanTarget:
Build is **green**: https://team-appium.ci.cloudbees.com/view/Appium%20Libs/job/Appium-android-driver/334/

@imurchie @jlipps Now your thoughts on this please.

lib/commands/network.js
imurchie: Yes. Sorry. I didn't see that it was `setWifiState` and `setDataState`, not the `setWifiAndDataState` or whatever.

imurchie:
Looks good!

imurchie:
Published as `appium-android-driver@1.10.6`.

Issue #146, opened 2016-05-19
The appium-android-driver cannot unlock the screen on some devices


nghiadhd:
Appium@1.5.2
Issue devices:
1. Galaxy Note 3: Model: SM-N9005, version: 4.4.2
2. Galaxy S4: Model: GT-I9500, version: 4.4.2
3. HTC one: Model: HTC6525LVW, version: 5.0.1

Other devices still work well

Note: In the past, the Galaxy S4 still works well, but after upgrade to newer OS version, it doesn't work any more

Issue: "Encountered internal error running command: Error: Screen did not unlock successfully"
Expectation: The screen should be unlocked

---

After do some investigating, found that the adb command to start the unlock doesn't work
`adb with args: ["-P",5037,"-s","xxxxxxxxxx","shell","am","start","-n","io.appium.unlock/.Unlock","-S","-a","android.intent.action.MAIN","-c","android.intent.category.LAUNCHER","-f","0x10200000"]`

I can reproduce the issue by adb command

adb -s <UDID> shell am start -n 'io.appium.unlock/.Unlock' -S -a 'android.intent.action.MAIN' -c 'android.intent.category.LAUNCHER' -f '0x10200000'

Some addresses

> The '-S' option doesn't work (should be remove from command)
> If the unlock is on foreground, the command doesn't make effect

My solution, use 'adb shell am force-stop' to kill 'unlock' and 'adb shell am start' to restart 'unlock' app. I tested on issue devices and confirm it works well

```
adb -s <UDID>  shell am force-stop 'io.appium.unlock
adb -s <UDID>  shell am start -n 'io.appium.unlock/.Unlock' -a 'android.intent.action.MAIN' -c 'android.intent.category.LAUNCHER' -f '0x10200000'
adb -s <UDID>  shell am start -n 'io.appium.unlock/.Unlock' -a 'android.intent.action.MAIN' -c 'android.intent.category.LAUNCHER' -f '0x10200000'
```

Weird that after force-stop, we should call 'am start' 2 times to invoke the 'unlock' app. That's should be the reason why '-S' option doesn't work

imurchie:
Interesting! Thank you for the detailed update. I'll take a look at this!

nghiadhd:
Thank @imurchie 

That's seem a critical issue. The test script always failed if the screen cannot unlocked

rkavalap:
[...]

PR #149, opened 2016-05-26
Further de-flaking unlock

imurchie:
Through the research of @nghiadhd, it seems that there is a way to make unlocking Android devices less flakey. Essentially (1) manually stop the unlock helper, (2) start the unlock helper twice in succession. Locally the app unlocks consistently with this.

See https://github.com/appium/appium-android-driver/issues/146

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import path from 'path';
 import { exec } from 'teen_process';
-import { retry, retryInterval, sleep } from 'asyncbox';
+import { retry, retryInterval } from 'asyncbox';
 import logger from './logger';
 import { fs } from 'appium-support';
 import { path as unicodeIMEPath } from 'appium-android-ime';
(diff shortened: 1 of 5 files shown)


lib/android-helpers.js
```diff
@@ -373,13 +373,23 @@ helpers.unlock = async function (adb) {
...
       activity: ".Unlock",
       action: "android.intent.action.MAIN",
       category: "android.intent.category.LAUNCHER",
-      flags: "0x10200000"
-    });
+      flags: "0x10200000",
+      stopApp: false
+    };
+    await adb.startApp(startOpts);
+    await adb.startApp(startOpts);
```
jlipps: I wonder why starting it twice would make a difference. Is it something we could fix in the unlock_apk itself? (Like maybe running its own internal logic twice?)

lib/android-helpers.js
```diff
@@ -373,13 +373,23 @@ helpers.unlock = async function (adb) {
...
+
+    // wait a moment, and check if it worked
     await sleep(1000);
```
jlipps: seeing this code---i know it's not part of this PR but don't you think we should use a retryInterval rather than a static sleep here?

lib/android-helpers.js
imurchie: That would be worth a try. To be perfectly honest the whole Android unlock process is opaque to me.

lib/android-helpers.js
imurchie: The whole thing is already within a `retryInterval` loop. Here we only need to check once, so I'm not sure a `retryInterval` would give us much. We could check at the very beginning of the loop, and break out if it is unlocked. But then the last time through would not have any check to see if it had worked.

jlipps: ok; what effect would simply removing this sleep have?

lib/android-helpers.js
nghiadhd: We cannot fix unlock_apk itself, This issue isn't bug of unlock code, the root cause is the unlock app does not invoke when we call 'am start' after 'force-stop' by ADB

@imurchie,  could you move 'forceStop' out side retryInterval, because retryInterval itself call adb.startApp many time

`await adb.forceStop('io.appium.unlock');

      logger.debug("Screen is locked, trying to unlock");
      let startOpts = {
        pkg: "io.appium.unlock",
        activity: ".Unlock",
        action: "android.intent.action.MAIN",
        category: "android.intent.category.LAUNCHER",
        flags: "0x10200000",
        stopApp: false
      };
      await adb.startApp(startOpts);
})`

imurchie: Moving it outside breaks things. It will work on a fresh emulator but when multiple tests are run it causes the emulator to become unresponsive.

imurchie:
Added a call to force stop the unlocker at the end of a session, which solves many of the problems (e.g., fixes #6262, #4821).

imurchie:
https://team-appium.ci.cloudbees.com/job/Appium-android-driver/342/console

jlipps:
👍

PR #153, opened 2016-06-08
add support to set UiAutomator Congfigurator values

truebit:
This is a sub PR for appium/appium#6561

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -45,10 +45,10 @@ class AndroidDriver extends BaseDriver {
     this.settings = new DeviceSettings({
       ignoreUnimportantViews: false,
       waitForIdleTimeout: 10000,
-      waitForSelectorTimeout:10000,
-      keyInjectionDelay:0,
-      actionAcknowledgmentTimeout:3000,
-      scrollAcknowledgmentTimeout:200
+      waitForSelectorTimeout: 10000,
+      keyInjectionDelay: 0,
+      actionAcknowledgmentTimeout: 3000,
+      scrollAcknowledgmentTimeout: 200
       }, this.onSettingsUpdate.bind(this));
     this.chromedriver = null;
     this.apkStrings = {};
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -296,7 +296,7 @@ class AndroidDriver extends BaseDriver {
   // Set UiAutomator Configurator related configurations
   // value sould be like {"method":"setWaitForIdleTimeout","value":5000}
   async setConfiguratorConfig (value) {
-    await this.bootstrap.sendAction("configurator", JSON.parse(value))
+    await this.bootstrap.sendAction("configurator", JSON.parse(value));
   }
 
   async deleteSession () {
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -42,8 +42,11 @@ class AndroidDriver extends BaseDriver {
     this.sessionChromedrivers = {};
     this.jwpProxyActive = false;
     this.jwpProxyAvoid = _.clone(NO_PROXY);
-    this.settings = new DeviceSettings({ignoreUnimportantViews: false},
-                                       this.onSettingsUpdate.bind(this));
+    this.settings = new DeviceSettings({ignoreUnimportantViews: false,
+                                        setWaitForIdleTimeout: 10000, setWaitForSelectorTimeout:10000,
+                                        setKeyInjectionDelay:0,
+                                        setActionAcknowledgmentTimeout:3000, setScrollAcknowledgmentTimeout:200
+                                       }, this.onSettingsUpdate.bind(this));
     this.chromedriver = null;
     this.apkStrings = {};
     this.acceptSslCerts = !!opts.acceptSslCerts;
@@ -154,6 +157,9 @@ class AndroidDriver extends BaseDriver {
   async onSettingsUpdate (key, value) {
     if (key === "ignoreUnimportantViews") {
       await this.setCompressedLayoutHierarchy(value);
+    }
+    if (key === "configurator") {
+      await this.setConfiguratorConfig(value);
     }
   }
 
@@ -285,6 +291,12 @@ class AndroidDriver extends BaseDriver {
   // Set CompressedLayoutHierarchy on the device
   async setCompressedLayoutHierarchy (compress) {
     await this.bootstrap.sendAction("compressedLayoutHierarchy", {compressLayout: compress});
+  }
+
+  // Set UiAutomator Configurator related configurations
+  // value sould be like {"method":"setWaitForIdleTimeout","value":5000}
+  async setConfiguratorConfig (value) {
+    await this.bootstrap.sendAction("configurator", JSON.parse(value))
   }
 
   async deleteSession () {
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -163,7 +163,6 @@ class AndroidDriver extends BaseDriver {
     } else {
       await this.setConfiguratorConfig({key, value});
     }
-
   }
 
   async startAndroidSession () {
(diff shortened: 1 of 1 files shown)


lib/driver.js
```diff
@@ -42,8 +42,11 @@ class AndroidDriver extends BaseDriver {
     this.sessionChromedrivers = {};
     this.jwpProxyActive = false;
     this.jwpProxyAvoid = _.clone(NO_PROXY);
-    this.settings = new DeviceSettings({ignoreUnimportantViews: false},
-                                       this.onSettingsUpdate.bind(this));
+    this.settings = new DeviceSettings({ignoreUnimportantViews: false,
```
jlipps: this needs a little formatting work. first, let's do one key/value per line. second, we do `key: value` (with a space after colon. third, we do:

```js
foo({
  new: 'line',
  before: 'keys/values'
});
```

to save horizontal space :-)

lib/driver.js
```diff
@@ -155,6 +158,9 @@ class AndroidDriver extends BaseDriver {
     if (key === "ignoreUnimportantViews") {
       await this.setCompressedLayoutHierarchy(value);
     }
+    if (key === "configurator") {
+      await this.setConfiguratorConfig(value);
```
jlipps: what would configurator 'values' look like here? wouldn't they actually be a combination of key and value? are they supposed to be JSON objects?

lib/driver.js
```diff
@@ -287,6 +293,12 @@ class AndroidDriver extends BaseDriver {
     await this.bootstrap.sendAction("compressedLayoutHierarchy", {compressLayout: compress});
   }
 
+  // Set UiAutomator Configurator related configurations
+  // value sould be like {"method":"setWaitForIdleTimeout","value":5000}
+  async setConfiguratorConfig (value) {
+    await this.bootstrap.sendAction("configurator", JSON.parse(value));
```
jlipps: ah yes I see you're expecting JSON objects. I'd love to see a bit of nicer error handling, something like:

```js
let param;
try {
  param = JSON.parse(value);
} catch (e) {
  log.errorAndThrow(`Configurator values must be valid JSON! Original error: ${e}`);
}
await this.bootstrap.sendAction("configurator", param);
```

jlipps:
couple comments, then again a test would be nice!

jlipps:
OK, I've had more time to grok what you are trying to do here, and I'd like to suggest a change in the approach.

Currently, you're asking the client to do something like:

```
POST /wd/hub/session/<sessionId/appium/settings
```

with a request body of:

```json
{"settings": {"configurator": "{\"setKeyIntervalDelay\": 1000}"}}
```

In other words, you're creating a "configurator" group of settings that have their own internal API. Instead what I'd like to see is simply all the configurator options as top-level settings. So the client request body would look like:

```json
{"settings": {"keyIntervalDelay": 1000}}
```

(Notice also no need to put `set` in `keyIntervalDelay`)

Then you'll need to change the way the code here works. Instead of:

```js
if (key === "configurator") {
  await this.setConfiguratorConfig(value);
}
```

You'd have:

```js
if (key === "keyIntervalDelay") {
  await this.setConfiguratorConfig({"setKeyIntervalDelay": value});
}
```

And so on with all the configurator config options you want to expose to the clients via this settings API.

(And you'd no longer need to `JSON.parse` in `setConfiguratorConfig`)

Does all this make sense? I think this is a great addition to the driver, thanks!

truebit:
ok, I am now on vacation until next Monday. Will get on it after vocation. Thanks for the advice:)

Concerning about the test, maybe we should create more methods to implement getting these configurations from Configurator? So tests could be done by setting firstly and then verifying via getting

Sent from my mobile device.

lib/driver.js
imurchie: I think it might be best to also check that it is a String before trying to parse it. I can see problems if an actual JSON object (rather than a stringified version) was passed in.

truebit:
@jlipps Sorry for the late reply. I tried to refactor as per your last comment using top level setting for all Configurator related configurations, it seems not making it less complex:
1. first change `onSettingsUpdate` as you said:
```javascript
  async onSettingsUpdate (key, value) {
    if (key === "ignoreUnimportantViews") {
      await this.setCompressedLayoutHierarchy(value);
    }
    else if (key === "waitForIdleTimeout") {
      await this.setConfiguratorConfig({"waitForIdleTimeout": value});
    }
    else if (key === "waitForSelectorTimeout") {
      await this.setConfiguratorConfig({"waitForSelectorTimeout": value});
    }
    else if (key === "keyInjectionDelay") {
      await this.setConfiguratorConfig({"keyInjectionDelay": value});
    }
    else if (key === "actionAcknowledgmentTimeout") {
      await this.setConfiguratorConfig({"actionAcknowledgmentTimeout": value});
    }
    else if (key === "scrollAcknowledgmentTimeout") {
      await this.setConfiguratorConfig({"scrollAcknowledgmentTimeout": value});
    }
  }
```
2. change `setConfiguratorConfig` like below:
```javascript
      // Set UiAutomator Configurator related configurations
      // value sould be like {"waitForIdleTimeout":5000}
      async setConfiguratorConfig (value) {
        for (k in value) { // maybe need to check value object is a dictionary?
          await this.bootstrap.sendAction(k, value);
        }
      }
```
3. Then change `appium-android-bootstrap` to handle all these settings
```java
    map.put("waitForIdleTimeout", new ConfiguratorHandler());
    map.put("waitForSelectorTimeout", new ConfiguratorHandler());
    map.put("keyInjectionDelay", new ConfiguratorHandler());
    map.put("actionAcknowledgmentTimeout", new ConfiguratorHandler());
    map.put("scrollAcknowledgmentTimeout", new ConfiguratorHandler());
```
Is that all you expected?

jlipps:
Hi @truebit the point wasn't necessarily to make the code less complex, it was to make it more compatible and user-friendly. But some improvements I could make to your examples:

```js
  async onSettingsUpdate (key, value) {
    if (key === "ignoreUnimportantViews") {
      await this.setCompressedLayoutHierarchy(value);
    } else {
      await this.setConfiguratorConfig({key, value});
    }
  }
```

and

```js
  // Set UiAutomator Configurator related configurations
  // value sould be like {"waitForIdleTimeout":5000}
  async setConfiguratorConfig (config) {
    await this.bootstrap.sendAction(config.key, config.value);
  }
```

but yeah, this is generally the right idea!

truebit:
I am totally newbie of javascript. 
I think you mean `await this.setConfiguratorConfig({key__:__ value});`( notice the colon
and 
for key-value object, it cannot be invoked by its key value with `.key` or `.value`. I find two ways, one is to use `for...in`, one is using something like `Object.keys(config)[0]` to get key and `config[key]` to get value

truebit:
but the second param of `this.bootstrap.sendAction` must be a key-value object

truebit:
Concerning about the tests, after update settings, how should we get the real Configurator result to verify it?

jlipps:
> Concerning about the tests, after update settings, how should we get the real Configurator result to verify it?

How would you verify that it is working as a user?

truebit:
> How would you verify that it is working as a user?

IMO this could hardly verify from a user's perspective in code. I think verify by using Configurator's `get` method.
Does appium settings api has method to handlingGET method? like `onSettingsUpdate` being done in POST method? If so, we could override that method to get Congurator values to do the verification.

jlipps:
Yeah that's the trick, it doesn't have such a method. Hmm. I suppose you could expose one from the bootstrap just for testing...

imurchie:
Where are we on this PR?

truebit:
@imurchie 
to create another method internally to get the Configurator values to  verify this method is too complex and there is no API in appium to support passing this value through.

Recently I found that `uiObject.click()` would wait for `Configurator.getwaitForSelectorTimeout()` timeout to find an element. But I do not know how to trigger that in Appium.

jlipps:
@truebit what about putting a value on the result of the `GET /session/:id` call?

jlipps:
@truebit what do you think about that idea?

3a4oT:
@jlipps not sure if you are the right person to ping but can it be related to  https://stackoverflow.com/questions/49628336/appium-android-set-custom-timeout-between-click-tap

mykola-mokhnach:
Closed as obsolete

PR #155, opened 2016-06-20
Fix release after figure out coordinates

imurchie:
We want to fix the release options _after_ we massage the options for the other actions. Otherwise we end up with weird offsets instead of actual positions. Fixes https://github.com/appium/appium/issues/6602


imurchie:
https://team-appium.ci.cloudbees.com/job/Appium-android-driver/347/console

jlipps:
LGTM. no unit tests that could keep this logic from regressing?

imurchie:
There are no unit tests currently. It would be lovely to have them, but it would be a lot of time.

PR #156, opened 2016-06-21
Update clear, and tests

imurchie:
Make use of `adb.clearTextField` to try to clear better in Android.

### DIFF
--- a/test/functional/commands/keyboard-e2e-specs.js
+++ b/test/functional/commands/keyboard-e2e-specs.js
@@ -3,16 +3,26 @@ import chaiAsPromised from 'chai-as-promised';
 import _ from 'lodash';
 import sampleApps from 'sample-apps';
 import AndroidDriver from '../../..';
-
+import B from 'bluebird';
 
 chai.should();
 chai.use(chaiAsPromised);
+
+const BUTTON_CLASS = 'android.widget.Button';
+const EDITTEXT_CLASS = 'android.widget.EditText';
+const TEXTVIEW_CLASS = 'android.widget.TextView';
+
+const PACKAGE = 'io.appium.android.apis';
+const TEXTFIELD_ACTIVITY = '.view.TextFields';
+const KEYEVENT_ACTIVITY = '.text.KeyEventText';
 
 let defaultAsciiCaps = {
   app: sampleApps('ApiDemos-debug'),
   deviceName: 'Android',
   platformName: 'Android',
-  newCommandTimeout: 90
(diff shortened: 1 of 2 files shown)


test/functional/commands/keyboard-e2e-specs.js
```diff
@@ -42,6 +44,12 @@ async function runTextEditTest (driver, testText, keys = false) {
   return el;
 }
 
+async function clearKeyEvents (driver) {
+  let el = _.last(await driver.findElOrEls('class name', 'android.widget.Button', true));
+  driver.click(el.ELEMENT);
+  await B.delay(500);
```
jlipps: maybe a comment about the delay?

test/functional/commands/keyboard-e2e-specs.js
```diff
@@ -42,6 +44,12 @@ async function runTextEditTest (driver, testText, keys = false) {
   return el;
 }
 
+async function clearKeyEvents (driver) {
```
jlipps: sorry I'm a little confused what this function name means. How does this function clear keyevents? Or is it sending "clear key" events? Maybe a note about what button this is tapping and how that is related might help?

test/functional/commands/keyboard-e2e-specs.js
```diff
@@ -83,22 +96,21 @@ let tests = [
...
 
 let languageTests = [
   { label: 'should be able to send Tamil', text: 'சோதனை' },
   { label: 'should be able to send Gujarati', text: 'પરીક્ષણ' },
   { label: 'should be able to send Chinese', text: '测试' },
   { label: 'should be able to send Russian', text: 'тестирование' },
-  // skip rtl languages, which don't clear correctly atm
-  // { label: 'should be able to send Arabic', 'تجريب'],
-  // { label: 'should be able to send Hebrew', 'בדיקות'],
+  { label: 'should be able to send Arabic', text: 'تجريب' },
```
jlipps: yay!

test/functional/commands/keyboard-e2e-specs.js
imurchie: Will do. Essentially it just clears the view (the view just takes any key events coming in and prints their details out in a TextView). Previously we were clearing it by reloading the activity, but that is both inefficient and potentially hides issues with how the events are being processed (less so for key events, but I removed it in the interest of consistency).

jlipps: ahhh ok

PR #159, opened 2016-06-29
Refactored find.js, touch.js, element.js

sravanmedarapu:
As per the suggestions made in appium-uiautoamtor2-driver [PR](https://github.com/appium/appium-uiautomator2-driver/pull/26), refactored findElOrEls, setElementValue and performMultiAction implementations to avoid code duplication.

- Refactored [find.js](https://github.com/sravanmedarapu/appium-android-driver/blob/master/lib/commands/find.js#L30) to invoke corresponding driver implementation of [`doFindElementOrEls()`](https://github.com/sravanmedarapu/appium-android-driver/blob/master/lib/commands/find.js#L46)

     1. appium-android-driver implementation [refer](https://github.com/sravanmedarapu/appium-android-driver/blob/master/lib/commands/find.js#L7)

     1. appium-uiautomator2-driver implementation  [refer](https://github.com/sravanmedarapu/appium-uiautomator2-driver/blob/948be847bfab33384fdcfc31292448552e5225a4/lib/commands/find.js#L10)

- Refactored [element.js](https://github.com/sravanmedarapu/appium-android-driver/blob/73480c674cd696900f26b18bda688656d7562d5d/lib/commands/element.js#L26) to invoke corresponding implementation of [`doSetElementValue()`](https://github.com/sravanmedarapu/appium-android-driver/blob/73480c674cd696900f26b18bda688656d7562d5d/lib/commands/element.js#L39) 

   1. appium-android-driver implementation [refer](https://github.com/sravanmedarapu/appium-android-driver/blob/73480c674cd696900f26b18bda688656d7562d5d/lib/commands/element.js#L42)

    1. appium-uiautomator2-driver implementation [refer](https://github.com/sravanmedarapu/appium-uiautomator2-driver/blob/master/lib/commands/element.js#L35)

- Refactored [touch.js](https://github.com/sravanmedarapu/appium-android-driver/blob/73480c674cd696900f26b18bda688656d7562d5d/lib/commands/touch.js#L284) to invoke corresponding implementation of [`doPerformMultiAction()`](https://github.com/sravanmedarapu/appium-android-driver/blob/73480c674cd696900f26b18bda688656d7562d5d/lib/commands/touch.js#L299) 

   1. appium-android-driver implementation [refer](https://github.com/sravanmedarapu/appium-android-driver/blob/73480c674cd696900f26b18bda688656d7562d5d/lib/commands/touch.js#L302)

    1. appium-uiautomator2-driver implementation [refer](https://github.com/sravanmedarapu/appium-uiautomator2-driver/blob/master/lib/commands/touch.js#L3)

@jlipps, @imurchie, can you please review this PR.

Note: Few more functions yet to be refactored, though created this PR to make sure that we are good to continue this approach down the line.

### DIFF
--- a/lib/commands/find.js
+++ b/lib/commands/find.js
@@ -2,7 +2,17 @@ import _ from 'lodash';
 import { errors } from 'appium-base-driver';
 
 
-let commands = {}, helpers = {}, extensions = {};
+let helpers = {}, extensions = {};
+
+/**
+ * Reason for isolating doFindElementOrEls from findElOrEls is for reusing findElOrEls
+ * across android-drivers (like appium-uiautomator2-driver) to avoid code duplication.
+ * Other android-drivers (like appium-uiautomator2-driver) need to override doFindElementOrEls
+ * to facilitate findElOrEls.
+ */
+helpers.doFindElementOrEls = async function (params) {
+  return await this.bootstrap.sendAction('find', params);
+};
 
 // stategy: locator strategy
 // selector: the actual selector for finding an element
@@ -26,7 +36,7 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
   let element;
   let doFind = async () => {
     try {
-      element = await this.bootstrap.sendAction('find', params);
+      element = await this.doFindElementOrEls(params);
     } catch (err) {
       if (err.message && err.message.match(/An element could not be located/)) {
         // we are fine with this, just indicate a retry
@@ -36,7 +46,7 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
     }
 
     // we want to return false if we want to potentially try again
-    if (mult) {
+    if (params.multiple) {
       return element && element.length !== 0;
     } else {
       return !_.isNull(element);
@@ -46,7 +56,7 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
   try {
     await this.implicitWaitForCondition(doFind);
   } catch (err) {
-    if (err.message && err.message.match(/Condition unmet/)){
+    if (err.message && err.message.match(/Condition unmet/)) {
       // only get here if we are looking for multiple elements
       // condition was not met setting res to empty array
       element = [];
@@ -65,6 +75,6 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
   }
 };
 
-Object.assign(extensions, commands, helpers);
-export { commands, helpers };
+Object.assign(extensions, helpers);
+export { helpers };
 export default extensions;
--- a/lib/commands/element.js
+++ b/lib/commands/element.js
@@ -36,6 +36,16 @@ helpers.setElementValue = async function (keys, elementId, replace = false) {
     unicodeKeyboard: this.opts.unicodeKeyboard
   };
 
+  return this.doSetElementValue(params);
+};
+
+/**
+ * Reason for isolating doSetElementValue from setElementValue is for reusing setElementValue
+ * across android-drivers (like appium-uiautomator2-driver) and to avoid code duplication.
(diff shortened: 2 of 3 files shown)


lib/commands/find.js
```diff
@@ -16,55 +35,39 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
     throw new Error("Must provide a selector when finding elements");
   }
 
-  let params = {
+  this.params = {
```
imurchie: I'm not terribly happy about making the parameters for a method be stored on the driver object. Can `params` not remain a local variable that gets passed to `doFindElementOrEls`?

lib/commands/find.js
```diff
@@ -16,55 +35,39 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
...
-      throw err;
-    }
-
-    // we want to return false if we want to potentially try again
-    if (mult) {
-      return element && element.length !== 0;
-    } else {
-      return !_.isNull(element);
-    }
+    return this.doFindElementOrEls();
```
imurchie: We've been enforcing `return await someAsyncCall();` so that we get proper translation of promises in the event that the promises come from different libs.

imurchie: Since this closes over `params`, it ought to work for a local variable.

imurchie:
So the idea is that other drivers could import these methods and then override `doSetElementValue`, etc., to get their own functionality? That seems good. We should be careful to extract out as little as possible, so that subclasses don't need to do too much. If error handling can be generalized, that would be awesome!

sravanmedarapu:
Yeah, other drivers just need to override [`doSetElementValue`](https://github.com/sravanmedarapu/appium-uiautomator2-driver/blob/master/lib/commands/element.js#L35) or any other specific function as per their corresponding driver behavior .

lib/commands/find.js
imurchie: Perhaps something like...

```js
helpers.doFindElementOrEls = async function (params) {
  return await this.bootstrap.sendAction('find', params);
};


helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
...

  let params = { ... };
  let element;
  let doFind = async () => {
    try {
      element = helpers.doFindElementOrEls(params);
    } catch (err) {
      if (err.message && err.message.match(/An element could not be located/)) {
      // we are fine with this, just indicate a retry
      return false;
    }

    // we want to return false if we want to potentially try again
    if (mult) {
      return element && element.length !== 0;
    } else {
      return !_.isNull(element);
    }
  };
...
};
```

lib/commands/find.js
```diff
@@ -36,7 +40,7 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
     }
 
     // we want to return false if we want to potentially try again
-    if (mult) {
+    if (params.multiple) {
```
imurchie: This is fine, but also there is no need for the change (since `mult` exists in this scope now).

lib/commands/find.js
```diff
@@ -2,7 +2,11 @@ import _ from 'lodash';
 import { errors } from 'appium-base-driver';
 
 
-let commands = {}, helpers = {}, extensions = {};
+let helpers = {}, extensions = {};
+
+helpers.doFindElementOrEls = async function (params) {
```
imurchie: It would be great to have a little comment on this method (and all the extracted-out methods) saying why they are there. Otherwise someone might come along later and move them back in, thinking they are just redundant functions!

imurchie:
Looks good! Can you squash into a single commit, and we can get this in?

imurchie:
Thanks for this!

imurchie:
Published as `appium-android-driver@1.10.13`.

PR #160, opened 2016-06-29
Add eslint and fix problems

imurchie:
Clean things up so we can better enforce coding standards.


jlipps:
yay! eventually we probably want to axe jshint and make eslint part of appium-gulp-plugins, right?

👍

imurchie:
Yes to axing jshint. eslint is available through our plugins (just `gulp eslint`). It should eventually be in `gulp lint`, too, but that has to wait until all the packages support it.

PR #161, opened 2016-06-29
Replace constant REMOTE_INSTALL_TIMEOUT with androidInstallTimeout capability

spearminty:
Proposed change in android driver to address appium issue of adb install timeout value hardcoded.
ref: https://github.com/appium/appium-adb/issues/147

My proposal to fix is to add a new capability to appium "--adb-install-timeout"

This PR has updated and passed tests

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -215,7 +215,7 @@ helpers.getRemoteApkPath = function (localApkMd5) {
   return remotePath;
 };
 
-helpers.resetApp = async function (adb, localApkPath, pkg, fastReset) {
+helpers.resetApp = async function (adb, localApkPath, pkg, fastReset, androidInstallTimeout = REMOTE_INSTALL_TIMEOUT) {
   if (fastReset) {
     logger.debug("Running fast reset (stop and clear)");
     await adb.stopAndClear(pkg);
@@ -226,12 +226,12 @@ helpers.resetApp = async function (adb, localApkPath, pkg, fastReset) {
     if (!await adb.fileExists(remotePath)) {
       throw new Error("Can't run slow reset without a remote apk!");
     }
-    await helpers.reinstallRemoteApk(adb, localApkPath, pkg, remotePath);
+    await helpers.reinstallRemoteApk(adb, localApkPath, pkg, remotePath, androidInstallTimeout);
   }
 };
 
 helpers.reinstallRemoteApk = async function (adb, localApkPath, pkg,
-                                             remotePath, tries = 2) {
+                                             remotePath, androidInstallTimeout, tries = 2) {
   await retry(tries, async () => {
     try {
       // first do an uninstall of the package to make sure it's not there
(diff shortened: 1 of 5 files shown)


lib/android-helpers.js
```diff
@@ -11,7 +11,6 @@ import Bootstrap from 'appium-android-bootstrap';
 import ADB from 'appium-adb';
 
 const REMOTE_TEMP_PATH = "/data/local/tmp";
-const REMOTE_INSTALL_TIMEOUT = 90000; // milliseconds
```
imurchie: This should remain, to be used as the default (as it stands, there is no default, so if the `adbInstallTimeout` is not given, what happens?).

lib/android-helpers.js
```diff
@@ -215,7 +214,7 @@ helpers.getRemoteApkPath = function (localApkMd5) {
   return remotePath;
 };
 
-helpers.resetApp = async function (adb, localApkPath, pkg, fastReset) {
+helpers.resetApp = async function (adb, localApkPath, pkg, fastReset, adbInstallTimeout) {
```
imurchie: We should give a default here.

```js
helpers.resetApp = async function (adb, localApkPath, pkg, fastReset, adbInstallTimeout = REMOTE_INSTALL_TIMEOUT) {
```

imurchie:
I like this being configurable. I would suggest it should have a default set to the original static number.

I think it might be better to be something like `androidInstallTimeout` instead of `adbInstallTimeout`, as well, so we don't encode the underlying technology into the name (what happens if/when `adb` is no longer used to install? Does this timeout still apply?).

imurchie:
Otherwise, awesome! Thank you for the work in getting this done!

spearminty:
I agree with your feedback and have made changes accordingly

test/functional/android-helper-e2e-specs.js
```diff
@@ -4,8 +4,12 @@ import helpers from '../../lib/android-helpers';
 import sampleApps from 'sample-apps';
 import ADB from 'appium-adb';
 
-let apiDemos = sampleApps('ApiDemos-debug');
-let appPackage = 'io.appium.android.apis';
+let opts =
```
jlipps: according to our style guide, please make this:

```js
let opts = {
  localApkPath: sampleApps('ApiDemos-debug'),
  pkg: 'io.appium.android.apis',
  androidInstallTimeout: 90000
};
```

(note spacing, line breaks, indentation)

jlipps:
made one comment. thanks @spearminty. I think once you've addressed this and squashed commits this will be good to go

spearminty:
OK- formatting fixed and commits squashed

imurchie:
Excellent! Waiting for build and then I'm happy with merging!

imurchie:
Functional build passed: https://team-appium.ci.cloudbees.com/job/Appium-android-driver/367/console

imurchie:
Thank you for doing this! It's great!

PR #162, opened 2016-07-01
installApkRemotely to parse correct capabilities from opts

spearminty:
Per my previous PR I tested the new release within Appium and I found that my original change is **defective**
ref: #161 

The opts values for localApkPath and pkg parsed in android-helpers.js : line 257 are referred incorrectly.

The tests for this PR passed, but when using official capability arguments passed from appium the appium-android-driver will fail.

I have updated use of variables localApkPath and pkg to their respective capabilities: app and appPackage.


jlipps:
good catch. LGTM

PR #163, opened 2016-07-01
Set default androidInstallTimeout

spearminty:
A default value for this timeout is set within android-helpers.js and also exists downstream in android-adb if the cap is not provided. 
If the cap is not provided this message...
`
logger.info(`Pushing ${appPackage} to device. Will wait up to ${androidInstallTimeout} ` +
                  `milliseconds before aborting`);
`
...will print a timeout value of 'undefined' yet still continue with the correct default value of 90000

This PR will fix the misleading message provided to the upstream appium consumer.

Apologizes for the stacked PRs around this single change

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -79,7 +79,8 @@ class AndroidDriver extends BaseDriver {
         tmpDir: await tempDir.staticDir(),
         fullReset: false,
         autoLaunch: true,
-        adbPort: DEFAULT_ADB_PORT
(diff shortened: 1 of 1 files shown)


spearminty:
bump

imurchie:
Whoops! 

Thanks for this. Makes sense!

imurchie:
Published as `appium-android-driver@1.10.14`

PR #165, opened 2016-07-10
Add support to save the screenshot in the user's mentioned directory.

akhillb:


lib/commands/actions.js
```diff
@@ -159,7 +159,8 @@ commands.pullFolder = async function (remotePath) {
 
 commands.getScreenshot = async function () {
   let localFile = temp.path({prefix: 'appium', suffix: '.png'});
-  const png = '/data/local/tmp/screenshot.png';
+  let png_dir = this.opts.androidScreenshotPath || '/data/local/tmp/';
+  const png = png_dir + '/screenshot.png';
```
imurchie: This should be like `const png = path.resolve(png_dir, 'screenshot.png');`

imurchie:
This desired cap should be added to the constraints, too. https://github.com/appium/appium-android-driver/blob/master/lib/desired-caps.js

imurchie:
This looks good other than a couple of little notes! Thanks for taking this on!

akhillb:
@imurchie thanks for pointing out the things i have missed. Made the changes now.

imurchie:
Very nice! Thank you!

imurchie:
Published in `appium-android-driver@1.10.16`

imurchie:
The final step is to add the capability to the documentation. It should go at the bottom of the "Android only" section of this document: https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/caps.md#android-only

The format should be the capability, a description of what it does (including mentioning the default value), and an example of a possible value.

akhillb:
@imurchie already made that change in this [PR](https://github.com/appium/appium/pull/6649)

imurchie:
Whoops! Thanks!

PR #166, opened 2016-07-12
kill emulator at start of session if -wipe-data required

alizelzele:
If the avd is defined and in avdArgs the -wipe-data argument is set, this update will close emulator before test to wipe its data.

This Pull Request depends on [adding killEmulator function to kill specific emulator with AVD name.](https://github.com/appium/appium-adb/pull/165), to kill emulator.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -49,8 +49,13 @@ helpers.prepareEmulator = async function (adb, opts) {
   let avdName = avd.replace('@', '');
   let runningAVD = await adb.getRunningAVD(avdName);
   if (runningAVD !== null) {
-    logger.debug("Not launching AVD because it is already running.");
-    return;
+    if (avdArgs && avdArgs.toLowerCase().indexOf("-wipe-data") > -1) {
+      logger.debug(`Killing '${avdName}' because it needs to be wiped at start.`);
+      await adb.killEmulator(avdName);
+    } else {
+      logger.debug("Not launching AVD because it is already running.");
+      return;
+    }
   }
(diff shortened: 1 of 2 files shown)


lib/android-helpers.js
```diff
@@ -49,8 +49,15 @@ helpers.prepareEmulator = async function (adb, opts) {
   let avdName = avd.replace('@', '');
   let runningAVD = await adb.getRunningAVD(avdName);
   if (runningAVD !== null) {
-    logger.debug("Not launching AVD because it is already running.");
-    return;
+    if (avdArgs.toLowerCase().indexOf("-wipe-data") > -1){
+      logger.debug("Killing AVD " + avdName + " because it needs to be wiped at start.");
```
imurchie: This should be an interpolated string

```js
logger.debug(`Killing AVD '${avdName}' because it needs to be wiped at start.`);
```

lib/driver.js
```diff
@@ -313,6 +313,11 @@ class AndroidDriver extends BaseDriver {
       }
       await this.adb.stopLogcat();
       await this.bootstrap.shutdown();
+      if (this.opts.avd && this.opts.avdArgs.toLowerCase().indexOf("-wipe-data") > -1){
+        let avdName = this.opts.avd.replace('@', '');
+        log.debug("closing emulator for avd " + avdName);
```
imurchie: Same.

imurchie:
This seems good. Once the `appium-adb` change gets in we can revisit.

alizelzele:
updated the request suing interpolated String.

imurchie:
Thanks! Awesome!

imurchie:
Published as `appium-android-driver@1.10.23`.

PR #168, opened 2016-07-14
Fix remaining installApkRemotely calls

imurchie:
The parameters to `installApkRemotely` were changed but two places were missed.

Cc: @spearminty


spearminty:
What functionality exposed this bug? Is this code under unit test coverage? Where are these general.js commands used vs calls to android-adb? 
I'd like to know so I can test better in the future

Thank you

imurchie:
I caught this while fixing an issue in `appium-selendroid-driver`. It seems like we do not at the moment have test coverage for these particular parts. Alas. It's in my queue but so is way too much other stuff.

PR #169, opened 2016-07-15
Rotate screenshot when device is not upright

tiagoshibata:
Fixes https://github.com/appium/appium/issues/3158.

Added [jimp](https://www.npmjs.com/package/jimp) as a dependency. I chose this library because it has no external/native dependencies and was simple to setup with npm.

Let me know if adding this dependency is fine and I will work a bit more on the code (mainly checking if I can easily optimize the read/rotate/write/re-read flow).

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -4,6 +4,8 @@ import { fs, mkdirp } from 'appium-support';
 import AdmZip from 'adm-zip';
 import path from 'path';
 import log from '../logger';
+import B from 'bluebird';
+import jimp from 'jimp';
 
 const swipeStepsPerSec = 28;
 const dragStepsPerSec = 40;
@@ -167,8 +169,18 @@ commands.getScreenshot = async function () {
     await fs.unlink(localFile);
   }
   await this.adb.pull(png, localFile);
-  let data = await fs.readFile(localFile);
-  let b64data = new Buffer(data).toString('base64');
+  let image = await jimp.read(localFile);
+  if (await this.adb.getApiLevel() < 23) {
+    // Android bug 8433742 - rotate screenshot if screen is rotated
+    let screenOrientation = await this.adb.getScreenOrientation();
+    try {
+      image = await image.rotate(-90 * screenOrientation);
+    } catch (err) {
+      log.warn(`Could not rotate screenshot due to error: ${err}`);
+    }
+  }
+  let b64data = (await B.promisify(image.getBuffer).call(image, jimp.MIME_PNG))
+                .toString('base64');
   await fs.unlink(localFile);
   return b64data;
 };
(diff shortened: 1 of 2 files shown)


imurchie:
`jimp` looks like a reasonable dependency. Thanks for taking this on!

lib/commands/actions.js
```diff
@@ -166,6 +167,16 @@ commands.getScreenshot = async function () {
     await fs.unlink(localFile);
   }
   await this.adb.pull(png, localFile);
+  let screenOrientation = await this.adb.getScreenOrientation();
+  if (screenOrientation) {
+    jimp.read(localFile).then(function (image) {
```
jlipps: assuming that `jimp.read` and `jimp.write` return promises, this section should be rewritten as:

```js
try {
  let image = await jimp.read(localFile);
  await image.rotate(90 * screenOrientation).write(localFile);
} catch (err) {
  log.warn(`Could not rotate screenshot due to error: ${err}`);
}
```

jlipps:
made one comment, thanks @tiagoshibata!

tiagoshibata:
Thanks, I'll update the code!

lib/commands/actions.js
tiagoshibata: I have rewritten it to use `jimp.getBuffer` instead of writing to disk and re-reading. My problem is that `jimp.write` and `jimp.getBuffer` don't return promises, but instead receive a callback. What do you recommend to do a call to `jimp.getBuffer` and wait until its callback has been triggered? Is there a suggested way to do this in Appium?

Currently, my code reads:
```js
  let screenOrientation = await this.adb.getScreenOrientation();
  let image = await jimp.read(localFile);
  try {
    image = await image.rotate(90 * screenOrientation);
  } catch (err) {
    log.warn(`Could not rotate screenshot due to error: ${err}`);
  }
  let b64data = // Some way to wait for image.getBuffer(jimp.MIME_PNG, function () {}) to finish and call toString('base64') on the data
```

Thanks for the help, JS isn't a language I am used to.

jlipps: sure, no problem! This is how you'd do it (assuming `import B from 'bluebird'` is at the top somewhere):

```js
let screenOrientation = await this.adb.getScreenOrientation();
let image = await jimp.read(localFile);
try {
  image = await image.rotate(90 * screenOrientation);
} catch (err) {
  log.warn(`Could not rotate screenshot due to error: ${err}`);
}
let b64data = (await B.promisify(image.getBuffer)(jimp.MIME_PNG))
                .toString('base64');
```

tiagoshibata:
I tested on more devices and found out the bug doesn't affect Android 6.0 upwards, since [this commit fixed the bug](https://github.com/android/platform_frameworks_base/commit/cf70d71781a40d197c3ecde89351976b27eb9a8a). I added a verification of the API level.

tiagoshibata:
Hi guys, let me know if something else is needed to merge this.

imurchie:
Looks good to me. Thanks for doing this!

imurchie:
Published in `appium-android-driver@1.10.17`.

PR #170, opened 2016-07-15
Add pre-commit hooks

imurchie:


package.json
```diff
@@ -50,8 +50,15 @@
   "scripts": {
     "prepublish": "./node_modules/.bin/gulp prepublish",
     "test": "./node_modules/.bin/gulp once",
-    "watch": "./node_modules/.bin/gulp"
+    "watch": "./node_modules/.bin/gulp",
+    "precommit-msg": "echo 'Pre-commit checks...' && exit 0",
```
jlipps: why do we need ` && exit 0`?

imurchie: It's mostly just to make the success explicit. The `pre-commit` package just looks for something truthy as a failure.

jlipps: ah ok

Issue #176, opened 2016-08-08
Use any random available port when starting ChromeDriver


midopa:
When I launch tests that use Chrome on multiple devices at the same time, all but one fails. This is because all the ChromeDriver sessions are trying to start on the same port, 9515.

I suggest that Appium uses the first available random port found for each ChromeDriver session. This will enable parallel Android browser tests on the same machine with no extra effort from users' POV.

imurchie:
This seems like a good idea! Thanks for taking it on, too. See my comment on your PR.

PR #177, opened 2016-08-08
Use random available port for ChromeDriver sessions

midopa:
WRT #176

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -1,6 +1,8 @@
 import _ from 'lodash';
 import logger from '../logger';
 import Chromedriver from 'appium-chromedriver';
+import PortFinder from 'portfinder';
+import B from 'bluebird';
 import { errors } from 'appium-base-driver';
 import { default as webviewHelpers,
          NATIVE_WIN, WEBVIEW_BASE, WEBVIEW_WIN, CHROMIUM_WIN } from '../webview-helpers';
@@ -192,6 +194,12 @@ async function setupExistingChromedriver (chromedriver) {
 }
 
 async function setupNewChromedriver (opts, curDeviceId, adbPort) {
+  // if a port wasn't given, pick a random available one
+  if (!opts.chromeDriverPort) {
+    let getPort = B.promisify(PortFinder.getPort, {context: PortFinder});
+    opts.chromeDriverPort = await getPort();
+    logger.debug(`A port was not given, using random port: ${opts.chromeDriverPort}`);
+  }
   let chromeArgs = {
     port: opts.chromeDriverPort,
     executable: opts.chromedriverExecutable,
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "io.appium.settings": "^2.1.0",
     "jimp": "^0.2.24",
     "lodash": "^3.10.0",
+    "portfinder": "^1.0.6",
     "source-map-support": "^0.3.1",
     "teen_process": "^1.4.0",
     "temp": "^0.8.3",


lib/commands/context.js
```diff
@@ -192,8 +194,16 @@ async function setupExistingChromedriver (chromedriver) {
 }
 
 async function setupNewChromedriver (opts, curDeviceId, adbPort) {
+  let chromePort;
+  PortFinder.getPort(function(err, port) {
+    chromePort = port;
+  });
+  // need to handle port finder's old callback style
+  while (chromePort === undefined) {
```
imurchie: This is the wrong way about things.

Try 

```js
import B from 'bluebird';

let getPort = B.promisify(PortFinder.getPort, {context: PortFinder});

let port = await getPort();
```

midopa: This is much better. Will do!

lib/commands/context.js
```diff
@@ -192,8 +194,11 @@ async function setupExistingChromedriver (chromedriver) {
 }
 
 async function setupNewChromedriver (opts, curDeviceId, adbPort) {
+  let getPort = BlueBird.promisify(PortFinder.getPort, {context: PortFinder});
+  let chromePort = await getPort();
+  logger.error(`Using ${chromePort} for new ChromeDriver session`);
   let chromeArgs = {
-    port: opts.chromeDriverPort,
+    port: chromePort,
```
jlipps: I don't think we should be overwriting `opts.chromeDriverPort` if it is set. Only if it's unset. Or, we need to do the random port-finding in a higher-level package (like Android Driver)

midopa: `opts.chromeDriverPort` was not being set by anything. I grepped all the code in `appium-android-driver` for that string.

jlipps: ok, sorry I was confused, I thought this was the `appium-chromedriver` package :-)

I'm a little conflicted because really in the future all this logic will live in the `driverData` section where a new driver will ensure it is not conflicting in terms of ports or other system resources with other drivers. but we can refactor that out.

Also, `--chromedriver-port` is a server flag in Appium that maps to the `chromeDriverPort` capability, and is probably how `opts.chromeDriverPort` gets populated. So I do think we want to respect it if it exists. Unfortunately it might always exist as a default...

imurchie: It will always exist, as `9515`. 

https://github.com/appium/appium/blob/master/lib/parser.js#L239

midopa: When is the `driverData` change coming? Where will it be located?

If the port is always given, then maybe the default should be changed to some invalid value (like -1) to tell the Android driver to select a random available port.

imurchie: If people are not relying on the port being 9515 (i.e., if when they don't set the port themselves they don't care what port is used) the default can be removed entirely, so it will be undefined if not set.

midopa: So, if we're on board with this change, two additions:

1. Remove/change default on Appium arg parser for ChromeDriver port
2. Only apply this logic if the given port is null.

jlipps: Yeah that makes sense. Change the default in `appium/appium` to be unset and then override with a random port here if indeed it is unset.

midopa: Something I noticed: In the logs, I see Appium tries to kill any old existing Chromedriver processes:
```
info Chromedriver Killing any old chromedrivers, running: pkill -15 -f "/Users/parks169/code/appium-android-driver/node_modules/appium-chromedriver/chromedriver/mac/chromedriver.*--port=9515"
```

The port is included in the `killAll` command. So if a Chromedriver process leaks with a random port, it will not be caught by this code.

One option is to remove the port in the command. This can introduce the problem of inadvertently killing other Chromedriver processes that are running but not spawned from Appium.

Another option is to base the `killAll` on the parent process/group. But I think if a previous Appium process leaked a Chromedriver process, our new Appium process won't be able to tell. It's just another PID number without any more information.

Maybe `killAll` should base its decision on the inactive time of each process? As in: Is it orphaned and/or stale?

lib/commands/context.js
```diff
@@ -192,6 +194,12 @@ async function setupExistingChromedriver (chromedriver) {
 }
 
 async function setupNewChromedriver (opts, curDeviceId, adbPort) {
+  // if a port wasn't given, pick a random available one
+  if (opts.chromeDriverPort === null || opts.chromeDriverPort === undefined) {
```
midopa: Note: `lint` Complained about using `x == null` here, despite it seeming to be the consensus for checking presence of a variable:
http://stackoverflow.com/a/2647888

imurchie: This is because for the vast majority of cases, `==` causes problems where `===` does not.

We generally just use `if (opts.chromeDriverPort)`. There is also a method in `appium-support`

```
import { util } from 'appium-support';

// ...

if (util.hasValue(opts.chromeDriverPort) {
// ...
```

midopa: I used `!opts.chromeDriverPort`. Tested in node and it works for both undefined and null fields in an object.

midopa:
Something I noticed: In the logs, I see Appium tries to kill any old existing Chromedriver processes:
```
info Chromedriver Killing any old chromedrivers, running: pkill -15 -f "/Users/parks169/code/appium-android-driver/node_modules/appium-chromedriver/chromedriver/mac/chromedriver.*--port=9515"
```

The port is included in the `killAll` command. So if a Chromedriver process leaks with a random port, it will not be caught by this code.

One option is to remove the port in the command. This can introduce the problem of inadvertently killing other Chromedriver processes that are running but not spawned from Appium.

Another option is to base the `killAll` on the parent process/group. But I think if a previous Appium process leaked a Chromedriver process, our new Appium process won't be able to tell. It's just another PID number without any more information.

Maybe `killAll` should base its decision on the inactive time of each process? As in: Is it orphaned and/or stale?

imurchie:
That is something that ought to be handled, yes. But it is also something that is currently the case, too, if the user specifies a port for ChromeDriver.

I think if we _can_ check the state of the processes, that would be a good solution.

midopa:
Sorry I nuked that comment. I realized the port isn't hardcoded, it's based on whatever's passed in the constructor. If nothing, then yes it'll use the default 9515. But since this change will always give it something, it's not a problem.

midopa:
Any more comments?

imurchie:
Just one, which is largely cosmetic but also for ease-of-understanding: everywhere else we use `B` as the variable name for the bluebird library. 

```js
import B from 'bluebird';
```

Other than that this looks good! Thanks!

imurchie:
Thank you, @midopa ! This is a good addition to the system.

imurchie:
Published in `appium-android-driver@1.10.19`.

PR #178, opened 2016-08-14
Passing the appWaitDurationParam to appium-adb startApp method

youngrafi:
This is required for appium/appium#780


jlipps:
👍

imurchie:
Thanks for this! This is great!

One final piece to the puzzle is to add the capability to the list [here](https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/caps.md#android-only).

imurchie:
Published in `appium-android-driver@1.10.20`.

PR #179, opened 2016-08-19
ensure we shut down logcat regardless of whether bootstrap was active

jlipps:
this ensures we don't leave procs hanging if we fail mid-start-session
fix appium/appium#6753

cc @imurchie


imurchie:
Good catch!

LGTM.

jlipps:
published as 1.10.21

PR #180, opened 2016-09-09
support reboot parameter for android

alizelzele:
If reboot option activated (will link pull request for that soon), it will always wipe data of emulator on start, and close emulator at the end.
It also use device name and platform version to create avd name, if it is not defined.

Android emulator sometimes get stuck if it is started for a long time without activity, and will not respond to any adb command, even if we kill adb-server, this update will kill emulator after session and wipe it on start to reduce the risk of this problem on production. the issue was already created about it. #167

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -107,6 +107,11 @@ class AndroidDriver extends BaseDriver {
         this.jwpProxyAvoid.push(['GET', new RegExp('^/session/[^/]+/screenshot')]);
       }
 
+      if (this.opts.reboot) {
+        this.setAvdFromCapabilities(caps);
+        this.addWipeDataToAvdArgs();
+      }
+
       // get device udid for this session
       let {udid, emPort} = await helpers.getDeviceInfoFromCaps(this.opts);
       this.opts.udid = udid;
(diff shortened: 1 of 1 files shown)


lib/driver.js
```diff
@@ -107,6 +107,25 @@ class AndroidDriver extends BaseDriver {
         this.jwpProxyAvoid.push(['GET', new RegExp('^/session/[^/]+/screenshot')]);
       }
 
+      if (this.opts.reboot) {
+        if (this.opts.avd) {
+          log.info('avd name defined, ignoring device name and platform version');
+        } else {
+          if (!caps.deviceName) {
+            log.errorAndThrow('avd or deviceName should be specified when reboot option is enables');
```
imurchie: enabled

imurchie:
This looks good to me, though I would prefer to have the logic pulled out into an appropriately-named method, for clarity.

alizelzele:
@imurchie Divided reboot code into two method, and fixed typo.

imurchie:
Looks good to me! Thanks for this!

imurchie:
Published in `appium-android-driver@1.10.26`.

PR #183, opened 2016-10-10
throw real error instead of deleteSession error

alizelzele:
If `createSession` throws exception it will try to delete session before throwing the exception. In case `deleteSession` throws another exception, appium will return `deleteSession`s exception instead of real error. If you are using grid ( + jenkins) you will see `deleteSession` error instead of the real reason.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -143,7 +143,11 @@ class AndroidDriver extends BaseDriver {
       await this.startAndroidSession(this.opts);
       return [sessionId, this.caps];
     } catch (e) {
-      await this.deleteSession();
+      // ignoring delete session exception if any and throw the real error
+      // that happened while creating the session.
+      try {
+        await this.deleteSession();
+      } catch (ign) {}
       throw e;
     }
   }


lib/driver.js
```diff
@@ -143,7 +143,10 @@ class AndroidDriver extends BaseDriver {
       await this.startAndroidSession(this.opts);
       return [sessionId, this.caps];
     } catch (e) {
-      await this.deleteSession();
+      try {
+        await this.deleteSession();
+      } catch (ignored) {
+      }
```
imurchie: Can you add a comment here describing the reason for this construct? It is non-obvious from the code, and will be so when others look at it later, outside the context of this PR.

alizelzele: @imurchie sure, done.

lib/driver.js
```diff
@@ -143,7 +143,12 @@ class AndroidDriver extends BaseDriver {
       await this.startAndroidSession(this.opts);
       return [sessionId, this.caps];
     } catch (e) {
-      await this.deleteSession();
+      // ignoring delete session exception if any and throw the real error
+      // that happened while creating the session.
+      try {
+        await this.deleteSession();
+      } catch (ignored) {
```
jlipps: for compactness I often write this as:

```js
} catch (ign) {}
```

But that's a minor stylistic issue

imurchie:
Thanks for this!

imurchie:
Published in `appium-android-driver@1.10.28`

PR #184, opened 2016-10-14
Add clearSystemFiles cap to delete generated app files

imurchie:
First pass, just deletes any copy of the app that was made.


imurchie:
Towards https://github.com/appium/appium/issues/5156

PR #186, opened 2016-10-30
add apis to read the performance data

heeseon:
the appium already has a way to gather the performance data by log.
but If the resulting data format is set to a log file, the user should conduct another parsing to show the data on the screen.
also I think user wants to decide the moment to read the data and associate the performance data and other information like test case #.
so I’d like to add apis to gather the performance data.


lib/commands/performance.js
```diff
@@ -0,0 +1,297 @@
...
+import _ from 'lodash';
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.postDismissAlert = async function () {
+  throw new errors.NotYetImplementedError();
+};
+
+const SUPPORTED_PERFORMANCE_DATA_TYPES = {
+  'cpuinfo': 'the amount of cpu by user and kernel process - cpu information for applications on real devices and simulators',
```
imurchie: For all of these keys (`cpuinfo`, `memoryinfo`, `batteryinfo`, and `networkinfo`) there is no need for quotes. The keys, as long as they are valid in JavaScript, can be bare.

lib/commands/performance.js
```diff
@@ -0,0 +1,297 @@
...
+  'networkinfo': 'the network statistics - network rx/tx information for applications on real devices and simulators'
+};
+
+commands.getPerformanceDataTypes = function () {
+  return _.keys(SUPPORTED_PERFORMANCE_DATA_TYPES);
+};
+
+commands.getPerformanceData = async function (applicationPackageName, performanceDataType, performanceDataReadTimeout) {
+
+  if (performanceDataType === "batteryinfo")
```
imurchie: The style we go by always includes braces:

```js
if (dataType === 'batteryinfo') {
  // ...
} else if (dataType === 'cpuinfo') {
  // ...
} // ...
```

lib/commands/performance.js
```diff
@@ -0,0 +1,297 @@
...
+  'memoryinfo': 'the amount of memory used by the process - memory information for applications on real devices and simulators',
+  'batteryinfo': 'the remaining battery power - battery power information for applications on real devices and simulators',
+  'networkinfo': 'the network statistics - network rx/tx information for applications on real devices and simulators'
+};
+
+commands.getPerformanceDataTypes = function () {
+  return _.keys(SUPPORTED_PERFORMANCE_DATA_TYPES);
+};
+
+commands.getPerformanceData = async function (applicationPackageName, performanceDataType, performanceDataReadTimeout) {
```
imurchie: In the context of the function, there is no need for "application" in the parameter names. It would be more succinct and readable to have

```js
commands.getPerformanceData = async function (packageName, dataType, dataReadTimeout) {
  // ...
```

lib/commands/performance.js
```diff
@@ -0,0 +1,297 @@
...
+    return this.getCPUInfo(applicationPackageName, performanceDataReadTimeout);
+  else if (performanceDataType === "memoryinfo") 
+    return this.getMemoryInfo(applicationPackageName, performanceDataReadTimeout);
+  else if (performanceDataType === "networkinfo")
+    return this.getNetworkTrafficInfo(performanceDataReadTimeout);
+  else 
+    throw new Error('No performance data of type \'' + performanceDataType + '\' found.');
+};
+
+helpers.getCPUInfo = async function (applicationPackageName, performanceDataReadTimeout) {
```
imurchie: Same vis-a-vis parameter names. `packageName` and `dataReadTimeout` would suffice.

lib/commands/performance.js
```diff
@@ -0,0 +1,297 @@
...
+  let returnValue = [];
+  returnValue[0] = ["user", "kernel"];
+
+  let cmd, data, start, end, user, kernel;
+  let loop = true;
+  let tryCount = 0;
+  cmd = ['dumpsys', 'cpuinfo', '|', 'grep', "\'" + applicationPackageName + "\'"];
```
imurchie: You should use string interpolation here

```js
`'${packageName}'`
```

I'm not sure what the escaping of the quotes is doing.

lib/commands/performance.js
```diff
@@ -0,0 +1,297 @@
...
+    
+  while ( loop === true ) {
```
imurchie: Can we get a little explanation of what this loop is doing? in a comment?

imurchie: Also, no need to check the `=== true` part

```js
while (loop) {
  // ...
```

lib/commands/performance.js
```diff
@@ -0,0 +1,297 @@
...
+
+helpers.getMemoryInfo = async function (applicationPackageName, performanceDataReadTimeout) {
+  let returnValue = [];
+  returnValue[0] = ["totalPrivateDirty", "nativePrivateDirty", "dalvikPrivateDirty", "eglPrivateDirty", "glPrivateDirty", "totalPss", "nativePss", "dalvikPss", "eglPss", "glPss", "nativeHeapAllocatedSize", "nativeHeapSize"];
+    
+  let loop = true;
+  let cmd, data, totalPrivateDirty, nativePrivateDirty, dalvikPrivateDirty, eglPrivateDirty, glPrivateDirty, totalPss, nativePss, dalvikPss, eglPss, glPss, nativeHeapSize, nativeHeapAllocatedSize;
+  let tryCount = 0;
+      
+  cmd = ['dumpsys', 'meminfo', "\'" + applicationPackageName + "\'", '|', 'grep', '-E', "\'Native|Dalvik|EGL|GL|TOTAL\'"];
```
imurchie: String interpolation here, as well.

lib/commands/performance.js
heeseon: ok, I tried, 
I have faced below issue when I try to commit the code.

hwanghuiseon-ui-MacBook-Pro:appium-android-driver hwangheeseon$ git commit -m "change to include braces"
Pre-commit checks...
[21:27:50] Using gulpfile ~/work/appium-android-driver/gulpfile.js
[21:27:50] Starting 'eslint'...
[21:28:05]
/Users/hwangheeseon/work/appium-android-driver/lib/commands/performance.js
  27:8  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style
  30:8  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style
  33:8  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style
  37:5  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style

✖ 4 problems (4 errors, 0 warnings)

[21:28:05] 'eslint' errored after 16 s
[21:28:05] ESLintError in plugin 'gulp-eslint'
Message:
    Failed with 4 errors
pre-commit:
pre-commit: We've failed to pass the specified git pre-commit hooks as the `lint`
pre-commit: hook returned an exit code (1). If you're feeling adventurous you can
pre-commit: skip the git pre-commit hooks by adding the following flags to your commit:
pre-commit:
pre-commit:   git commit -n (or --no-verify)
pre-commit:
pre-commit: This is ill-advised since the commit is broken.
pre-commit:

imurchie:
The CLA bot is having problems. Ignore it.

lib/commands/performance.js
heeseon: I applied

lib/commands/performance.js
heeseon: I applied

lib/commands/performance.js
heeseon: I applied

lib/commands/performance.js
heeseon: I applied

jlipps:
@heeseon the CLA bot is not having problems anymore, but you need to configure your git client to use your github email or username. See https://help.github.com/articles/setting-your-email-in-git/

lib/commands/performance.js
```diff
@@ -0,0 +1,339 @@
...
+import _ from 'lodash';
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.postDismissAlert = async function () {
+  throw new errors.NotYetImplementedError();
+};
+
+const SUPPORTED_PERFORMANCE_DATA_TYPES = {
+  cpuinfo : 'the amount of cpu by user and kernel process - cpu information for applications on real devices and simulators',
```
jlipps: according to our conventions, there's no space between a key and the colon, so this should read:

```js
  cpuinfo: 'the amount...',
```

lib/commands/performance.js
```diff
@@ -0,0 +1,339 @@
...
+  let cmd, data, start, end, user, kernel;
+  let loop = true;
+  let tryCount = 0;
+  cmd = ['dumpsys', 'cpuinfo', '|', 'grep', `'${packageName}'`];
+  
+  //sometimes, the function of 'adb.shell' fails. when I tested this function on the target of 'Galaxy Note5', 
+  //adb.shell(dumpsys cpuinfo) returns cpu datas for other application packages, but I can't find the data for packageName.
+  //It usually fails 30 times and success for the next time,  
+  //Since then, he has continued to succeed.
+  while ( loop ) {
```
jlipps: no need for spaces around `loop`

lib/commands/performance.js
```diff
@@ -0,0 +1,339 @@
...
+
+helpers.getBatteryInfo = async function (dataReadTimeout) {
+  let returnValue = [];
+  returnValue[0] = ["power"];
+  let cmd, data, start, end, power;
+  let loop = true;
+  let tryCount = 0;
+  cmd = ['dumpsys', 'battery', '|', 'grep', "\'level\'"];
+
+  while ( loop ) {
```
jlipps: checkout the `asyncbox` package we use elsewhere in Appium, and specifically the `retry` command. You can use it here and in other places instead of reimplementing it

jlipps: I guess I should say the `retryInterval` command

lib/commands/performance.js
```diff
@@ -0,0 +1,339 @@
...
+    else {
+      ++ tryCount;
+      data = await this.adb.shell(cmd);        
+    }
+        
+    if (data !== "" && data !== undefined && data !== null){
+      let arrayList = data.split("\n");
+      let arrayList2;  
+      for (let i=0 ; i < arrayList.length ; i++){
+        let testString = arrayList[i].replace(/\s/g, ","); // 문자열 끝 부분의 공백 제거
```
jlipps: can you translate your comments to English for us and other developers?

lib/commands/performance.js
```diff
@@ -0,0 +1,339 @@
...
+
+        if ((arrayList2[0] === "Native" || arrayList2[1] === "Native") && (arrayList2[1] === "Heap" || arrayList2[2] === "Heap")){//native heap
+          nativePrivateDirty = arrayList2[3] ;
+          nativePss = arrayList2[2];
+          nativeHeapAllocatedSize = arrayList2[6];
+          nativeHeapSize = arrayList2[8];
+        } else if ((arrayList2[0] === "Dalvik" || arrayList2[1] === "Dalvik") && (arrayList2[1] === "Heap" || arrayList2[2] === "Heap")) { //dalvik heap
+          dalvikPrivateDirty = arrayList2[4];
+          dalvikPss = arrayList2[3];
+        } else if ( (arrayList2[0] === "Dalvik" || arrayList2[1] === "Dalvik") && (arrayList2[1] === "Other" || arrayList2[2] === "Other") ) ;//dalvik others
```
jlipps: braces here are a little confusing. our style is that every `if/else` statement should have braces; no one-line if/else

lib/commands/performance.js
```diff
@@ -0,0 +1,339 @@
...
+                  operations = data.substring(delimiter+1, end).trim(); 
+                        
+                }             
+              }
+              returnValue[returnIndex ++] = [bucketStart, bucketDuration, activeTime, rxBytes, rxPackets, txBytes, txPackets, operations];      
+            }
+          }
+        }
+        
+      }
```
jlipps: this method has lots of duplication and would benefit from being rewritten in a more functional style using methods from libraries like lodash.

test/unit/commands/performance-specs.js
```diff
@@ -0,0 +1,39 @@
...
+import sinon from 'sinon';
+import Bootstrap from 'appium-android-bootstrap';
+import AndroidDriver from '../../..';
+
+let driver;
+let sandbox = sinon.sandbox.create();
+chai.should();
+chai.use(chaiAsPromised);
+
+describe('getperformancedata', () => {
```
jlipps: You have this in the unit tests, but I don't see you stubbing out the methods, which makes me think this is actually still running adb commands like an e2e test

lib/commands/performance.js
heeseon: fixed

lib/commands/performance.js
heeseon: fixed

lib/commands/performance.js
heeseon: is there any way to break the loop statement? 
for example, when I call retryInterval(dataReadTimeout, 1000, fn)
in the loop, the function succed to read capability data at the second try, so there is no need to try to call any more.
how to stop the loop?

lib/commands/performance.js
heeseon: fixed

lib/commands/performance.js
heeseon: fixed

lib/commands/performance.js
heeseon: fixed for the isXXX function

lib/commands/performance.js
```diff
@@ -0,0 +1,339 @@
+import { errors } from 'appium-base-driver';
+import { sleep } from 'asyncbox';
+import _ from 'lodash';
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.postDismissAlert = async function () {
```
imurchie: Why is this function here? This is already in `lib/commands/alert.js`.

lib/commands/performance.js
```diff
@@ -0,0 +1,339 @@
...
+};
+
+// returns the information type of the system state which is supported to read as like cpu, memory, network traffic, and battery.
+//input - (packageName) the package name of the application 
+//        (dataType) the type of system state which wants to read. It should be one of the keys of the SUPPORTED_PERFORMANCE_DATA_TYPES
+//        (dataReadTimeout) the number of attempts to read
+// output - table of the performance data, The first line of the table represents the type of data. The remaining lines represent the values of the data.
+//
+// in case of battery info : [[power], [23]]
+// in case of memory info : [[totalPrivateDirty, nativePrivateDirty, dalvikPrivateDirty, eglPrivateDirty, glPrivateDirty, totalPss, nativePss, dalvikPss, eglPss, glPss, nativeHeapAllocatedSize, nativeHeapSize], [18360, 8296, 6132, null, null, 42588, 8406, 7024, null, null, 26519, 10344]]
```
imurchie: If I'm reading this correctly, this returns a "table", which is an array of arrays, in which the first array is keys and the second array is the corresponding values? So for memory info it gives

```
[
  [totalPrivateDirty, nativePrivateDirty, dalvikPrivateDirty, eglPrivateDirty, glPrivateDirty, totalPss, nativePss, dalvikPss, eglPss, glPss, nativeHeapAllocatedSize, nativeHeapSize], 
  [18360, 8296, 6132, null, null, 42588, 8406, 7024, null, null, 26519, 10344]
]
```

Is there no way to make it as associative array? Since this is an entirely new API, why not have it as simple to use as possible. So the above would be

```
{
  totalPrivateDirty: 18360, 
  nativePrivateDirty: 8296, 
  dalvikPrivateDirty: 6132, 
  eglPrivateDirty: null, 
  glPrivateDirty: null, 
  totalPss: 42588, 
  nativePss: 8406, 
  dalvikPss: 7024, 
  eglPss: null, 
  glPss: null, 
  nativeHeapAllocatedSize: 26519, 
  nativeHeapSize: 10344
}
```

heeseon: ok, I understand what you mean. 

the reason why I defined the return type like that 

1. 
for memory, cpu, and battery, there could have a value for a key.
but for network, there could have many values for a key.
So, There is the key duplication.

2. 
User can use the result data as the type of Array in Java.
so, user does not need to add the library for the type of json.

3. 
for the network statistics,
The key is different between real device and emulator.
I have not tested many devices and emulator, I think there might be some device or emulator which has the different key. 
so, I made the key as variable not constant.
I think It is easy to manage the code for like this variations.

I thank you for your recommendation.
In Korea, It is difficult to get the development culture like this.
Anyway, If you still think that you prefer the json type than the Array, then I will change the code.

imurchie: I guess that makes some sense.

lib/commands/performance.js
jlipps: the way that `retryInterval` works is, the first time `fn` returns something rather than throws, it will stop looping. So that is how you "break".

heeseon: applied

heeseon:
What I have to do more?

imurchie:
One last thing: Can you squash into a single commit?

heeseon:
Sorry, I can't understand what you mean. Could you explain it more in detail?

imurchie:
I mean that we would like this functionality to be in a single commit, rather than 24. Git provides functionality to "squash" the commits so that we do not have irrelevant things in the commit history.

You can google it. This is a good introduction: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

heeseon:
Thank you for your answer. I understand what you mean. I can do it 9 hours later.

heeseon:
Is something wrong to squash merge in github?

imurchie:
Did doing it manually not work? It is generally better in our experience to give the author the control. I can do it if you don't want to.

heeseon:
I need history that I did contribute the open source. If you do it as my account, so I have the history, than please help me. 
But I still have to contribute something more to appium like crash report, please guide me to do myself it next time.

Thank you!!!

imurchie:
Whatever we do, you will be in the history. We want that as much as you!
And we appreciate all your work

heeseon:
then now am I contributor for appium-android-driver ? 

How I check it? I can't check my name both in master branch and in contributor list ?

imurchie:
You can see the contributors here: https://github.com/appium/appium-android-driver/graphs/contributors

PR #187, opened 2016-10-31
fix incorrect parameter key in installApp (fix appium/appium#6695)

jlipps:
cc @imurchie

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -102,8 +102,8 @@ commands.installApp = async function (appPath) {
 
   let {apkPackage} = await this.adb.packageAndLaunchActivityFromManifest(appPath);
   let opts = {
-    localApkPath: appPath,
-    pkg: apkPackage,
+    app: appPath,
+    appPackage: apkPackage,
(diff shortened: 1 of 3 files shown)


lib/commands/general.js
```diff
@@ -102,7 +102,7 @@ commands.installApp = async function (appPath) {
 
   let {apkPackage} = await this.adb.packageAndLaunchActivityFromManifest(appPath);
   let opts = {
-    localApkPath: appPath,
+    app: appPath,
     pkg: apkPackage,
```
imurchie: This looks like it is looking for `appPackage`, not `pkg`.

https://github.com/appium/appium-android-driver/blob/c5c8c1bb0b855283c83c760e41a91ca03b003204/lib/android-helpers.js#L261

jlipps: right you are! fixed

imurchie:
If you rebase with master the CI troubles should be gone.

imurchie:
Published in `appium-android-driver@1.10.31`.

PR #192, opened 2016-11-12
Add device information commands

midopa:
- Get display density
- Get visibility and bounds of status bar and navigation bars
- Check if soft keyboard is present

I think this will also require updates in other packages to expose the commands/endpoints to users.

Since the commands aren't exposed by default, I tested by replacing the internals of `hideKeyboard` before moving the logic to their own commands. Tested on emulators and a device.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -46,6 +46,11 @@ commands.back = function () {
   return this.bootstrap.sendAction('pressBack');
 };
 
+commands.isKeyboardShown = async function () {
+  let keyboardInfo = await this.adb.isSoftKeyboardPresent();
+  return keyboardInfo.isKeyboardShown;
+};
+
 commands.hideKeyboard = async function () {
   let {isKeyboardShown, canCloseKeyboard} = await this.adb.isSoftKeyboardPresent();
   if (!isKeyboardShown) {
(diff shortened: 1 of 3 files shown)


lib/commands/general.js
```diff
@@ -216,6 +221,105 @@ commands.closeApp = async function () {
...
+  }
+  log.errorAndThrow('Failed to get display density property.');
+};
+
+/**
+ * Parses the given window manager Surface string to get info.
+ * @param line: To parse. This is assumed to be valid.
+ * @return: Visibility and bounds of the Surface.
+ */
+helpers.parseSurfaceLine = function (line) {
```
imurchie: Unless there is a reason for this to be on the object (like sub-classes use it?), I'd rather this be internal to this module (`function parseSurfaceLine` instead of `helpers.parseSurfaceLine`). We have altogether too many methods just dangling off the object that are not necessary.

lib/commands/general.js
```diff
@@ -216,6 +221,105 @@ commands.closeApp = async function () {
...
+  let result = {
+    visible: false,
+    x: 0,
+    y: 0,
+    width: 0,
+    height: 0
+  };
+  result.visible = (line.indexOf('shown=true') !== -1);
```
imurchie: Can this not happen in the initialization? 

```js
let result = {
  visible: (line.indexOf('shown=true') !== -1),
  x: 0,
  y: 0,
  width: 0,
  height: 0
};
```

lib/commands/general.js
```diff
@@ -216,6 +221,105 @@ commands.closeApp = async function () {
...
+
+  // the surface bounds are in the format:
+  // "rect=(0.0,1184.0) 768.0 x 96.0"
+  //       ^ location   ^ size
+  // cut out the stuff before the 'rect' and then split the numbers apart
+  let bounds = line.split('rect=')[1]
+    .replace(/[\(\), x]+/g, ' ')
+    .trim()
+    .split(' ');
+  result.x = parseFloat(bounds[0]);
```
imurchie: All of these, too, I suppose.

lib/commands/general.js
```diff
@@ -216,6 +221,105 @@ commands.closeApp = async function () {
...
+
+  return result;
+};
+
+/**
+ * Extracts status and navigation bar information from the window manager output.
+ * @param lines: Output from dumpsys command
+ * @return: Visibility and bounds info of status and navigation bar
+ */
+helpers.parseWindows = function (lines) {
```
imurchie: This method does not seem to need to be an exposed method, either.

lib/commands/general.js
midopa: How would I get access to the function for unit testing?

imurchie: The easiest way is to export it from the file.

midopa:
Updated

lib/commands/general.js
```diff
@@ -216,6 +221,100 @@ commands.closeApp = async function () {
   await this.adb.forceStop(this.opts.appPackage);
 };
 
+commands.getDisplayDensity = async function () {
+  let out = await this.adb.shell(['getprop', 'ro.sf.lcd_density']);
+  if (out) {
+    return parseInt(out);
```
imurchie: There is a potential issue with `parseInt` when there is no radix specified. It ought to be added

```js
return parseInt(out, 10);
```

lib/commands/general.js
```diff
@@ -216,6 +221,100 @@ commands.closeApp = async function () {
...
+  }
+  out = await this.adb.shell(['getprop', 'qemu.sf.lcd_density']);
+  if (out) {
+    return parseInt(out);
```
imurchie: Same.

lib/commands/general.js
midopa: What would you like to happen if the parse result is `NaN`? Should an error be thrown?

midopa: Nvm, I think I'll just debug log it and let it flow down to the last `log.errorAndThrow` in the case of NaNs.

midopa:
Updated

lib/commands/general.js
```diff
@@ -216,6 +221,111 @@ commands.closeApp = async function () {
...
 };
 
+commands.getDisplayDensity = async function () {
+  let out = await this.adb.shell(['getprop', 'ro.sf.lcd_density']);
+  if (out) {
+    let val = parseInt(out, 10);
+    // if the value is NaN, try getting the emulator property
+    if (!isNaN(val)) {
+      return val;
+    } else {
```
imurchie: This `else` can be removed, now, since you're returning.

```js
let val = parseInt(out, 10);
if (!isNaN(val)) {
  return val;
}
// if the value is NaN, try getting the emulator property
log.debug(`Parsed density value was NaN: "${out}"`);
```

midopa:
Updated

imurchie:
Looks good to me now! Thanks for your patience with going through it.

midopa:
No probs! Hopefully my commit-amend and force-push method didn't make it harder to review? It's just the general process I follow unless there's a really big change.

imurchie:
No, that's fine for the most part. That's how I do it too, so rebasing and squashing is not an issue down the line.

PR #193, opened 2016-11-18
Allow way to pass chromeAndroidPackage

yaphatak:
This is useful when you are launching an app from another app.
Fixes https://github.com/appium/appium/issues/7194

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -214,6 +214,9 @@ async function setupNewChromedriver (opts, curDeviceId, adbPort) {
   if (opts.chromeUseRunningApp) {
     caps.chromeOptions.androidUseRunningApp = opts.chromeUseRunningApp;
   }
+  if (opts.chromeAndroidPackage) {
+    caps.chromeOptions.androidPackage = opts.chromeAndroidPackage;
+  }
   if (opts.chromeAndroidActivity) {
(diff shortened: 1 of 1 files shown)


imurchie:
Looks good! Can you sign the CLA (again, unfortunately, but since joining the JS Foundation we have to use their system).

yaphatak:
@imurchie I realized that. I am already following with our internal advisor on this.

Also, don't you think we should provide way to pass an `optional` parameter for `setContext` to pass `androidPackage` for `android` and `bundleId` for `ios`. Because today we can do multiple app testing in both `android` and `ios`

imurchie:
That is probably a good idea. It's a method from the Selenium spec, so things would have to be changed there, though.

yaphatak:
@imurchie Yes and that's why I added @jlipps in this conversation :)

yaphatak:
@imurchie signed

jlipps:
@yaphatak I'm a bit confused. when you get contexts, does the one you want not appear in the list?

imurchie:
Published in `appium-android-driver@1.10.35`.

yaphatak:
@jlipps I just saw this. I think its good to have `chromeAndroidProcess ` and it looks like it is already supported now. For `chromeAndroidProcess `  we still have to run some command to get coorect process. Today we get context names as `webview_pkgname`. So I am thinking extract pkg name from context anme

PR #198, opened 2016-12-09
added grant permissions

sravanmedarapu:
#https://github.com/appium/appium-adb/pull/189

#https://github.com/appium/appium/issues/7497

@imurchie, @jlipps

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -271,12 +271,14 @@ class AndroidDriver extends BaseDriver {
       if (this.opts.fastReset) {
         await helpers.resetApp(this.adb, this.opts.app, this.opts.appPackage, this.opts.fastReset);
       }
+      await this.grantPermissions();
       return;
     }
     if (!this.opts.skipUninstall) {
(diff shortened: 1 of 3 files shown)


lib/driver.js
```diff
@@ -271,12 +271,26 @@ class AndroidDriver extends BaseDriver {
       if (this.opts.fastReset) {
         await helpers.resetApp(this.adb, this.opts.app, this.opts.appPackage, this.opts.fastReset);
       }
+      if (this.opts.autoGrantPermissions) {
+        try {
+          await this.adb.grantAllPermissions(this.opts.appPackage, this.opts.app);
```
jlipps: does this even make sense? we know in this case that `this.opts.app` is not defined. does it make sense to grant permissions in that case?

jlipps: and if it _does_ make sense, I don't like the duplicated code. we should factor out into a function and reuse.

jlipps:
published in 1.10.38

PR #200, opened 2016-12-30
add chromeAndroidProcess support

seveniruby:
try to support wechat webview automation
detail https://testerhome.com/topics/6954

a demo for weixin (wechat) testcase
already passed by local testing

```scala
  test("test weixin h5") {
    val capability = new DesiredCapabilities()
    capability.setCapability("app", "")
    capability.setCapability("appPackage", "com.tencent.mm")
    capability.setCapability("appActivity", ".ui.LauncherUI")
    capability.setCapability("deviceName", "emulator-5554")
    capability.setCapability("fastReset", "false")
    capability.setCapability("fullReset", "false")
    capability.setCapability("noReset", "true")
    capability.setCapability("resetKeyboard", "true")
    capability.setCapability("automationName", "appium")
    capability.setCapability("platformName", "android")
    capability.setCapability("chromeAndroidProcess", "com.tencent.mm:tools")

    val url = "http://127.0.0.1:4723/wd/hub"
    val driver = new AndroidDriver[WebElement](new URL(url), capability)
    driver.findElementByXPath("//*[@text='我']").click
    driver.findElementByXPath("//*[@text='收藏']").click
    driver.findElementByXPath("//*[contains(@text, '美团外卖')]").click
    println(driver.getPageSource)
    println(driver.getContextHandles)
    driver.context("WEBVIEW_com.tencent.mm:tools")
    println(driver.getPageSource)
  }

```

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -220,6 +220,9 @@ async function setupNewChromedriver (opts, curDeviceId, adbPort) {
   if (opts.chromeAndroidActivity) {
     caps.chromeOptions.androidActivity = opts.chromeAndroidActivity;
   }
+  if (opts.chromeAndroidProcess) {
+    caps.chromeOptions.androidProcess = opts.chromeAndroidProcess;
+  }
   if (opts.enablePerformanceLogging) {
     caps.loggingPrefs = {performance: 'ALL'};
   }


chenhengjie123:
That's awesome! Wechat is very popular in China, but due to it use Tencent's own webview on android platform, it cannot be automated like other webviews. It would be very nice if appium can support it!

jlipps:
thanks @seveniruby! question though: could you do this before by passing in the `chromeOptions` capability?

jlipps:
published in 1.10.39

seveniruby:
i found this in appium changelog 
```
Add chromeAndroidPackage capability which will be passed to chromeOptions
```
so i just add another chromeAndroidProcess 

i had test with ChromeOption by this code

```scala
    val options = new ChromeOptions()
    options.setExperimentalOption("androidPackage", "com.tencent.mm")
    options.setExperimentalOption("androidUseRunningApp", true)
    options.setExperimentalOption("androidActivity", ".plugin.webview.ui.tools.WebViewUI")
    options.setExperimentalOption("androidProcess", "com.tencent.mm:tools")
    capability.setCapability(ChromeOptions.CAPABILITY, options)

```
this not work. 

i think we should support ChromeOptions for standard. this will be same to WebDriver

jlipps:
@seveniruby thanks for trying that out. I think you're right. I'm not sure why we don't support the regular chromeOptions object, as I thought we did, and then just overrode certain fields on it when necessary

seveniruby:
sorry for that , my origin testcase result was wrong just because my env
var have a http_proxy.
chromedriver will apply the proxy setting when access http://127.0.0.1:8000,
so the result was wrong.

when i clear the http_proxy env var, the testcase just passed.
you can revert the my pr, and remove code with androidProcess and
androidPackage support.

for weixin or wechat webview automation just need add such code

```

val options = new ChromeOptions()
options.setExperimentalOption("androidProcess", "com.tencent.mm:tools")
capability.setCapability(ChromeOptions.CAPABILITY, options)

```


2017-01-05 8:01 GMT+08:00 Jonathan Lipps <notifications@github.com>:

jlipps:
@seveniruby ah ok makes sense, thanks for clarifying. i did recall that we had made the chromeoptions accessible!

PR #203, opened 2017-01-13
Capability to extract chromeAndroidPackagee from context_name

yaphatak:
This will allow us to to connect to multiple apps with webivew in same test session.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -113,6 +113,12 @@ helpers.startChromedriverProxy = async function (context) {
   } else {
     let opts = _.cloneDeep(this.opts);
     opts.chromeUseRunningApp = true;
+    if (opts.extractChromeAndroidPackageFromContextName) {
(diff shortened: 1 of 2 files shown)


imurchie:
What happens if this is always on? Does it affect the functioning of webviews?

Also, if this is a capability, it also needs to be added to https://github.com/appium/appium-android-driver/blob/master/lib/desired-caps.js

jlipps:
So @yaphatak if I understand your change correctly, you want people to be able to do something like this:

```
driver.setContext("WEBVIEW_com.android.foo")
```

even if `com.android.foo` is not the currently running app?

yaphatak:
That is correct. That's why this wont be default behavior. We can do same using `chromeAndroidProcess` but then you need to know `processid`

imurchie:
Excellent! Thanks!

imurchie:
Published in `appium-android-driver@1.11.0`

PR #204, opened 2017-01-13
configure app before installing it

yaphatak:


imurchie:
Looks good to me. Thanks!

PR #206, opened 2017-01-17
Update to pass on different API levels

imurchie:
Quarantine a number of tests (see https://github.com/appium/appium/issues/7707).

Fix performance API and its tests.

### DIFF
--- a/test/functional/commands/find/by-uiautomator-e2e-specs.js
+++ b/test/functional/commands/find/by-uiautomator-e2e-specs.js
@@ -67,7 +67,10 @@ describe('Find - uiautomator', function () {
     let el = await driver.findElOrEls('-android uiautomator', 'new UiSelector().clickable(true).className(android.widget.TextView).index(1)', false);
     await driver.getText(el.ELEMENT).should.eventually.equal('Accessibility');
   });
-  it('should find an element with recursive UiSelectors', async () => {
+  it('should find an element with recursive UiSelectors', async function () {
+    // TODO: figure out why this fails with 7.1.1
+    if (await driver.adb.getApiLevel() >= '24') return this.skip();
+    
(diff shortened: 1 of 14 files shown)


test/functional/commands/find/by-uiautomator-e2e-specs.js
```diff
@@ -67,7 +67,9 @@ describe('Find - uiautomator', function () {
     let el = await driver.findElOrEls('-android uiautomator', 'new UiSelector().clickable(true).className(android.widget.TextView).index(1)', false);
     await driver.getText(el.ELEMENT).should.eventually.equal('Accessibility');
   });
-  it('should find an element with recursive UiSelectors', async () => {
+  it('should find an element with recursive UiSelectors', async function () {
+    // TODO: figure out why this fails with 7.1.1
+    if (await driver.adb.getApiLevel() >= '24') this.skip();
```
jlipps: from mocha docs: 

> Best practice: To avoid confusion, do not execute further instructions in a test or hook after calling this.skip().

I looked this up because I was indeed confused whether `skip()` altered control flow or not. Looks like it does! Maybe a `return this.skip()` just to be very clear?

test/functional/commands/url-e2e-specs.js
```diff
@@ -1,29 +1,54 @@
...
+
     driver = new AndroidDriver();
     await driver.createSession(defaultCaps);
   });
   after(async () => {
     await driver.deleteSession();
   });
+
   it('should be able to start a data uri via setUrl', async () => {
+    if (defaultCaps.browserName >= 'Chrome') {
```
jlipps: it's a little opaque to me what the `>=` comparison is meant to do

imurchie: Yes, that was a typo (I did all this on the plane, and it must have slipped through).

PR #209, opened 2017-02-05
Refactor unlock

vrunoa:
In order for the unlock to work on Android Emulators 2.0 and 6.0+ devices I refactor the android-driver to use specific methods and deprecate the unlock app using some new capabilities:

```json
{
"unlockType": "pin",
"unlockKey": "111"
}
```

# options
* unlockType: ["pin", "password", "pattern"]

In case of having a device locked with pattern, we set  the device pattern buttons as number of a regular phone
 
| • • • | --->  | 1 2 3 |
| • • • | ---> | 4 5 6 |
| • • • | ---> | 7 8 9 |

For ex.: The unlock key for unlockKey this pattern would be *"159874236"*

Please review @jlipps @imurchie @mykola-mokhnach
[https://github.com/appium/appium/issues/7696](https://github.com/appium/appium/issues/7696)


lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+    await driver.hideKeyboard();
+    await sleep(100);
+  }
+  // dismiss notifications
+  await adb.shell(["service", "call", "notification", "1"]);
+  await adb.back();
+  if (await adb.getApiLevel() > 21) {
+    await adb.shell(["wm", "dismiss-keyguard"]);
+  } else {
+    logger.info("swipeUp");
```
mykola-mokhnach: can we provide better log description?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+  let isKeyboardShown = await driver.isKeyboardShown();
+  if (isKeyboardShown) {
+    await driver.hideKeyboard();
+    await sleep(100);
+  }
+  // dismiss notifications
+  await adb.shell(["service", "call", "notification", "1"]);
+  await adb.back();
+  if (await adb.getApiLevel() > 21) {
+    await adb.shell(["wm", "dismiss-keyguard"]);
```
mykola-mokhnach: fast return would look better here than using _else_ (and will save us some lines of code):

```
...
if (await adb.getApiLevel() > 21) {
  await adb.shell(["wm", "dismiss-keyguard"]);
  return;
}
await helpers.swipeUp(driver);
```

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+  } else {
+    logger.info("swipeUp");
+    await helpers.swipeUp(driver);
+  }
+};
+
+helpers.swipeUp = async function (driver) {
+  let windowSize = await driver.getWindowSize();
+  let x0 = parseInt(windowSize.x/2);
```
mykola-mokhnach: please format

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+  let y0 = windowSize.y - 10;
```
mykola-mokhnach: how these magic numbers are calculated?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+  logger.info(`Trying to unlock device using pin ${capabilities.unlockKey}`);
+  await helpers.dismissKeyguard(driver, adb);
+  let key = capabilities.unlockKey.split("");
+  if (await adb.getApiLevel() >= 21) {
+    let els = await driver.findElOrEls("id", "com.android.systemui:id/digit_text", true);
+    if (els === null || els.length === 0) {
+      throw new Error("Error finding unlock pin buttons!");
+    }
+    let pins = {};
+    for (let e in els) {
```
mykola-mokhnach:
```javascript
let e of els
```

 might be a better solution here. You could optimise the other loops in this function as well

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+      let el = await driver.findElOrEls("id", `com.android.keyguard:id/key${pin}`, false);
+      if (el === null) {
+        throw new Error("Error finding unlock pin button!");
+      }
+      await driver.click(el.ELEMENT);
+    }
+    let el = await driver.findElOrEls("id", "com.android.keyguard:id/key_enter", false);
+    await driver.click(el.ELEMENT);
+  }
+  await sleep(100);
```
mykola-mokhnach: Why it is necessary to sleep?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+
+helpers.passwordUnlock = async function (adb, driver, capabilities) {
+  logger.info(`Trying to unlock device using password ${capabilities.unlockKey}`);
+  await helpers.dismissKeyguard(driver, adb);
+  let key = capabilities.unlockKey;
+  // Replace blank spaces with %s
+  key = helpers.encodePassword(key);
+  // Why adb ? It was less flaky
+  await adb.shell(["input", "text", key]);
+  await sleep(100);
```
mykola-mokhnach: This magic sleep delay can be extracted to a constant

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+  await adb.shell(["input", "keyevent", KEYCODE_NUMPAD_ENTER]);
+  await sleep(100);
+};
+
+helpers.getPatternKeyPosition = function (key, pos0, piece) {
+  let xPos = (key, x, piece) => {
+    return Math.round(x + ((key % 3) || 3) * piece - piece/2);
```
mykola-mokhnach: It would be better to have these magic numbers (9 and 3)  as constants

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+  let keys = capabilities.unlockKey.split("");
+  /* We set the device pattern buttons as number of a regular phone
+   *  | • • • |     | 1 2 3 |
+   *  | • • • | --> | 4 5 6 |
+   *  | • • • |     | 7 8 9 |
+  */
+  // The pattern view buttons is not seeing by the uiautomator, so we are going to
+  // try clicking on the buttons using the parent view bounds and math.
+  let el;
+  if (await adb.getApiLevel() >= 21) {
```
mykola-mokhnach: This whole condition could be replaced with

```javascript
let apiLevel = await adb.getApiLevel();
let el = await driver.findElOrEls("id", `com.android.${apiLevel >= 21 ? 'systemui' : 'keyguard'}:id/lockPatternView`, false);
```

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+    el = await driver.findElOrEls("id", "com.android.systemui:id/lockPatternView", false);
+  } else {
+    el = await driver.findElOrEls("id", "com.android.keyguard:id/lockPatternView", false);
+  }
+  let pos0 = await driver.getLocation(el.ELEMENT);
+  let size = await driver.getSize(el.ELEMENT);
+  let piece = size.width/3;
+  let actions = [];
+  let lastPos;
+  for (let i in keys) {
```
mykola-mokhnach: This an be optimised like:

```
if (keys.length < 2) {
  throw new Error(....);
}
let keyPos = helpers.getPatternKeyPosition(keys[0], pos0, piece);
actions.push({action: 'press', options: {element: null, x: keyPos.x, y: keyPos.y}});
for (keyPos of keys.slice(1)) {
  let moveTo = {x:0, y:0};
  ...
}
```

lib/unlock-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+    let keyPos = helpers.getPatternKeyPosition(keys[i], pos0, piece);
+    if (parseInt(i) === 0) {
+      actions.push({action: 'press', options: {element: null, x: keyPos.x, y: keyPos.y}});
+      lastPos = keyPos;
+      continue;
+    }
+    let moveTo = {x:0, y:0};
+    if (keyPos.x > lastPos.x) moveTo.x = piece;
```
mykola-mokhnach: Please avoid bracketless if for one-liners. See https://github.com/appium/appium/blob/master/docs/en/contributing-to-appium/style-guide-2.0.md for more details

test/unit/driver-specs.js
```diff
@@ -53,6 +53,59 @@ describe('driver', () => {
...
       sandbox.restore();
     });
+    it('should verify the unlock types', async () => {
+      await driver.createSession({platformName: 'Android', deviceName: 'device', app: '/path/to/some.apk'});
+      driver.unlocker.isValidUnlockType('pin').should.equal(true);
+      driver.unlocker.isValidUnlockType('pattern').should.equal(true);
+      driver.unlocker.isValidUnlockType('password').should.equal(true);
+      driver.unlocker.isValidUnlockType('fingerprint').should.equal(false);
+    });
+    it('should verify the unlock keys for eahc type', async () => {
```
mykola-mokhnach: typo here

mykola-mokhnach:
I suppose these changed will also require making a separate documentation entry.

lib/android-helpers.js
```diff
@@ -372,32 +373,57 @@ helpers.pushStrings = async function (language, adb, opts) {
...
+  let unlockType = capabilities.unlockType;
+  if (!unlocker.isValidUnlockType(unlockType)) {
+    throw new Error(`Invalid unlock type ${unlockType}`);
+  }
+  let unlockKey = capabilities.unlockKey;
+  if (!unlocker.isValidKey(unlockType, unlockKey)) {
+    throw new Error(`Missing unlockKey ${unlockKey} capability for unlockType ${unlockType}`);
+  }
+  let unlockMethod;
+  if (unlockType === unlocker.PIN_UNLOCK) {
```
mykola-mokhnach: UPD: It seems like JS can also handle string constants as _switch_ arguments: http://stackoverflow.com/questions/2896626/switch-statement-for-string-matching-in-javascript

vrunoa: @mykola-mokhnach sure, let's use constant number and export them in the appium module

lib/driver.js
```diff
@@ -51,6 +51,13 @@ class AndroidDriver extends BaseDriver {
     this.chromedriver = null;
     this.apkStrings = {};
     this.bootstrapPort = opts.bootstrapPort || DEVICE_PORT;
+    this.unlocker = helpers.unlocker;
+
+    // Move this up to do constructor since we are going to need
```
vrunoa: No, it's a comment because this was outside of the driver class itself. I can remove the comment I just leave it there for review.

lib/unlock-helpers.js
vrunoa: @mykola-mokhnach sorry, what do you mean by format ?

lib/driver.js
mykola-mokhnach: I think it would be better to remove the comment.

lib/unlock-helpers.js
mykola-mokhnach: -> let x0 = parseInt(windowSize.x / 2);
Code style guide says there should be spaces between operands.

lib/unlock-helpers.js
```diff
@@ -0,0 +1,185 @@
...
+    await adb.shell(["wm", "dismiss-keyguard"]);
+    return;
+  }
+  logger.info("Swiping up to dismiss keyguard");
+  await helpers.swipeUp(driver);
+};
+
+helpers.swipeUp = async function (driver) {
+  let windowSize = await driver.getWindowSize();
+  let x0 = parseInt(windowSize.x / 2);
```
jlipps: one more nit for `parseInt`, we always explicitly specify a base, e.g., `parseInt(windowSize.x / 2, 10);`

lib/unlock-helpers.js
jlipps: yes, this should be `of`, not `in`

jlipps: e.g.:

```js
for (let e of els) {
  let text = await driver.getAttribute("text", e.ELEMENT);
  ...
}
```

lib/unlock-helpers.js
jlipps: yeah, `sleep`s warrant a comment

lib/unlock-helpers.js
```diff
@@ -0,0 +1,185 @@
...
+      }
+      await driver.click(el.ELEMENT);
+    }
+    let el = await driver.findElOrEls("id", "com.android.keyguard:id/key_enter", false);
+    await driver.click(el.ELEMENT);
+  }
+  await sleep(100);
+};
+
+helpers.passwordUnlock = async function (adb, driver, capabilities) {
```
jlipps: no real reason to pass all the caps in here, right? when all we need is `unlockKey`?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,185 @@
...
+  // try clicking on the buttons using the parent view bounds and math.
+  let el;
+  if (await adb.getApiLevel() >= 21) {
+    el = await driver.findElOrEls("id", "com.android.systemui:id/lockPatternView", false);
+  } else {
+    el = await driver.findElOrEls("id", "com.android.keyguard:id/lockPatternView", false);
+  }
+  let pos0 = await driver.getLocation(el.ELEMENT);
+  let size = await driver.getSize(el.ELEMENT);
+  let piece = size.width/3;
```
jlipps: `size.width / 3`; i'm surprised running `gulp lint` isn't catching this.

lib/unlock-helpers.js
vrunoa: Using adb shell was less flaky that using the android driver to send keys, but even this method was still a bit flaky. The sleep helps. Is this good for adding as comment?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,194 @@
...
+  let apiLevel = await adb.getApiLevel();
+  let el = await driver.findElOrEls("id", `com.android.${apiLevel >= 21 ? 'systemui' : 'keyguard'}:id/lockPatternView`, false);
+  let pos0 = await driver.getLocation(el.ELEMENT);
+  let size = await driver.getSize(el.ELEMENT);
+  let piece = size.width/3;
+  let actions = [];
+  let lastPos;
+  for (let key of keys) {
+    let keyPos = helpers.getPatternKeyPosition(key, pos0, piece);
+    if (parseInt(key, 10) === 0) {
```
mykola-mokhnach: I don't think this is going to work the way you expect.

lib/unlock-helpers.js
```diff
@@ -0,0 +1,194 @@
...
+
+helpers.isValidKey = function (type, key) {
+  if (typeof key === 'undefined') {
+    return false;
+  }
+  if (type === PIN_UNLOCK) {
+    return /^[0-9]+$/.test(key.trim());
+  }
+  if (type === PATTERN_UNLOCK) {
+    if (!/^[0-9]+$/.test(key.trim())) {
```
mykola-mokhnach: is zero a valid value for pattern?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,194 @@
...
+  let actions = [
+    {action: 'press', options: {element: null, x: x0, y: y0}},
+    {action: 'moveTo', options: {element: null, x: x0, y: yP}},
+    {action: 'release'}
+  ];
+  await driver.performTouch(actions);
+};
+
+helpers.encodePassword = function (key) {
+  return key.replace(/\s+/ig, "%s");
```
mykola-mokhnach: Does it work as expected for multiple spaces not separated by other chars?

```
"   " -> "%s" 
or
"   " -> "%s%s%s"
```

vrunoa: The result should be `"   " -> "%s%s%s"`. Let me add an unit test for that.

lib/unlock-helpers.js
vrunoa: Nope, you're right. It should be `[1-9]`

lib/unlock-helpers.js
vrunoa: Yeap. I have an undefined as the last element. Going back to use `in`

lib/unlock-helpers.js
```diff
@@ -0,0 +1,197 @@
...
+    let els = await driver.findElOrEls("id", "com.android.systemui:id/digit_text", true);
+    if (els === null || els.length === 0) {
+      throw new Error("Error finding unlock pin buttons!");
+    }
+    let pins = {};
+    for (let e of els) {
+      let text = await driver.getAttribute("text", e.ELEMENT);
+      pins[text] = e;
+    }
+    for (let i in keys) {
```
jlipps: i really don't like `in`; as a project we have a policy of not using it. if straight-up `of` doesn't work, then what about something like:

```
for (let [i, key] of _.pairs(keys)) {
```

?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,197 @@
...
+   *  | • • • |     | 7 8 9 |
+
+  The pattern view buttons is not seeing by the uiautomator, so we are going to
+  try clicking on the buttons using the parent view bounds and math.
+  */
+  let apiLevel = await adb.getApiLevel();
+  let el = await driver.findElOrEls("id", `com.android.${apiLevel >= 21 ? 'systemui' : 'keyguard'}:id/lockPatternView`, false);
+  let pos0 = await driver.getLocation(el.ELEMENT);
+  let size = await driver.getSize(el.ELEMENT);
+  let piece = size.width/3;
```
jlipps: `size.width / 3`

lib/android-helpers.js
```diff
@@ -372,32 +373,54 @@ helpers.pushStrings = async function (language, adb, opts) {
...
+  let unlockKey = capabilities.unlockKey;
+  if (!unlocker.isValidKey(unlockType, unlockKey)) {
+    throw new Error(`Missing unlockKey ${unlockKey} capability for unlockType ${unlockType}`);
+  }
+  const unlockMethod = {
+    [PIN_UNLOCK]: unlocker.pinUnlock,
+    [PASSWORD_UNLOCK]: unlocker.passwordUnlock,
+    [PATTERN_UNLOCK]: unlocker.patternUnlock,
+  }[unlockType];
+  await unlockMethod(adb, driver, capabilities);
```
imurchie: I agree with this. Definitely should be wary of passing caps around.

lib/android-helpers.js
```diff
@@ -372,32 +373,54 @@ helpers.pushStrings = async function (language, adb, opts) {
...
+
+helpers.unlock = async function (driver, adb, capabilities) {
   if (!(await adb.isScreenLocked())) {
     logger.info("Screen already unlocked, doing nothing");
     return;
   }
-  logger.info("Unlocking screen");
-
-  await retryInterval(10, 1000, async () => {
+  await retryInterval(1, 1000, async () => {
```
imurchie: I don't think `retryInterval` with a retry number of `1` adds anything useful (i.e., it executes the function once and exits).

lib/android-helpers.js
```diff
@@ -372,32 +373,54 @@ helpers.pushStrings = async function (language, adb, opts) {
...
-      flags: "0x10200000",
-      stopApp: false
-    };
-    await adb.startApp(startOpts);
-    await adb.startApp(startOpts);
-
     // check if it worked, twice
+    if (typeof capabilities.unlockType === 'undefined') {
+      // Leave the old unlock to avoid breaking existing tests
+      logger.warn("Using app unlock, this is going to be deprecated!");
```
imurchie: Is this deprecated? Or pending deprecation? @jlipps shall we just deprecate it now?

lib/commands/actions.js
```diff
@@ -119,7 +119,7 @@ commands.isLocked = async function () {
 };
 
 commands.unlock = async function () {
-  return await androidHelpers.unlock(this.adb);
+  return await androidHelpers.unlockWithHelperApp(this.adb);
```
imurchie: Does this mean running the `unlock` command mid-session cannot use the other methods? Might it not be possible to allow it, if they are passed in through the caps at the beginning of the session?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,197 @@
...
+    return false;
+  }
+  if (type === PIN_UNLOCK) {
+    return /^[0-9]+$/.test(key.trim());
+  }
+  if (type === PATTERN_UNLOCK) {
+    if (!/^[1-9]+$/.test(key.trim())) {
+      return false;
+    }
+    return !(/([0-9]).*?\1/.test(key.trim()));
```
imurchie: Why does one range start with 0 and one with 1?

lib/android-helpers.js
vrunoa: @imurchie should we leave that on 10 tries ?

lib/unlock-helpers.js
vrunoa: My bad. Both times should be `[1-9]`

lib/android-helpers.js
vrunoa: Pending deprecation, I dont want to brake any test working on real devices using 5.1 or less, or using emulators 1.0

lib/android-helpers.js
imurchie: It depends on the flakiness, really. But if 1 is enough for it to work, the `retryInterval` ought to be removed.

lib/android-helpers.js
imurchie: If it is deprecated it'll still be there, until it is removed. It just is not the suggested way.

lib/unlock-helpers.js
vrunoa: I use `in` but it failed since it was a string and when I use `of` it add an undefined method at the end of the list. I'll put `of` again using a split over the string.

lib/android-helpers.js
vrunoa: 1 is good for the new method, but we are keeping the old method, maybe we can leave 10 until we deprecated the unlockWithHelperApp

lib/android-helpers.js
jlipps: alternatively, just refactor so that the conditional is outside the retry:

```js
if (typeof capabilities.unlockType === 'undefined') {
  logger.warn("Using app unlock, this is going to be deprecated!");
    await helpers.unlockWithHelperApp(adb);
  });
} else {
  await helpers.unlockWithUIAutomation(driver, adb, {unlockType: capabilities.unlockType, unlockKey: capabilities.unlockKey});
}
```

lib/android-helpers.js
```diff
@@ -411,23 +411,28 @@ helpers.unlock = async function (driver, adb, capabilities) {
...
+    });
+  } else {
+    await helpers.unlockWithUIAutomation(driver, adb, {unlockType: capabilities.unlockType, unlockKey: capabilities.unlockKey});
+    await helpers.verifyUnlock(adb);
+  }
+};
+
+helpers.verifyUnlock = async function (adb) {
+  await retryInterval(2, 1000, async () => {
+    if (!await adb.isScreenLocked()) {
```
mykola-mokhnach: I'd rather use positive condition to make it easier to read:

```javascript
...
if (await adb.isScreenLocked()) {
  throw new Error("Screen did not unlock successfully, retrying");
}
logger.debug("Screen unlocked successfully");
...
```

lib/commands/actions.js
```diff
@@ -118,8 +118,8 @@ commands.isLocked = async function () {
   return await this.adb.isScreenLocked();
 };
 
-commands.unlock = async function () {
-  return await androidHelpers.unlockWithHelperApp(this.adb);
+commands.unlock = async function (unlock_capabilities) {
```
mykola-mokhnach: ->unlockCapabilities

lib/driver.js
```diff
@@ -94,7 +94,10 @@ class AndroidDriver extends BaseDriver {
       if (!this.opts.javaVersion) {
         this.opts.javaVersion = await helpers.getJavaVersion();
       }
-
+      this.useUnlockHelperApp = true;
```
mykola-mokhnach: -> this.useUnlockHelperApp = !_.isUndefined(this.caps.unlockType);

lib/unlock-helpers.js
```diff
@@ -87,16 +87,14 @@ helpers.pinUnlock = async function (adb, driver, capabilities) {
       let text = await driver.getAttribute("text", e.ELEMENT);
       pins[text] = e;
     }
-    for (let i in keys) {
-      let pin = keys[i];
+    for (let pin of keys.split(/s*/)) {
```
mykola-mokhnach:
```
split(/s*/) 
```

has been already done above. Do we need to do it again here?

lib/unlock-helpers.js
```diff
@@ -87,16 +87,14 @@ helpers.pinUnlock = async function (adb, driver, capabilities) {
...
       let el = pins[pin];
       await driver.click(el.ELEMENT);
     }
     let el = await driver.findElOrEls("id", "com.android.systemui:id/key_enter", false);
     await driver.click(el.ELEMENT);
   } else {
-    for (let i in keys) {
-      let pin = keys[i];
+    for (let pin of keys.split(/s*/)) {
```
mykola-mokhnach: same as above

lib/unlock-helpers.js
```diff
@@ -162,8 +160,7 @@ helpers.patternUnlock = async function (adb, driver, capabilities) {
   let piece = size.width / 3;
   let actions = [];
   let lastPos;
-  for (let i in keys) {
-    let key = key[i];
+  for (let key of keys.split(/s*/)) {
     let keyPos = helpers.getPatternKeyPosition(key, pos0, piece);
     if (parseInt(key, 10) === 0) {
```
mykola-mokhnach: **key** will never be equal to zero according to the algorithm, so this condition is never going to be executed.

lib/unlock-helpers.js
```diff
@@ -76,7 +76,7 @@ helpers.encodePassword = function (key) {
 helpers.pinUnlock = async function (adb, driver, capabilities) {
   logger.info(`Trying to unlock device using pin ${capabilities.unlockKey}`);
   await helpers.dismissKeyguard(driver, adb);
-  let keys = capabilities.unlockKey.split("");
+  let keys = capabilities.unlockKey.split(/s*/);
```
mykola-mokhnach: did you mean 

```
/\s*/
```

?

lib/android-helpers.js
vrunoa: Sure, i'll change it. But this was made like this when I got here 😄

imurchie: Isn't that so often the case!

lib/android-helpers.js
```diff
@@ -428,10 +428,10 @@ helpers.unlock = async function (driver, adb, capabilities) {
 
 helpers.verifyUnlock = async function (adb) {
   await retryInterval(2, 1000, async () => {
-    if (!await adb.isScreenLocked()) {
-      logger.debug("Screen unlocked successfully");
-    } else {
+    if (await adb.isScreenLocked()) {
       throw new Error("Screen did not unlock successfully, retrying");
+    } else {
```
mykola-mokhnach: _else_ is not needed here. _throw_ operator will anyway interrupt the flow.

lib/unlock-helpers.js
```diff
@@ -0,0 +1,221 @@
...
+  ];
+  await driver.performTouch(actions);
+};
+
+helpers.encodePassword = function (key) {
+  return key.replace(/\s/ig, "%s");
+};
+
+helpers.stringKeyToArr = function (key) {
+  return key.trim().replace(/\s+/, '').split(/\s*/);
```
mykola-mokhnach: should the replacement regexp be global here?

vrunoa: sorry I don't understand, what do you mean ?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,221 @@
...
+};
+
+helpers.pinUnlock = async function (adb, driver, capabilities) {
+  logger.info(`Trying to unlock device using pin ${capabilities.unlockKey}`);
+  await helpers.dismissKeyguard(driver, adb);
+  let keys = helpers.stringKeyToArr(capabilities.unlockKey);
+  if (await adb.getApiLevel() >= 21) {
+    let els = await driver.findElOrEls("id", "com.android.systemui:id/digit_text", true);
+    if (els === null || els.length === 0) {
```
mykola-mokhnach: you may use _.isEmpty(els) shortcut here

mykola-mokhnach: same comment to other _driver.findElOrEls_ calls

vrunoa: Great! Yeah, I have to incursionate more into *lodash*

lib/unlock-helpers.js
mykola-mokhnach:
```
.replace(/\s+/, '')
```

will replace only the very first occurrence and

```
.replace(/\s+/g, '')
```

will replace all occurrences

vrunoa: Oh, yeah. You're right.

vrunoa:
@mykola-mokhnach @imurchie the documentation for the new unlock method should be here or in the appium repo ?

lib/driver.js
```diff
@@ -210,8 +216,12 @@ class AndroidDriver extends BaseDriver {
     this.caps.deviceManufacturer = await this.adb.getManufacturer();
 
     // Let's try to unlock before installing the app
-    // unlock the device
-    await helpers.unlock(this.adb);
+    // unlock the device using the HelperApp in case a user
+    // dont set an unlock type
+    if (this.useUnlockHelperApp === true) {
```
mykola-mokhnach: nit: -> if (this.useUnlockHelperApp) {

mykola-mokhnach: same below

lib/unlock-helpers.js
```diff
@@ -0,0 +1,222 @@
...
+      actions.push({action: 'press', options: {element: null, x: keyPos.x, y: keyPos.y}});
+      lastPos = keyPos;
+      continue;
+    }
+    let moveTo = {x:0, y:0};
+    let diffX = keyPos.x - lastPos.x;
+    if (diffX > 0) {
+      moveTo.x = piece;
+      if (Math.abs(diffX) > piece) {
+        moveTo.x+= piece;
```
mykola-mokhnach: nit: +=, -= are operators and should be separated with spaces from both sides

lib/unlock-helpers.js
```diff
@@ -0,0 +1,222 @@
...
+  let xPos = (key, x, piece) => {
+    return Math.round(x + ((key % cols) || cols) * piece - piece / 2);
+  };
+  let yPos = (key, y, piece) => {
+    return Math.round(y + (Math.ceil(((key % pins) || pins) / cols) * piece - piece / 2));
+  };
+  return {x: xPos(key, initPos.x, piece), y: yPos(key, initPos.y, piece)};
+};
+
+helpers.getPatternActions = function (keys, initPos, piece) {
```
mykola-mokhnach: I think there should be a verification of the fact that keys array should at least include 2 elements.

vrunoa: Changing the valid *regex* for the pattern

lib/driver.js
```diff
@@ -375,7 +382,6 @@ class AndroidDriver extends BaseDriver {
     }
     // some cleanup we want to do regardless, in case we are shutting down
     // mid-startup
-    await this.adb.forceStop('io.appium.unlock');
```
vrunoa: Re-added
```javascript
if (this.useUnlockHelperApp) {
  await this.adb.forceStop('io.appium.unlock');
}
```

lib/unlock-helpers.js
vrunoa:
```javascript
if (!/^[1-9]{2,9}$/.test(key.trim())) {
    return false;
}
```

vrunoa:
@imurchie @mykola-mokhnach is there anything left on this PR that you think I should change ?

lib/unlock-helpers.js
```diff
@@ -0,0 +1,222 @@
...
+helpers.patternUnlock = async function (adb, driver, capabilities) {
+  logger.info(`Trying to unlock device using pattern ${capabilities.unlockKey}`);
+  await helpers.dismissKeyguard(driver, adb);
+  let keys = helpers.stringKeyToArr(capabilities.unlockKey);
+  /* We set the device pattern buttons as number of a regular phone
+   *  | • • • |     | 1 2 3 |
+   *  | • • • | --> | 4 5 6 |
+   *  | • • • |     | 7 8 9 |
+
+  The pattern view buttons is not seeing by the uiautomator, so we are going to
```
jlipps: `are not seen by uiautomator`

lib/driver.js
```diff
@@ -227,6 +237,9 @@ class AndroidDriver extends BaseDriver {
       }
     });
 
+    // Let's try to unlock the device
+    await helpers.unlock(this, this.adb, this.caps);
```
jlipps: Imagine that `this.useUnlockHelperApp` is true. Now we've gotten to this line. I'm a little confused why we call `helpers.unlock` here, which calls `helpers.unlockWithHelperApp` inside of it, even though we've already called it above in this case?

lib/android-helpers.js
```diff
@@ -372,39 +373,65 @@ helpers.pushStrings = async function (language, adb, opts) {
...
-    await adb.startApp(startOpts);
-
-    // check if it worked, twice
-    await retryInterval(2, 1000, async () => {
-      if (!await adb.isScreenLocked()) {
-        logger.debug("Screen unlocked successfully");
-      } else {
-        throw new Error("Screen did not unlock successfully, retrying");
-      }
+  if (typeof capabilities.unlockType === 'undefined') {
```
jlipps: we used `_.isUndefined` below for this same purpose, maybe we should do that here to be consistent?

lib/android-helpers.js
```diff
@@ -418,7 +445,9 @@ helpers.initDevice = async function (adb, opts) {
     defaultIME = await helpers.initUnicodeKeyboard(adb);
   }
   await helpers.pushSettingsApp(adb);
-  await helpers.pushUnlock(adb);
+  if (typeof opts.unlockType === 'undefined') {
```
jlipps: and also here?

lib/driver.js
vrunoa: yeah, my bad. doesnt make sense to call this twice.

vrunoa:
@mykola-mokhnach @imurchie @jlipps  what do you think on this document for introducing the the unlock with uiautomation ?
https://github.com/vrunoa/appium-android-driver/blob/refactor-unlock/UNLOCK.md

test/functional/unlocker-e2e-specs.js
```diff
@@ -0,0 +1,53 @@
...
+describe('unlock tests', function () {
+  let driver;
+  before(() => {
+    driver = new AndroidDriver();
+  });
+  afterEach(async () => {
+    await driver.deleteSession();
+  });
+  /*
+  it('should unlock the device using a PIN', async () => {
```
imurchie: I'd prefer these to be skipped (https://mochajs.org/#inclusive-tests) rather than commented out. That way we can see that they are skipped in the test report, and audit what is there but not being run with a simple search.

vrunoa: Noice! didnt knew about the `.skip` method for mocha.

UNLOCK.md
```diff
@@ -0,0 +1,40 @@
+# Unlock with UIAutomation (New)
```
jlipps: would prefer to see this in `docs/unlock.md` (new dir, no caps). looks great though!

test/functional/unlocker-e2e-specs.js
jlipps: agree about `.skip`

vrunoa:
@imurchie @jlipps @mykola-mokhnach I added a method to unlock using fingerprint, it's only supported for `Android 6+ emulators` since lower versions of Android dont have fingerprint support and AFAIK there's not support for simulating fingerprint for real devices. 

Please review:
https://github.com/vrunoa/appium-android-driver/blob/refactor-unlock/lib/unlock-helpers.js#L82-L88
https://github.com/vrunoa/appium-android-driver/blob/refactor-unlock/lib/android-helpers.js#L376-L392
https://github.com/vrunoa/appium-android-driver/blob/refactor-unlock/test/unit/android-helper-specs.js#L445-L460
https://github.com/vrunoa/appium-android-driver/blob/refactor-unlock/test/unit/driver-specs.js#L56-L62

vrunoa:
Cool! thanks for the all the reviews and feedback @mykola-mokhnach @jlipps @imurchie

jlipps:
published in `appium-android-driver@1.12.0`

PR #212, opened 2017-02-24
Enable mock locations for appium settings app

umutuzgur:
* Enable mock location for all apps below Android 6 and set io.appium.settings as the mock location app on Android 6 and above to fix the mock location issue

This PR fixes the issue below
https://github.com/appium/appium/issues/7647

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -335,6 +335,14 @@ helpers.initUnicodeKeyboard = async function (adb) {
   return defaultIME;
 };
 
+helpers.setMockLocationApp = async function (adb, app) {
+  if (parseInt(await adb.getApiLevel(), 10) < 23) {
+    await adb.shell(['settings', 'put', 'secure', 'mock_location', '1']);
+  } else {
+    await adb.shell(['appops', 'set', app, 'android:mock_location', 'allow']);
+  }
+};
+
 helpers.pushSettingsApp = async function (adb) {
   logger.debug("Pushing settings apk to device...");
   await adb.install(settingsApkPath, false);
@@ -419,6 +427,7 @@ helpers.initDevice = async function (adb, opts) {
   }
   await helpers.pushSettingsApp(adb);
   await helpers.pushUnlock(adb);
+  await helpers.setMockLocationApp(adb, 'io.appium.settings');
   return defaultIME;
 };
 
--- a/test/unit/android-helper-specs.js
+++ b/test/unit/android-helper-specs.js
@@ -383,6 +383,22 @@ describe('Android Helpers', () => {
       mocks.adb.verify();
     });
   }));
+  describe('setMockLocationApp', withMocks({adb}, (mocks) => {
+    it('should enable mock location for api level below 23', async () => {
+      mocks.adb.expects('getApiLevel').returns(Promise.resolve("18"));
+      mocks.adb.expects('shell').withExactArgs(['settings', 'put', 'secure', 'mock_location', '1']).once()
+        .returns('');
+      await helpers.setMockLocationApp(adb, 'io.appium.settings');
+      mocks.adb.verify();
+    });
+    it('should enable mock location for api level 23 and above', async () => {
+      mocks.adb.expects('getApiLevel').returns(Promise.resolve("23"));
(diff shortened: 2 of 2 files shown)


lib/android-helpers.js
```diff
@@ -335,6 +335,14 @@ helpers.initUnicodeKeyboard = async function (adb) {
   return defaultIME;
 };
 
+helpers.setMockLocationApp = async function (adb, app) {
+  if (adb.getApiLevel() < 23) {
```
jlipps: i believe `getApiLevel` is an `async` function, so this should be:

```js
if (await adb.getApiLevel() < 23) {
```

this was probably "working" because `Object < 23` evaluates to `false`, so you fall into the "correct" conditional fork.

umutuzgur: Thanks @jlipps

umutuzgur:
I can write something but it would be really trivial since it is just based on the documentation from Andorid. I can share the link to the corresponding url thou

jlipps:
well, what it would catch is issues like the one i caught in my review. very lightweight. i don't think we need a functional test of the behavior necessarily.

umutuzgur:
Let me write something then

test/unit/android-helper-specs.js
```diff
@@ -383,6 +383,22 @@ describe('Android Helpers', () => {
       mocks.adb.verify();
     });
   }));
+  describe('setMockLocationApp', withMocks({adb}, (mocks) => {
+    it('should enable mock location for api level below 23', async () => {
+      mocks.adb.expects('getApiLevel').returns("18");
```
jlipps: I think this line (and the one in the other test) should read:

```js
mocks.adb.expects('getApiLevel').returns(Promise.resolve("18"));
```

because `getApiLevel` is an async function and should return a Promise.

lib/android-helpers.js
```diff
@@ -335,6 +335,14 @@ helpers.initUnicodeKeyboard = async function (adb) {
   return defaultIME;
 };
 
+helpers.setMockLocationApp = async function (adb, app) {
+  if (await adb.getApiLevel() < 23) {
```
jlipps: I also just realized, `getApiLevel` presumably returns a string? At least, looking in the code, I couldn't see us doing any conversion here. Maybe to have a safer comparison, we should convert this to a number:

```js
if (parseInt(await adb.getApiLevel(), 10) < 23) {
```

alternatively, we could update the library method to return a number itself; not sure if that will break how anyone else is depending on the API already, though.

umutuzgur: I don't know if I should change that much. I can parse it into an int. There is already a call doing what I'm doing in this file. That should also change probably

jlipps: yes, there's nothing wrong with what you're doing. technically it's OK since JS does implicit conversion when comparing strings and numbers, but i never like to rely on that kind of implicit stuff. i'm just wondering if we should take the opportunity now to be more explicit, since we're looking at it.

umutuzgur: I would go ahead and do that right now if this was a mono repo but I believe if we even decide  to change adb.getApiLevel we should expect a string for now and change adb repo after we merge this. With projects in the structure like appium, with a lot of small repos, I like go step by step usually and make sure that the thing I'm changing doesn't rely on a change in a different repo. Even though, as you said, JS does implicit conversion and it probably wouldn't break anything

umutuzgur:
Thanks. It seems like there is a problem with the CI btw

jlipps:
@umutuzgur if you run all the functional and unit tests locally and they pass, then I'll go ahead and merge

umutuzgur:
@jlipps I just need to call `npm test` right? If so everything is passing with npm test

jlipps:
No, you need to run these:

```
gulp once # run lint / unit tests
gulp e2e-test # run functional tests
```

for functional tests, we look for an AVD with the name `NEXUS_S_18_X86`, there's an env var you can use to override that but I forget what it is at the moment. Sorry, this info should be in the README.

umutuzgur:
Working for both `NEXUS_S_18_X86` and `NEXUS_S_24_X86` and mock location is enabled for both

jlipps:
published in `appium-android-driver@1.12.0`

PR #215, opened 2017-03-19
Add Android 7 WebView Browser Tester support

vrunoa:
Android 7 emulators without Google Play Services doesnt have the Stock Browser or Chrome by default, it comes with *WebView Browser Tester* (chromium 51.x)

Added the package and activity to run tests over this browser, it works with chromedriver `2.23` or below, so it will work by setting a custom `chromedriverExecutable`, since `appium-chromedriver` current version is `2.26`. 

Please review @imurchie @mhan

### DIFF
--- a/test/functional/webview-browser-tester-e2e-specs.js
+++ b/test/functional/webview-browser-tester-e2e-specs.js
@@ -1,0 +1,46 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import AndroidDriver from '../..';
+import path from 'path';
+import { sleep } from 'asyncbox';
+
+chai.should();
+chai.use(chaiAsPromised);
+
+const AVD_ANDROID_24_WITHOUT_GMS = "Nexus_5_API_24";
+const capabilities = {
+  "browserName": "chromium-webview",
+  "avd": AVD_ANDROID_24_WITHOUT_GMS,
+  "platformName": "Android",
+  "platformVersion": "7.0",
+  "deviceName": "Android Emulator",
+  "chromedriverExecutable": path.join(process.cwd(), "chromedriver")
+};
+
+describe('createSession', function () {
+  let driver;
+  before(() => {
+    driver = new AndroidDriver();
+  });
+  afterEach(async () => {
+    await driver.deleteSession();
+  });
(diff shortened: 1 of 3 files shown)


vrunoa:
We could check it like this;
```javascript
let output = await driver.adb.shell('pm list packages');
let api = await driver.getApiLevel();
if (api === 24 && output.indexOf('package:com.google.android.gms') === -1) {
  // this is a android 7 without gms
}
```
what do you think @imurchie ?
Also, test are going to run only on avds ? Otherwise we should add another validation to verify it's running over an emulator.

imurchie:
Yeah. There should be examples in the tests of skipping tests for real devices. In the iOS tests we usually run with the environment variable `REAL_DEVICE` set to something truthy. The in the test code we can do `if (process.env.REAL_DEVICE) return this.skip();

test/functional/webview-browser-tester-e2e-specs.js
```diff
@@ -0,0 +1,53 @@
...
+  "platformVersion": "7.0",
+  "deviceName": "Android Emulator",
+  "chromedriverExecutable": path.join(process.cwd(), "chromedriver")
+};
+
+describe('Android 7 Webview Browser tester', function () {
+  let driver, test = this;
+  before(() => {
+    if (process.env.REAL_DEVICE) {
+      test.pending = true;
```
imurchie: This can more easily be written without the `=>` syntax (so we get the normal scoping) and getting rid of the `test` variable:
```js
before(function () {
  this.pending = process.env.REAL_DEVICE;
});
```
But also, what happens if it _is_ pending? Does the test not run?

vrunoa: Yes, it's skipped but set as _pending_. If we use `it.skip` it shows like it's "passing"

screen shot 2017-03-20 at 1 31 48 pm

what do you guys prefer ? I took a screenshot of both cases.

imurchie: Can you try `return this.skip();`? That should show up as pending.

vrunoa: Done. It was different when I did `it.skip`

imurchie:
Published in `appium-android-driver@1.15.0`.

PR #216, opened 2017-03-20
add apis for recording the screen

heeseon:
added two apis 
1. startRecordingScreen
2. stopRecordingScreen


imurchie:
This has a lot of stuff that already exists in the codebase.

heeseon:
then.. I can't apply it ? I resolved timing issue, please check the code, and unit test case.

imurchie:
You need to remove all the performance stuff. That came in before. It looks like you worked off an old branch.

heeseon:
removed all the performance stuff

imurchie:
I do not understand why you've deleted a bunch of other stuff to put this in. Are you intending for there to be no methods for getting performance stats?

What more reasonably should happen is you should start a new branch, off of the latest master for the driver, and add a `lib/commands/recording.js` file with these methods, and then the associated tests.

PR #218, opened 2017-03-23
record the screen while testing

heeseon:
add two apis
- startRecordingScreen
- stopRecordingScreen


heeseon:
in my local environment, there is no issue as like ci error
/home/travis/build/appium/appium-android-driver/test/functional/commands/recordscreen-e2e-specs.js
  4:24  error  Unable to resolve path to module 'sample-apps'  import/no-unresolved

I am not sure why it happens..

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+  //call the command  
+  data = adb.shell(cmd); 
+
+  
+  //there is the delay time to start recording the screen, so, wait until it is ready.
+  //the ready condition is
+  //1. check the movie file is created
+  //2. check it is started to capture the screen 
+  cmd = ['ls', '-al', filePath]; 
+  while (loop) {
```
imurchie: There is a method for this in `asyncbox`, called `retryInterval`. So this could be rewritten as something like:
```js
import { retryInterval } from 'asyncbox';

// ...

// try 10 times, waiting a second in between each try
  let fileExists = await adb.shell(cmd);
  if (fileExists.indexOf("No such file") === -1) {
    // I don't actually understand what the code is trying to do, so
    // fill this in
  }
}
```

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+    data = await adb.shell(cmd); 
+    indexNoSuchFile = data.indexOf("No such file");
+
+    if (indexNoSuchFile < 0) {
+      var arrayList2 = data.split(" ");
```
imurchie: What is the content of `arrayList2`? Perhaps a descriptive name would be helpful. Also use `let` over `var`.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+      length = _.size(arrayList2);
```
imurchie: `arrayList2.length` would suffice.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+  if (length > 0){
+    var arrayList2 = data.split(" ");
+    length = _.size(arrayList2);
+
+    //kill the process and wait for that it is completed
+    for (let i = 1 ; i < length ; ++ i){
+      if ( _.size(arrayList2[i]) > 0 ){
+        var screenrecordPID = arrayList2[i] * 1;
+
+        while (loop) {
```
imurchie: Again, `retryInterval` would work better.

test/functional/commands/recordscreen-e2e-specs.js
```diff
@@ -0,0 +1,148 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import AndroidDriver from '../../..';
+import sampleApps from 'sample-apps';
```
imurchie: This is the source of the CI failure. This package no longer uses `sample-apps`, importing `android-apidemos` directly instead. If you update your dependencies (`rm -rf node_modules && npm install`) you will see the same error.

It is best to import `DEFAULT_CAPS` from https://github.com/appium/appium-android-driver/blob/master/test/functional/desired.js

See https://github.com/appium/appium-android-driver/blob/master/test/functional/commands/element-e2e-specs.js#L5 for an example.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
+import _ from 'lodash';
+import ADB from 'appium-adb';
+import { sleep } from 'asyncbox';
+
+let adb = new ADB();
```
heeseon: I changed it.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+*                if it is -1, the default value is 4Mbps. You can increase the bit rate to improve video quality,
+*                but doing so results in larger movie files.
+*                for example, 6000000
+*
+*/
+commands.startRecordingScreen = async function ( filePath, videoSize, timeLimit, bitRate ) {
+  let cmd, data, length, index, indexNoSuchFile;
+  let loop = true;
+
+  cmd = ['getprop', 'ro.build.version.sdk'];
```
heeseon: I changed it

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+  data = await adb.shell(cmd); 
+
+  //this function is suppported on the device running android 4.4(api level 19)
+  if (data < 19) throw new Error('The screenrecord command is able to record the display of devices running Android 4.4 (API level 19) and higher.');
+
+  cmd = ['ls', '\'' + filePath + '\'', '|', 'grep', '\'No such file\''];
```
heeseon: changed

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+  data = await adb.shell(cmd); 
+
+  length = _.size(data);
+
+  //if there's same file in the path, then thorws error
+  if (length === 0) throw new Error('\'' + filePath + '\' is already exist.');
```
heeseon: changed

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+
+
+  //make adb command 
+  cmd = ['screenrecord', '\'' + filePath + '\''];
+
+  index = 2;
```
heeseon: changed

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+
+  if ( videoSize !== "default" ) {
```
heeseon: I define user interface as like 
                @param videoSize the format is widthxheight.
                 if it is "default", the default value is the device's native display resolution 

if you have good idea for it, then please let me know it, I will change the interface.

lib/commands/recordscreen.js
heeseon: I tested your code,
the function works asynchronously, so it returns to client at the first try. 
so, I have to change the performance code which is previous pull request.

lib/commands/recordscreen.js
heeseon: changed

heeseon: changed

lib/commands/recordscreen.js
heeseon: I tested your code,
the function works asynchronously, so it returns to client at the first try.
so, I have to change the performance code which is previous pull request.

test/functional/commands/recordscreen-e2e-specs.js
heeseon: changed

lib/commands/recordscreen.js
heeseon: for the string variable, I used _.size function. 
Is it good to keep for consistency?

lib/commands/recordscreen.js
imurchie: I'm not sure what you mean. This will continue in a loop until nothing throws an exception, and then pass. https://github.com/jlipps/asyncbox#retry

test/functional/commands/recordscreen-e2e-specs.js
heeseon: changed

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+    cmd [index ++] = "--size";
```
heeseon: changed

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,144 @@
...
+  data = await adb.shell(cmd); 
+
+  length = _.size(data);
+
+  //if there's same file in the path, then thorws error
+  if (length === 0) throw new Error('\'' + filePath + '\' is already exist.');
+
+
+  //make adb command 
+  cmd = ['screenrecord', '\'' + filePath + '\''];
```
heeseon: changed

test/functional/commands/recordscreen-e2e-specs.js
```diff
@@ -0,0 +1,145 @@
...
+import { sleep } from 'asyncbox';
+import _ from 'lodash';
+import DEFAULT_CAPS from '../desired';
+
+chai.should();
+chai.use(chaiAsPromised);
+
+let driver;
+let caps = _.defaults({
+  appActivity: '.view.TextFields'
```
imurchie: Needs an `appPackage` to run.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,161 @@
...
+  //call the command  
+  data = adb.shell(cmd); 
+
+  
+  //there is the delay time to start recording the screen, so, wait until it is ready.
+  //the ready condition is
+  //1. check the movie file is created
+  //2. check it is started to capture the screen 
+  cmd = ['ls', '-al', `${filePath}`]; 
+  while (loop) {
```
imurchie: This infinite loop will not work for us. If for any reason there is an error, this will just keep going and going.

Parsing the output like this is not going to fly. Try something like this:
```js
    let data = await adb.shell(cmd);
    if (data.indexOf('No such file') !== -1) {
      throw new Error('No video file found');
    }
    // https://regex101.com/r/mwmAhQ/1
    let match = /^[-rwx]{10}\s+\S+\s+\S+\s+(\d+)\s+/.exec(data.trim());
    if (!match) {
      throw new Error(`Unable to get list data for file '${filePath}'`);
    }
    let size = parseInt(match[1], 10);
    if (size < 32) {
      throw new Error(`File '${filePath}' is not yet full created`);
    }
  });
```

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,161 @@
...
+  if ( timeLimit !== -1 ) {
+    cmd.push('--time-limit', timeLimit);
+  } 
+
+  if ( bitRate !== -1 ) {
+    cmd.push('--bit-rate', bitRate);
+  } 
+
+  //call the command  
+  data = adb.shell(cmd);
```
imurchie: Need to `await` this call.

```js
await adb.shell(cmd);
```

imurchie: Sorry, wrong about that. But this needs a comment as to why it is _not_ being `await`-ed.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,161 @@
...
+commands.startRecordingScreen = async function ( filePath, videoSize, timeLimit, bitRate ) {
+  let cmd, data, length, indexNoSuchFile;
+  let loop = true;
+  let opts = {};
+  let {udid, emPort} = await helpers.getDeviceInfoFromCaps(opts);
+  
+  opts.udid = udid;
+  opts.emPort = emPort;
+
+  let adb = await helpers.createADB(opts.javaVersion,
```
heeseon: I am not sure what you mean

first, create one globally, and share it?
or 
second, adb is already attached to the device, so I can refer it

if you mean second, then please let me know how to refer the adb variable.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,161 @@
...
+  //this function is suppported on the device running android 4.4(api level 19)
+  if (data < 19) throw new Error('The screenrecord command is able to record the display of devices running Android 4.4 (API level 19) and higher.');
+
+  cmd = ['ls', `${filePath}`, '|', 'grep', 'No such file'];
+  data = await adb.shell(cmd); 
+
+  length = _.size(data);
+
+  //if there's same file in the path, then thorws error
+  if (length === 0) throw new Error(`${filePath} is already exist.`);
```
heeseon: length is used at the other code in the same function as below

length = _.size(result_ls_filePath_toArray);

lib/commands/recordscreen.js
heeseon: I already tested it. 
it returns the result at the first try.

I think there is some error in transpile code or, it originally operates.
please test the performance code as below. the performance code use the way which you recommanded.

//should change the first parameter with the application package name
driver.getPerformanceData("xxx.xxx.xxx", "cpuinfo", 10000);  

in case of "cpuinfo"
the adb usually can't get the value for the first call, it needs several calls.

you can know this function returns null.

lib/commands/recordscreen.js
heeseon: I will added it tonight.

test/functional/commands/recordscreen-e2e-specs.js
heeseon: I am wondering why it passed locally if it has a problem.

heeseon: anyway, I added it

lib/commands/recordscreen.js
imurchie: There should already be an instance of `ADB` on the driver object. It is `this.adb`.

lib/commands/recordscreen.js
imurchie: It does _not_ return immediately. I just tried. You must have missed something. The `getPerformanceData` method also uses `retryInterval` in the correct manner.

lib/commands/recordscreen.js
imurchie: But this particular instance is only used here. It is re-initialized later. So it is not necessary. Just `if (_.size(data) === 0)` would suffice.

lib/commands/recordscreen.js
heeseon: The getPerformanceData returns immediately.

lib/commands/recordscreen.js
heeseon: I changed it

lib/commands/recordscreen.js
imurchie: Setting retries to `5` I get it repeated five times, as expected, in the CPU performance code:
```
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","am","start","-W","-n","io.appium.android.apis/.view.TextFields","-S"]
    getPerformanceData
dbug ADB Getting connected devices...
dbug ADB 1 device(s) connected
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Getting connected devices...
dbug ADB 1 device(s) connected
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Getting connected devices...
dbug ADB 1 device(s) connected
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Getting connected devices...
dbug ADB 1 device(s) connected
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Getting connected devices...
dbug ADB 1 device(s) connected
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
dbug ADB Running '/Users/isaac/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","dumpsys","cpuinfo","|","grep","'io.appium.android.apis'"]
```

lib/commands/recordscreen.js
heeseon: I changed the code.

lib/commands/recordscreen.js
heeseon: please test the code as not the unit case, but application level test.
and please check the transpiled code which I attached in communication

heeseon: imurchie,

I can't check the period and test result in your test debugging message, and I can't check adb returns message.

Please check user level test case and the debugging message.

and check the transpiled code of the getPerformanceData API.

heeseon:
imurchie,

I can't check the period and test result in your test debugging message, and I can't check adb returns message.

Please check user level test case and the debugging message.

and check the transpiled code of the getPerformanceData API.

heeseon:
Then I will change the code like the function of "getPerformanceData", But it can cause the problem in user level usage. 
So, I will add the current code as comment in the file. Is it ok?
If you agreed with this, I will change the code.

heeseon:
I changed the looping code from while~sleep statement to retryInterval statement.
please review the changed code.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,129 @@
...
+*                  for example, "1280x720"
+* @param timeLimit the maximum recording time, in seconds. if it is -1, the default and maximum value is 180 (3 minutes).
+* @param bitRate the video bit rate for the video, in megabits per second.
+*                if it is -1, the default value is 4Mbps. You can increase the bit rate to improve video quality,
+*                but doing so results in larger movie files.
+*                for example, 6000000
+*
+*/
+commands.startRecordingScreen = async function ( filePath, videoSize, timeLimit, bitRate ) {
+  let cmd, data, length, indexNoSuchFile;
```
imurchie: `adb shell screenrecord` is only available on real devices. So this needs to short-circuit when tried on emulators, or else the error will be confusing. Something like
```js
  if (this.isEmulator()) {
    throw new errors.NotImplementedError('Android screen recording does not work on emulators');
  }
```

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,129 @@
...
+
+  
+  //there is the delay time to start recording the screen, so, wait until it is ready.
+  //the ready condition is
+  //1. check the movie file is created
+  //2. check it is started to capture the screen 
+  return await retryInterval(RETRY_PAUSE, RETRY_PAUSE, async () => {
+    cmd = ['ls', '-al', `${filePath}`]; 
+    data = await this.adb.shell(cmd);
+    indexNoSuchFile = data.indexOf("No such file");
```
imurchie: If the file is not found, throw an exception. This is what triggers the `retryInterval` to pause and then go into the next iteration. Otherwise this will only happen once.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,129 @@
...
+
+  //call the command asynchronously 
+  data = this.adb.shell(cmd); 
+
+  
+  //there is the delay time to start recording the screen, so, wait until it is ready.
+  //the ready condition is
+  //1. check the movie file is created
+  //2. check it is started to capture the screen 
+  return await retryInterval(RETRY_PAUSE, RETRY_PAUSE, async () => {
```
imurchie: This method can be simplified with regular expressions:
```js
  await retryInterval(100, RETRY_PAUSE, async () => {
    if (screenRecordErr) {
      return;
    }

    let lsData = await this.adb.shell(['ls', '-al', filePath]);
    if (lsData.indexOf('No such file') !== -1) {
      throw new Error(`Unable to find file '${filePath}'`);
    }

    let match = /[-rwxd]{10}\s*\S*\s*\S*\s*(\d*)\s*\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}\s[\S\.]*/.exec(lsData);
    if (!match) {
      throw new Error(`Unable to parse result of ls: '${lsData}'`);
    }
    let size = parseInt(match[1], 10);
    if (size < 32) {
      throw new Error(`File not yet above 32 bytes`);
    }
  });
```

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,129 @@
...
+      }
+    } 
+  });
+};
+
+/**
+* stop recording the screen.
+*/
+commands.stopRecordingScreen = async function ( ) {
+  let cmd, data, length;
```
imurchie: Try `await this.adb.killProcessesByName('screenrecord');`.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,129 @@
...
+commands.startRecordingScreen = async function ( filePath, videoSize, timeLimit, bitRate ) {
+  let cmd, data, length, indexNoSuchFile;
+
+  data = await this.adb.getApiLevel();
+
+  //this function is suppported on the device running android 4.4(api level 19)
+  if (data < 19) throw new Error('The screenrecord command is able to record the display of devices running Android 4.4 (API level 19) and higher.');
+
+  cmd = ['ls', `${filePath}`, '|', 'grep', 'No such file'];
+  data = await this.adb.shell(cmd);
```
imurchie: See `await this.adb.fileExists(filePath);`

lib/commands/recordscreen.js
imurchie: And there is a method `adb.ls`

lib/commands/recordscreen.js
vrunoa: @imurchie `adb shell screenrecord` is able to run on emulators. Which platform did you try?

imurchie: 7, 6, 5, 4.4. It never worked on any configuration.

vrunoa: mmm I just try `adb shell screenrecord /sdcard/screen.mp4` on an android 6 emulator and it worked. We could check if there's some args missing or extra for the `screenrecord` to work on emus.

imurchie: Hmm. Well, something is different and needs to be figured out before this can land, whatever the case.

imurchie: Perhaps it came in in some particular tools version? Perhaps the right check is just to try to see if the command is there, before running it.

vrunoa: this method seems to be failing when it's called from the script, no idea why. I can dig a bit more over the weekend. 
btw; the time limit is 3minutes, we may need to set this on the documentation. 
https://developer.android.com/studio/command-line/adb.html#screenrecord

I will also set defaults for the extra args 
`async function(filePath, videoSize = 'default', timeLimit = 180, bitRate = -1) {`

vrunoa:
@imurchie @heeseon screenrecord is no longer supported on android 7+

lib/commands/recordscreen.js
heeseon: @imurchie ok it sounds good try. 
@vrunoa I can't find that android 7+ does not support the 'adb shell screenrecord' command.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,129 @@
...
+  if ( timeLimit !== -1 ) {
+    cmd.push('--time-limit', timeLimit);
+  } 
+
+  if ( bitRate !== -1 ) {
+    cmd.push('--bit-rate', bitRate);
+  } 
+
+  //call the command asynchronously 
+  data = this.adb.shell(cmd);
```
heeseon: ok it sounds good, I will change the code.

lib/commands/recordscreen.js
heeseon: It just happens the delay before it is ready for recording the screen. it is not the exception(error).
I think that the 'while~ stop' statement is more readable and reasonable style.

lib/commands/recordscreen.js
heeseon: I am not sure why it has to cause the exception, it is not readable because it is not real exception. I think that the 'while~ stop' statement is more readable and reasonable style.

lib/commands/recordscreen.js
heeseon: I can't still verify the  'adb.killProcessesByName' method. The other methods of the 'await this.adb.xx' does not work synchronously. If you assure me of that, then I will change the code.

lib/commands/recordscreen.js
heeseon: ok i will change the code

heeseon:
@vrunoa I can't find that android 7+ does not support the 'adb shell screenrecord' command. can you show me the document for it?

vrunoa:
@heeseon sorry I cant find any documentation but it's not working for any android 7+ emulator with the latest tools.

lib/commands/recordscreen.js
heeseon: changed.

lib/commands/recordscreen.js
heeseon: applied

lib/commands/recordscreen.js
heeseon: applied

lib/commands/recordscreen.js
heeseon: applied

heeseon:
@imurchie CI failed because of the Error: Android screen recording does not work on emulators.

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,99 @@
...
+  if ( timeLimit !== -1 ) {
+    cmd.push('--time-limit', timeLimit);
+  } 
+
+  if ( bitRate !== -1 ) {
+    cmd.push('--bit-rate', bitRate);
+  } 
+
+  //call the command asynchronously 
+  data = this.adb.shell(cmd).catch((err) => {
```
vrunoa: `data` is not used further in the code, I dont think you need to declare it and assign it

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,99 @@
...
+  //there is the delay time to start recording the screen, so, wait until it is ready.
+  //the ready condition is
+  //1. check the movie file is created
+  //2. check it is started to capture the screen 
+  loop = 1;
+  while (loop < RETRY_PAUSE) {
+    let lsData = await this.adb.shell(['ls', '-al', filePath]);
+
+    if (lsData.indexOf('No such file') < 0) {
+      let match = /[-rwxd]{10}\s*\S*\s*\S*\s*(\d*)\s*\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}\s[\S\.]*/.exec(lsData);
```
vrunoa: can you add a comment about the regex on how it should work ? Perhaps leaving an example. should this be a const at the top of the script ?

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,99 @@
...
+
+  if (this.isEmulator()) {
+    throw new Error('Android screen recording does not work on emulators');
+  }
+  
+  data = await this.adb.getApiLevel();
+
+  //this function is suppported on the device running android 4.4(api level 19)
+  if (data < 19) throw new Error('The screenrecord command is able to record the display of devices running Android 4.4 (API level 19) and higher.');
+
```
vrunoa: you might want to use `log.errorAndThrow` instead of `throw new Error` to be consistent with all the other code

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,99 @@
...
+        if (size > 32) {
+          return;
+        }
+      }
+      await sleep(1000);
+    } else
+      await sleep(1000);
+
+    loop ++;
+  }
```
vrunoa: you're calling sleep in both if and else, maybe it should go outside of the block

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,99 @@
...
+*
+*/
+commands.startRecordingScreen = async function ( filePath, videoSize, timeLimit, bitRate ) {
+  let cmd, data, fileExists, loop;
+
+  if (this.isEmulator()) {
+    throw new Error('Android screen recording does not work on emulators');
+  }
+  
+  data = await this.adb.getApiLevel();
```
vrunoa: maybe it's better to call this var `api` or `apiLevel` instead of `data`

vrunoa:
@heeseon you try it running `./node_modules/.bin/gulp transpile && ./node_modules/.bin/mocha --compilers js:babel-core/register -t 0 test/functional/commands/recordscreen-e2e-specs.js`

lib/commands/recordscreen.js
heeseon: I deleted it.

lib/commands/recordscreen.js
heeseon: I added a comment.

lib/commands/recordscreen.js
heeseon: I searched the pattern in appium. I found that 
in case of the try{}catch() pattern, it usually used log.errorAndThrow.
so, I did not use log.errorAndThrow in this code, but I applied it in the try{}catch patten in my code. thank you!

lib/commands/recordscreen.js
heeseon: i applied

lib/commands/recordscreen.js
heeseon: applied

heeseon:
@vrunoa as your guide, I did it. it passed, but still the e2e-tests are pending.

heeseon:
@imurchie I'd like to finish this.

heeseon:
@imurchie @vrunoa how I complete this?

imurchie:
I still cannot get this to do anything. Does it work locally for you, @heeseon ?

test/functional/commands/recordscreen-e2e-specs.js
```diff
@@ -0,0 +1,150 @@
...
+      data = await driver.adb.shell(cmd); 
+
+      length = _.size(data);
+      // the same file is exist, then delete the file
+      if (length <= 0){
+        cmd = ['rm', '/sdcard/test.mp4'];
+        data = await driver.adb.shell(cmd); 
+      }
+
+      if (driver.adb.isEmulatorConnected('emulator-5554'))
```
imurchie: This is _always_ true. It needs to be `await`-ed.

imurchie:
I can get it to work with the following: https://gist.github.com/imurchie/82dff49f5580289e66875fbaa5cf3c73

test/functional/commands/recordscreen-e2e-specs.js
heeseon: ok I changed.

heeseon:
the difference is using the retryInterval-exception statement. so I changed it.

heeseon:
@imurchie I applied the code as your guide. but the ci test is failed. so I rollback the code.

test/functional/commands/recordscreen-e2e-specs.js
```diff
@@ -42,7 +42,7 @@ describe('recording the screen', () => {
         data = await driver.adb.shell(cmd); 
       }
       
-      isEmulated = await driver.adb.isEmulatorConnected('emulator-5554');
+      isEmulated = driver.adb.isEmulatorConnected('emulator-5554');
```
vrunoa: as @imurchie commented, this need to use *await* since _isEmulatorConnected* method is async. Otherwise you would be getting a Promise instead of the actual boolean result.

heeseon: I don't know why ci test failed after this was changed. so I rollbacked now.

imurchie:
Please provide a link to the pertinent failed job on Travis. I cannot tell which one it is to have any advice.

test/functional/commands/recordscreen-e2e-specs.js
heeseon: changed

test/unit/commands/recordscreen-specs.js
```diff
@@ -0,0 +1,135 @@
...
+
+chai.should();
+chai.use(chaiAsPromised);
+
+describe('recording the screen', () => {
+  let driver = new AndroidDriver();
+  let adb = new ADB();
+  driver.adb = adb;
+
+  describe('recording the screen', withMocks({adb, driver}, (mocks) => {
```
imurchie: This test does not do what it is supposed to do. It should be testing `driver.startRecordingScreen` but it just goes through the steps of that method and mocks everything. Nothing is being tested at all.

The tests ought to mock the calls that `startRecordingScreen` makes, to direct the flow in particular directions (a test for each possible direction), to test the logic within `startRecordingScreen` itself. As it is, that method is mocked, and so nothing inside is tested.

So, for instance, the first two error conditions are exercised by:
```js
      it('should fail to recording the screen on an emulator', async function () {
        mocks.driver.expects('isEmulator').returns(true);
        await driver.startRecordingScreen().should.eventually.be.rejectedWith(/Android screen recording does not work on emulators/);
      });
      it('should fail to recording the screen on a device with API level 18', async function () {
        mocks.driver.expects('isEmulator').returns(false);
        mocks.adb.expects('getApiLevel').returns(18);
        await driver.startRecordingScreen().should.eventually.be.rejectedWith(/Unable to record screen on API level '18'. Minimum level is 19/);
      });
```

lib/commands/recordscreen.js
```diff
@@ -0,0 +1,103 @@
...
+*                  use a size supported by your device's Advanced Video Coding (AVC) encoder.
+*                  for example, "1280x720"
+* @param timeLimit the maximum recording time, in seconds. if it is -1, the default and maximum value is 180 (3 minutes).
+* @param bitRate the video bit rate for the video, in megabits per second.
+*                if it is -1, the default value is 4Mbps. You can increase the bit rate to improve video quality,
+*                but doing so results in larger movie files.
+*                for example, 6000000
+*
+*/
+commands.startRecordingScreen = async function ( filePath, videoSize, timeLimit, bitRate ) {
```
imurchie: This still doesn't work for me. And isn't tested in CI, since that is all emulators.

heeseon: Now, it is passed. I don't know why it happens. I think the ci test under the emulator environment is not good.

test/unit/commands/recordscreen-specs.js
heeseon: I applied

heeseon: I think that 
the recording function is out of business that 
making an emulator environment in unit test case, and checking whether it is in emulator environment in unit test case.

imurchie:
Merging since it doesn't break any existing functionality.

PR #219, opened 2017-03-25
Fix for Appium Settings app install and network state change

mykola-mokhnach:
Please merge the PR after https://github.com/appium/appium-adb/pull/208 and https://github.com/appium/io.appium.settings/pull/7 are merged and published

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -16,6 +16,7 @@ const REMOTE_INSTALL_TIMEOUT = 90000; // milliseconds
 const CHROME_BROWSERS = ["Chrome", "Chromium", "Chromebeta", "Browser",
                          "chrome", "chromium", "chromebeta", "browser",
                          "chromium-browser", "chromium-webview"];
+const SETTINGS_HELPER_PKG_ID = 'io.appium.settings';
 
 let helpers = {};
 
@@ -346,7 +347,8 @@ helpers.setMockLocationApp = async function (adb, app) {
 
 helpers.pushSettingsApp = async function (adb) {
   logger.debug("Pushing settings apk to device...");
-  await adb.install(settingsApkPath, false);
+  await adb.installOrUpgrade(settingsApkPath, SETTINGS_HELPER_PKG_ID);
+  await adb.grantAllPermissions(SETTINGS_HELPER_PKG_ID);
 };
 
 helpers.pushUnlock = async function (adb) {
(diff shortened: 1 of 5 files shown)


lib/commands/network.js
```diff
@@ -46,7 +46,7 @@ commands.setNetworkConnection = async function (type) {
   if (!airplaneMode) {
     await this.wrapBootstrapDisconnect(async () => {
       await this.setWifiState(wifi);
-      await this.adb.setDataState(data);
+      await this.adb.setDataState(data, !_.isUndefined(this.opts.avd));
```
imurchie: I'd like to see this check be pulled out into a method like `isEmulator`.

mykola-mokhnach: fixed

mykola-mokhnach: fixed

lib/android-helpers.js
```diff
@@ -29,6 +30,13 @@ helpers.parseJavaVersion = function (stderr) {
   return null;
 };
 
+helpers.isEmulator = function (opts = null) {
```
imurchie: These helpers are not attached to the driver object, so this method is unavailable. I would add it directly to the driver object, and use the driver's copy of `opts`.

mykola-mokhnach: fixed and updated unit tests

test/unit/android-helper-specs.js
```diff
@@ -377,12 +377,19 @@ describe('Android Helpers', () => {
...
+    it('should install settingsApp for real device', async () => {
+      mocks.adb.expects('installOrUpgrade').once()
+        .returns(true);
+      mocks.adb.expects('grantAllPermissions').withExactArgs('io.appium.settings').once()
+        .returns(true);
       await helpers.pushSettingsApp(adb);
       mocks.adb.verify();
     });
+    it('should not install settingsApp for simulator', async () => {
+      mocks.adb.expects('installOrUpgrade').never();
```
imurchie: Why is this expected to be called never? All `helpers.pushSettingsApp` does is call `installOrUpgrade` and `grantAllPermissions` with no conditional. The real device/simulator condition is checked in `initDevice`, so in the test for that is where the distinction should be made. Here you just need the one test, checking that it does indeed call both the intended methods.

mykola-mokhnach: yes. you're right

imurchie:
Published in `appium-android-driver@1.16.0`.

PR #221, opened 2017-03-28
Dismiss chrome welcome

vrunoa:
Some Chrome browsers are failing to avoid the Welcome dialog using the `--no-first-run` flag making tests to fail. 
I look at chromeOptions args in here;

* https://sites.google.com/a/chromium.org/chromedriver/capabilities
* http://peter.sh/experiments/chromium-command-line-switches/
* https://cs.chromium.org/chromium/src/chrome/common/chrome_switches.cc?q=kNoFirstRun&sq=package:chromium&type=cs&l=487

Unless i'm wrong I believe this flag is the one that should avoid the welcome dialog, since it's working for Android 7 gms emulators that had Chrome 51, but not for Android 7.1 emus with Chrome 53. 
We can avoid this by dismissing the dialog using uiautomation

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -337,11 +337,11 @@ class AndroidDriver extends BaseDriver {
       let el = await this.findElOrEls('id', 'com.android.chrome:id/terms_accept', false);
       await this.click(el.ELEMENT);
       try {
-        el = await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false);
-        if (!_.isUndefined(el)) {
-          await this.click(el.ELEMENT);
-        }
+        let el = await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false);
+        await this.click(el.ELEMENT);
       } catch (e) {
+        // DO NOTHING, THIS DEVICE DIDNT LAUNCH THE SIGNIN DIALOG
+        // IT MUST BE A NON GMS DEVICE
         log.warn(`This device didnt show Chrome SignIn dialog, ${e.message}`);
       }
     } else {
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -337,7 +337,7 @@ class AndroidDriver extends BaseDriver {
       let el = await this.findElOrEls('id', 'com.android.chrome:id/terms_accept', false);
       await this.click(el.ELEMENT);
       el = await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false).catch((e) => {
-        // DO NOTHING, THIS DEVICE DONT LAUNCH THE SIGNIN DIALOG,
+        // DO NOTHING, THIS DEVICE DIDNT LAUNCH THE SIGNIN DIALOG,
         // IT MUST BE A NON GMS DEVICE
         log.warn(`This device didnt show Chrome SignIn dialog, ${e.message}`);
       });
@@ -345,7 +345,7 @@ class AndroidDriver extends BaseDriver {
         await this.click(el.ELEMENT);
       }
     } else {
-      log.info("Chrome welcome dialog never show up! Continue");
+      log.info("Chrome welcome dialog never showoed up! Continuing");
     }
   }
 
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -261,6 +261,25 @@ class AndroidDriver extends BaseDriver {
       opts.chromeOptions.args.indexOf('--no-first-run') !== -1);
   }
 
+  async dismissChromeWelcome () {
+    log.info("Trying to dismiss Chrome welcome");
+    let activity = await this.getCurrentActivity();
+    if (activity === "org.chromium.chrome.browser.firstrun.FirstRunActivity") {
+      let el = await this.findElOrEls('id', 'com.android.chrome:id/terms_accept', false);
+      await this.click(el.ELEMENT);
+      try {
+        let el = await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false);
+        await this.click(el.ELEMENT);
+      } catch (e) {
+        // DO NOTHING, THIS DEVICE DIDNT LAUNCH THE SIGNIN DIALOG
+        // IT MUST BE A NON GMS DEVICE
+        log.warn(`This device didnt show Chrome SignIn dialog, ${e.message}`);
+      }
+    } else {
+      log.info("Chrome welcome dialog never showoed up! Continuing");
+    }
+  }
+
   async initAutoWebview () {
     if (this.opts.autoWebview) {
       let viewName = this.defaultWebviewName();
@@ -274,7 +293,6 @@ class AndroidDriver extends BaseDriver {
       });
     }
   }
-
 
   async initAUT () {
     // populate appPackage, appActivity, appWaitPackage, appWaitActivity,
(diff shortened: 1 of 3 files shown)


lib/driver.js
```diff
@@ -323,6 +323,30 @@ class AndroidDriver extends BaseDriver {
...
+    let activity = await this.getCurrentActivity();
+    if (activity === "org.chromium.chrome.browser.firstrun.FirstRunActivity") {
+      let el = await this.findElOrEls('id', 'com.android.chrome:id/terms_accept', false);
+      await this.click(el.ELEMENT);
+      el = await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false).catch((e) => {
+        // DO NOTHING, THIS DEVICE DONT LAUNCH THE SIGNIN DIALOG,
+        // IT MUST BE A NON GMS DEVICE
+        log.warn(`This device didnt show Chrome SignIn dialog, ${e.message}`);
+      });
+      if (!_.isUndefined(el)) {
```
jlipps: can't imagine there's any way that `el` could be undefined here?

lib/driver.js
```diff
@@ -323,6 +323,30 @@ class AndroidDriver extends BaseDriver {
...
+    }
+  }
+
+  async dismissChromeWelcome () {
+    log.info("Trying to dismiss Chrome welcome");
+    let activity = await this.getCurrentActivity();
+    if (activity === "org.chromium.chrome.browser.firstrun.FirstRunActivity") {
+      let el = await this.findElOrEls('id', 'com.android.chrome:id/terms_accept', false);
+      await this.click(el.ELEMENT);
+      el = await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false).catch((e) => {
```
jlipps: rather than mixing promises in, let's use try/catch:

```js
try {
  el = await this.findElOrEls(...);
} catch (e) {
  log.warn(`...`);
  return;
}
await this.click(el.ELEMENT);
```

lib/driver.js
```diff
@@ -323,6 +323,30 @@ class AndroidDriver extends BaseDriver {
...
+        // DO NOTHING, THIS DEVICE DONT LAUNCH THE SIGNIN DIALOG,
```
jlipps: `this device didn't launch`

lib/driver.js
```diff
@@ -323,6 +323,30 @@ class AndroidDriver extends BaseDriver {
...
+        // IT MUST BE A NON GMS DEVICE
+        log.warn(`This device didnt show Chrome SignIn dialog, ${e.message}`);
+      });
+      if (!_.isUndefined(el)) {
+        await this.click(el.ELEMENT);
+      }
+    } else {
+      log.info("Chrome welcome dialog never show up! Continue");
```
jlipps: `never showed up`

`continuing`

lib/driver.js
vrunoa: @jlipps if `await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false)` doesnt found the element if will go straight to catch, right ? Or will return an _undefined_(or _null_) value ?

jlipps: it will throw an error so yes it will go into the catch block. that is why in my code sample i `return`ed from the catch block, ensuring that outside of that `el` is defined

lib/driver.js
```diff
@@ -323,6 +323,30 @@ class AndroidDriver extends BaseDriver {
...
+
+  async dismissChromeWelcome () {
+    log.info("Trying to dismiss Chrome welcome");
+    let activity = await this.getCurrentActivity();
+    if (activity === "org.chromium.chrome.browser.firstrun.FirstRunActivity") {
+      let el = await this.findElOrEls('id', 'com.android.chrome:id/terms_accept', false);
+      await this.click(el.ELEMENT);
+      try {
+        el = await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false);
+        if (!_.isUndefined(el)) {
```
jlipps: this will never be undefined here, rather it will go into the catch block

vrunoa: perfect. changing it and adding an unit test

lib/driver.js
```diff
@@ -251,6 +255,31 @@ class AndroidDriver extends BaseDriver {
...
+      try {
+        let el = await this.findElOrEls('id', 'com.android.chrome:id/negative_button', false);
+        await this.click(el.ELEMENT);
+      } catch (e) {
+        // DO NOTHING, THIS DEVICE DIDNT LAUNCH THE SIGNIN DIALOG
+        // IT MUST BE A NON GMS DEVICE
+        log.warn(`This device didnt show Chrome SignIn dialog, ${e.message}`);
+      }
+    } else {
+      log.info("Chrome welcome dialog never showoed up! Continuing");
```
imurchie: showed

lib/driver.js
```diff
@@ -251,6 +255,31 @@ class AndroidDriver extends BaseDriver {
     await this.initAutoWebview();
   }
 
+  shouldDismissChromeWelcome (opts) {
```
mykola-mokhnach: what's the point to pass opts here if the method already has access to _this.opts_ ?

lib/driver.js
```diff
@@ -251,6 +255,31 @@ class AndroidDriver extends BaseDriver {
...
+    return (!_.isUndefined(opts.chromeOptions) &&
```
mykola-mokhnach: extra brackets

lib/driver.js
```diff
@@ -251,6 +255,31 @@ class AndroidDriver extends BaseDriver {
...
+      !_.isUndefined(opts.chromeOptions.args) &&
+      opts.chromeOptions.args.indexOf('--no-first-run') !== -1);
+  }
+
+  async dismissChromeWelcome () {
+    log.info("Trying to dismiss Chrome welcome");
+    let activity = await this.getCurrentActivity();
+    if (activity === "org.chromium.chrome.browser.firstrun.FirstRunActivity") {
```
mykola-mokhnach: does the activity always have the same id?

mykola-mokhnach: it would be faster to just return immediately here if activity !== ...

test/unit/driver-specs.js
```diff
@@ -250,6 +250,18 @@ describe('driver', () => {
...
+  describe('shouldDismissChromeWelcome', () => {
+    before(async () => {
+      driver = new AndroidDriver();
+    });
+    it('should verify chromeOptions args', () => {
+      driver.shouldDismissChromeWelcome({}).should.equal(false);
+      driver.shouldDismissChromeWelcome({"chromeOptions":{}}).should.equal(false);
+      driver.shouldDismissChromeWelcome({"chromeOptions":{"args":[]}}).should.equal(false);
+      driver.shouldDismissChromeWelcome({"chromeOptions":{"args":["--disable-dinosaur-easter-egg"]}}).should.equal(false);
+      driver.shouldDismissChromeWelcome({"chromeOptions":{"args":["--no-first-run"]}}).should.equal(true);
```
mykola-mokhnach: IMHO _should.be.true_ looks a bit better

lib/driver.js
```diff
@@ -251,6 +255,31 @@ class AndroidDriver extends BaseDriver {
     await this.initAutoWebview();
   }
 
+  shouldDismissChromeWelcome (opts) {
+    return (!_.isUndefined(opts.chromeOptions) &&
+      !_.isUndefined(opts.chromeOptions.args) &&
```
mykola-mokhnach: I'd prefer to verify that args is an array here

lib/driver.js
vrunoa: I have tried chrome version from 53 to 56 and it's always the same activity name.

lib/driver.js
vrunoa: I prefer to avoid using instance(global) vars but let me change it.

jlipps: yeah, either we should use the instance var, or this method should just be a non-member function outside of the class definition.

test/functional/chrome-e2e-specs.js
```diff
@@ -0,0 +1,32 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import AndroidDriver from '../..';
+
+chai.should();
+chai.use(chaiAsPromised);
+
+const capabilities = {
+  "browserName": "chrome",
+  "avd": "Nexus_5_API_25",
```
jlipps: should we hardcode the avd name here or take it from process.env? What do we do in other tests?

imurchie:
Merging and making note to check with the tests and make sure all is well.

imurchie:
Published in `appium-android-driver@1.17.0`.

test/functional/chrome-e2e-specs.js
vrunoa: no idea how are you guys managing e2e tests, but yeah a `process.env` var it's a good idea

PR #222, opened 2017-03-29
Fix the default background command behaviour and remove invalid option

mykola-mokhnach:
Set the _stopApp_ parameter to _false_ for background command. I believe the value of _dontStopAppOnReset_ should not  affect this particular endpoint and it was just a copypaste issue.

_stopAppOnReset_ capability seems to be obsolete and is neither defined nor used anywhere else in android driver:

```
$ grep -Ri 'stopAppOnReset' *
commands/general.js:                                         dontStopAppOnReset) {
commands/general.js:  // dontStopAppOnReset is both an argument here, and a desired capability
commands/general.js:  if (!util.hasValue(dontStopAppOnReset)) {
commands/general.js:    dontStopAppOnReset = !!this.opts.dontStopAppOnReset;
commands/general.js:    stopApp: !dontStopAppOnReset
commands/general.js:    stopApp: !this.opts.dontStopAppOnReset,
desired-caps.js:  dontStopAppOnReset: {
```

Do not restore the app from background if negative number of seconds is provides - the same behaviour we have for iOS

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -118,8 +118,13 @@ commands.installApp = async function (appPath) {
 };
 
 commands.background = async function (seconds) {
-  let {appPackage, appActivity} = await this.adb.getFocusedPackageAndActivity();
   await this.adb.goToHome();
+  if (seconds < 0) {
+    // if user passes in a negative seconds value, interpret that as the instruction
+    // to not bring the app back at all
+    return true;
+  }
+  let {appPackage, appActivity} = await this.adb.getFocusedPackageAndActivity();
   await B.delay(seconds * 1000);
   return this.adb.startApp({
     pkg: this.opts.appPackage,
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
@imurchie yw
it would be nice to update unit tests for this project to use mocks properly, like it is done in appium-adb, for example. So then it will be easier to write them.

imurchie:
Yes it would. I think all the machinery for that was written after the unit tests for this package were written, for the most part.

lib/commands/general.js
```diff
@@ -118,8 +118,11 @@ commands.installApp = async function (appPath) {
 };
 
 commands.background = async function (seconds) {
-  let {appPackage, appActivity} = await this.adb.getFocusedPackageAndActivity();
   await this.adb.goToHome();
+  if (seconds < 0) {
```
jlipps: might be nice to just have a comment here:

```js
// if user passes in a negative seconds value, interpret that as the instruction
// to not bring the app back at all
```

mykola-mokhnach: added the comment and squashed the stuff

PR #223, opened 2017-04-04
Fix emulator detection

mykola-mokhnach:
I dunno why, but it seems like the **avd** option is not equal to _undefined_ by default, but rather _null_
This patch should fix emulator detection for such case. See the issue report https://github.com/appium/appium/issues/8164 for the reference


imurchie:
Published in `appium-android-driver@1.17.1`.

PR #225, opened 2017-04-10
Add tests for webview-helpers.js to cover existing functionality

blutter:
I've added some tests to cover existing functionality in webview-helpers.js that I'd like to contribute.


imurchie:
Thanks for this!

PR #227, opened 2017-04-15
Add androidInstallPath capability

vrunoa:
Add possibility to set custom android install path using `androidInstallPath` capability. On some devices internal storage has low space for pushing large apps to `/data/local/tmp` so users may want to push the app to the sdcard.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -215,20 +215,22 @@ helpers.getLaunchInfo = async function (adb, opts) {
   return {appPackage, appWaitPackage, appActivity, appWaitActivity};
 };
 
-helpers.getRemoteApkPath = function (localApkMd5) {
-  let remotePath = `${REMOTE_TEMP_PATH}/${localApkMd5}.apk`;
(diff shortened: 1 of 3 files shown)


lib/android-helpers.js
```diff
@@ -215,20 +215,22 @@ helpers.getLaunchInfo = async function (adb, opts) {
   return {appPackage, appWaitPackage, appActivity, appWaitActivity};
 };
 
-helpers.getRemoteApkPath = function (localApkMd5) {
-  let remotePath = `${REMOTE_TEMP_PATH}/${localApkMd5}.apk`;
+helpers.getRemoteApkPath = function (localApkMd5, androidInstallPath) {
+  let remotePath = (androidInstallPath || REMOTE_TEMP_PATH).replace(/\/$/, '');
```
jlipps: another way to avoid the replace is:

```js
let remotePath = path.join(androidInstallPath || REMOTE_TEMP_PATH, `${localApkMd5}.apk`);
```

PR #230, opened 2017-04-24
Remove adm-zip dependency

dpgraham:
* Wrote a test for `commands.pullFolder` that verifies it works correctly with adm-zip
* Removed adm-zip dependency and made the test fail
* Added appium support zip dependency and replaced it with that and made the tests work again

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,13 +28,12 @@
     "lib": "lib"
   },
   "dependencies": {
-    "adm-zip": "^0.4.7",
     "appium-adb": "^2.16.0",
     "appium-android-bootstrap": "^2.7.5",
     "appium-android-ime": "^2.0.0",
     "appium-base-driver": "^2.0.0",
     "appium-chromedriver": "^2.9.0",
-    "appium-support": "^2.4.0",
+    "appium-support": "^2.8.0",
     "appium-unlock": "^0.1.0",
     "asyncbox": "^2.0.4",
     "babel-runtime": "=5.8.24",
@@ -73,6 +72,7 @@
     "eslint-plugin-import": "^1.9.2",
     "eslint-plugin-mocha": "^3.0.0",
     "gulp": "^3.8.11",
+    "mock-fs": "^4.2.0",
     "pre-commit": "^1.1.3",
     "sinon": "^1.16.1",
     "unzip": "^0.1.11",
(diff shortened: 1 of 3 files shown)


test/unit/commands/actions-specs.js
```diff
@@ -48,4 +50,41 @@ describe('Actions', () => {
       driver.bootstrap.sendAction.calledWith('drag').should.be.true;
     });
   });
+  describe('pullFolder', () => {
+    it('should pull a folder and return base64 zip', async () => {
+
+      let adbPullStub;
+      const pull = async (ignore, localPath) => {
+        await mkdirp(path.resolve(localPath));
```
imurchie: Why not just make this a fixture in `test/assets`?

test/unit/commands/actions-specs.js
```diff
@@ -48,4 +50,41 @@ describe('Actions', () => {
...
+      }
+
+      // Get the zipped buffer contents and write it to a .zip file
+      const buffer = await driver.pullFolder('/does/not/matter');
+      (typeof(buffer)).should.equal('string');
+      const zippedDir = await tempDir.openDir();
+      await fs.writeFile(path.resolve(zippedDir, 'zipped.zip'), buffer, 'base64');
+      
+      // Extract the zip file
+      const unzippedDir = await tempDir.openDir();
```
imurchie: Can't this all be done in streams or something so there doesn't need to be file system writes in a unit test?

dpgraham: I found a library called `mock-fs` that overrides `fs` and does everything in memory. I think I'll use that as a dev-dependency.

test/unit/commands/actions-specs.js
dpgraham: This was actually unnecessary. I took it out.

dpgraham:
Changes made. Uses `mock-fs` library as a dev-dependency.

package.json
```diff
@@ -72,7 +71,9 @@
     "eslint-plugin-babel": "^3.3.0",
     "eslint-plugin-import": "^1.9.2",
     "eslint-plugin-mocha": "^3.0.0",
+    "fancy-log": "^1.3.0",
```
dpgraham: I got rid of it. There was an error that happened earlier that said that the module was missing so I tried fixing it by adding it as a dev-dep. That was unneccessarry. I just did `rm -rf node_modules && npm install && gulp once` and it worked out fine.

jlipps: 👍

dpgraham:
Tagged and published to 1.17.4 (https://www.npmjs.com/package/appium-android-driver)

PR #231, opened 2017-04-26
Make all functional tests run in Travis

imurchie:
Forgot that `mocha` on its own does not recurse into subdirectories.


imurchie:
Seems like! One of the workers takes much longer, so I'm going to merge this, then see about moving those tests around and even things out, so the build doesn't take so long.

Might try the same thing with ios driver.

PR #234, opened 2017-05-04
Add sendSMS method

vrunoa:
Integrate *appium-adb* `.sendSMS` method

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -123,7 +123,6 @@ class AndroidDriver extends BaseDriver {
 
       // get device udid for this session
       let {udid, emPort} = await helpers.getDeviceInfoFromCaps(this.opts);
-      this.opts.avd = this.opts.avd || /emulator/.test(udid);
       this.opts.udid = udid;
       this.opts.emPort = emPort;
 
@@ -166,7 +165,7 @@ class AndroidDriver extends BaseDriver {
   }
 
   isEmulator () {
-    return !!this.opts.avd;
+    return !!(this.opts.avd || /emulator/.test(this.opts.udid));
   }
 
   setAvdFromCapabilities (caps) {
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -123,6 +123,7 @@ class AndroidDriver extends BaseDriver {
 
       // get device udid for this session
       let {udid, emPort} = await helpers.getDeviceInfoFromCaps(this.opts);
+      this.opts.avd = this.opts.avd || /emulator/.test(udid);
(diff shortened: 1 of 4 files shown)


imurchie:
I don't think `isEmulator` is correctly implemented. I just tried out a simple functional test for this and it failed because it didn't understand that it was on an emulator. Looking, in the regular flow of things there is no place where `this.opts.avd` is set.

vrunoa:
oh ok. i'll try to check and improve that

imurchie:
Any luck? I couldn't think of a solid way to do it.

vrunoa:
Sorry Isaack, I couldnt sit down and take a look yet.

lib/driver.js
```diff
@@ -123,6 +123,7 @@ class AndroidDriver extends BaseDriver {
 
       // get device udid for this session
       let {udid, emPort} = await helpers.getDeviceInfoFromCaps(this.opts);
+      this.opts.avd = this.opts.avd || /emulator/.test(udid);
```
vrunoa: @imurchie what do you think of this approach?

imurchie: Yeah, this was essentially the only way I could think of, too.

jlipps: so we are potentially setting `this.opts.avd` to `true` rather than an avd name? hmm  hmm. i wonder if it'd be more sensible to just track the udid in `this.opts.udid` or something like that, and then augment `isEmulator` to test for either case? just not sure that saturating `this.opts.avd` with more semantics is a good idea.

imurchie: I agree. We already have `this.opts.udid` set. https://github.com/appium/appium-android-driver/blob/master/lib/driver.js#L126

vrunoa: we can move this into the isEmulator method like
```javascript
isEmulator () {
  return !!(this.opts.avd || /emulator/.test(this.opts.udid));
}
```

test/unit/driver-specs.js
```diff
@@ -171,12 +171,14 @@ describe('driver', () => {
       driver.opts.udid = "01234567889";
       driver.isEmulator().should.equal(false);
     });
-    it('fingerprint should be rejected if avd is not defined', () => {
-      driver.opts = {};
+    it('fingerprint should be rejected if isEmulator is false', () => {
+      driver.opts.avd = undefined;
```
imurchie: Alternatively, you can mock the `isEmulator` call to return false, and assert that it is called.

imurchie:
Published in `appium-android-driver@1.20.0`.

PR #235, opened 2017-05-08
Make sure remote path for apk is in posix format

imurchie:
On Windows systems we end up with Windows-formatted paths for things on the device. See https://github.com/appium/appium/issues/8377


imurchie:
Published in `appium-android-driver@1.19.1`

PR #236, opened 2017-05-11
New emulator methods

vrunoa:
Add the new methods we have for emulators

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -11,7 +11,6 @@ import Bootstrap from 'appium-android-bootstrap';
 import ADB from 'appium-adb';
 import { default as unlocker, PIN_UNLOCK, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGERPRINT_UNLOCK } from './unlock-helpers';
 
-
 const REMOTE_TEMP_PATH = "/data/local/tmp";
 const REMOTE_INSTALL_TIMEOUT = 90000; // milliseconds
 const CHROME_BROWSERS = ["Chrome", "Chromium", "Chromebeta", "Browser",
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -61,7 +61,7 @@ helpers.prepareEmulator = async function (adb, opts) {
   }
   if (!_.isUndefined(opts.networkSpeed)) {
     let networkSpeed = this.ensureNetworkSpeed(adb, opts.networkSpeed);
-    avdArgs = [...avdArgs, '-netspeed', networkSpeed];
+    avdArgs += ` -netspeed ${networkSpeed}`;
   }
   await adb.launchAVD(avd, avdArgs, language, locale, avdLaunchTimeout,
                       avdReadyTimeout);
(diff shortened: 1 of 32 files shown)


vrunoa:
@jlipps yeah definitely. Wanna drive this ? I would like to see how you guys handle this in the past.

vrunoa:
* _powerAC, powerCapacity_ would be used for all the this other platforms appium supports: ios, windows desktop, osx desktop. Not sure if this would make sense for youi.tv
* same for _fingerprint_, mobile platforms + some windows and osx machines has fingerprint integrated
* I believe _sendSMS, gsmSignal, gsmCall, gsmVoice_ would be only available for mobile platforms

jlipps:
Yes, we can drive this, however I think it'd be fun for you to have a look at the current [mobile spec](https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md) to get a sense of how things are written. Care to take an initial stab and then we can walk through the rest of the process together?

vrunoa:
Sure @jlipps i'll take a look tonight. Ping me whenever you want(and have the time) after that

lib/commands/actions.js
```diff
@@ -182,6 +182,41 @@ commands.sendSMS = async function (phoneNumber, message) {
   await this.adb.sendSMS(phoneNumber, message);
 };
 
+commands.gsmCall = async function (phoneNumber, action) {
```
mykola-mokhnach: should these command return something?

vrunoa: *adb* doesnt return any value when calling it, so I guess not. If adbExec fails it will throw an error, right ?

mykola-mokhnach: Right. It was just a small reminder to make sure we didn't forget anything

vrunoa:
@jlipps would the new methods go under _Other devices features_ section ? https://github.com/SeleniumHQ/mobile-spec/compare/master...vrunoa:master

jlipps:
In this case I'm proposing that they each get their own sections in the spec, not that they be left out of the spec :-)

vrunoa:
Something like this? but well written 😄 
https://github.com/SeleniumHQ/mobile-spec/compare/master...vrunoa:master#diff-ee2352b7c156f97e44f211c294a46f2dR278

jlipps:
@vrunoa that looks great. The only thing missing is the actual endpoints and request/response values you would suggest.

vrunoa:
@jlipps this is ready. I'll add the endpoints on the appium-base-driver

filmaj:
Woo! This is rad. Thanks for taking this on.

vrunoa:
@jlipps going to include the new network speed method also on this PR. So, it's not ready to merge.

lib/android-helpers.js
```diff
@@ -59,9 +59,25 @@ helpers.prepareEmulator = async function (adb, opts) {
       return;
     }
   }
+  /* TODO - update appium-adb version to latest once https://github.com/appium/appium-adb/pull/245
+  is merged adn published, then finish the network speed implementation
+  if (_.isUndefined(opts.networkSpeed)) {
+    let networkSpeed = this.ensureNetworkSpeed(adb, opts.networkSpeed);
+    avdArgs = _.concat(avdArgs, ['-netspeed', networkSpeed]);
```
jlipps: should now be able to do:

```js
avdArgs = [...avdArgs, '-netspeed', networkSpeed];
```

vrunoa:
@imurchie @jlipps ready for another review.

imurchie:
Rebase and fix conflicts, and we should be good!

PR #237, opened 2017-05-15
Fix test-helper-specs path related tests when running on Windows

blutter:
The test-helper-specs tests do not account for different path segment separation characters on Windows.

I have normalized these paths so that the tests pass on Windows and also tested that the same tests pass on OSX.


imurchie:
Thanks for this!

PR #238, opened 2017-05-17
Update webview-helpers to handle crosswalk webviews 

blutter:
There has been a long running issue with finding crosswalk webviews on Android as logged in  https://github.com/appium/appium/issues/4597

There is a two part fix:
1. Fix chromedriver (there is a patch in codereview at https://codereview.chromium.org/2375613002/)
2. Update appium-android-driver to find the corresponding crosswalk webviews (this PR)

I have updated the tests to cover this scenario and implemented the changes to the webview-helper code.

### DIFF
--- a/lib/webview-helpers.js
+++ b/lib/webview-helpers.js
@@ -6,6 +6,8 @@ const NATIVE_WIN = "NATIVE_APP";
 const WEBVIEW_WIN = "WEBVIEW";
 const WEBVIEW_BASE = `${WEBVIEW_WIN}_`;
 const CHROMIUM_WIN = "CHROMIUM";
+const CROSSWALK_SOCKET_SUFFIX = "_devtools_remote";
+const CROSSWALK_REGEXP = `(\\S*)${CROSSWALK_SOCKET_SUFFIX}`;
 
 let helpers = {};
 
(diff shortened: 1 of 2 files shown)


lib/webview-helpers.js
```diff
@@ -16,23 +18,35 @@ let helpers = {};
...
       }
-    } else if (webviewPid) {
-      // for multiple webviews a list of 'WEBVIEW_<index>' will be returned
-      // where <index> is zero based (same is in selendroid)
-      webviews.push(WEBVIEW_BASE + webviewPid[1]);
+    } else {
+      let webviewPid;
+      let crosswalkWebviewSocket;
+
+      if ((webviewPid = line.match(new RegExp(WEBVIEW_REGEXP)))) {
```
jlipps: this block looks duplicated; should we pull it out of the conditional so it's not?

blutter:
Thanks. It shouldn't hurt to merge this before chromedriver is updated as it is already broken for crosswalk webviews. The only risk is that the chromedriver patch gets some further changes down the track that break my changes. But given that it is already broken I would prefer that this change goes in now as it means one less thing for my team to maintain/patch.

lib/webview-helpers.js
blutter: Sure - I originally pulled it up out of the conditional but didn't think it was as readable.

I'll have another go at refactoring.

blutter: ok - updated with an alternative to avoid DRY

lib/webview-helpers.js
```diff
@@ -16,23 +18,36 @@ let helpers = {};
 // same as chrome-backed webviews)
 // TODO: some of this function belongs in appium-adb
 async function webviewsFromProcs (adb, deviceSocket) {
+  const WEBVIEW_REGEXP = `@?webview_devtools_remote_(\\d+)`;
```
imurchie: Why not just make this a RegExp object and skip the step in line 38. It could also go above wth the rest of the constants.

lib/webview-helpers.js
```diff
@@ -44,6 +59,14 @@ async function webviewsFromProcs (adb, deviceSocket) {
 // not some other running app
 // TODO: this should be called procFromPid and exist in appium-adb
 async function procFromWebview (adb, webview) {
+  if (webview.match(new RegExp(WEBVIEW_BASE + "(\\d+)")) === null) {
```
imurchie: Use string interpolation instead of concatenating strings
```js
`${WEBVIEW_BASE}(\\d+)`
```

This is also constant, so could be a constant like the other regular expressions.

lib/webview-helpers.js
```diff
@@ -16,23 +18,36 @@ let helpers = {};
...
+    let crosswalkWebviewSocket;
+    if ((webviewPid = line.match(new RegExp(WEBVIEW_REGEXP)))) {
       // for multiple webviews a list of 'WEBVIEW_<index>' will be returned
       // where <index> is zero based (same is in selendroid)
       webviews.push(WEBVIEW_BASE + webviewPid[1]);
+    } else if ((crosswalkWebviewSocket = line.match(new RegExp("@" + CROSSWALK_REGEXP)))) {
+      if (deviceSocket)
+      {
+        if (crosswalkWebviewSocket[0].slice(1) === deviceSocket) {
+          webviews.push(WEBVIEW_BASE + crosswalkWebviewSocket[1]);
```
imurchie: String interpolation here and line 49.

lib/webview-helpers.js
blutter: Done.

Though IMHO I consider this less readable as the scope of the constant is only the function.

lib/webview-helpers.js
blutter: Thanks - Done

lib/webview-helpers.js
blutter: Done as well

blutter:
I've also fixed the eslint issue.

PR #239, opened 2017-05-18
Only set mock location helper for real device

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/8487

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -467,11 +467,11 @@ helpers.initDevice = async function (adb, opts) {
   }
   if (!opts.avd) {
     await helpers.pushSettingsApp(adb);
+    await helpers.setMockLocationApp(adb, SETTINGS_HELPER_PKG_ID);
   }
   if (_.isUndefined(opts.unlockType)) {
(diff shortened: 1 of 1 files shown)


lib/android-helpers.js
```diff
@@ -467,11 +467,11 @@ helpers.initDevice = async function (adb, opts) {
   }
   if (!opts.avd) {
```
vrunoa: @mykola-mokhnach would it be better to call `isEmulator` ?

mykola-mokhnach: no, helpers don't have access to driver context

vrunoa: we seen `opts.avd` fail with @imurchie,  since you can avoid setting the avd capability if the emulator is already running

mykola-mokhnach: then it should be another bug. We have to synchronize **avd** capability value in such case

jlipps: > no, helpers don't have access to driver context

we could pass in the result of `isEmulator()` to this function as a param though...

PR #241, opened 2017-05-31
Add remoteAdbHost to desired caps

serggl:
Promote `remoteAdbHost` as a desired capability (https://github.com/appium/appium/issues/8564)

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -176,8 +176,8 @@ helpers.getDeviceInfoFromCaps = async function (opts = {}) {
 };
 
 // returns a new adb instance with deviceId set
-helpers.createADB = async function (javaVersion, udid, emPort, remoteAdbHost, adbPort, suppressKillServer) {
-  let adb = await ADB.createADB({javaVersion, remoteAdbHost, adbPort, suppressKillServer});
+helpers.createADB = async function (javaVersion, udid, emPort, adbPort, suppressKillServer, remoteAdbHost) {
+  let adb = await ADB.createADB({javaVersion, adbPort, suppressKillServer, remoteAdbHost});
 
   adb.setDeviceId(udid);
   if (emPort) {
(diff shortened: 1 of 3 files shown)


lib/android-helpers.js
```diff
@@ -175,8 +176,8 @@ helpers.getDeviceInfoFromCaps = async function (opts = {}) {
 };
 
 // returns a new adb instance with deviceId set
-helpers.createADB = async function (javaVersion, udid, emPort, adbPort, suppressKillServer) {
-  let adb = await ADB.createADB({javaVersion, adbPort, suppressKillServer});
+helpers.createADB = async function (javaVersion, udid, emPort, remoteAdbHost, adbPort, suppressKillServer) {
+  let adb = await ADB.createADB({javaVersion, remoteAdbHost, adbPort, suppressKillServer});
```
dpgraham: Can we just move `remoteAdbHost` after `suppressKillServer` to make it a non-breaking change?

imurchie:
To mitigate that, and because the parameter list is getting pretty long, it could be changed into an options hash.

lib/android-helpers.js
serggl: @dpgraham agree. Just pushed a change

dpgraham:
@imurchie Do you want just the last variable to be a hash? Like this:

`helpers.createADB = async function (javaVersion, udid, emPort, adbPort, suppressKillServer, {remoteAdbHost})`

That would maintain backward compatibility.

Otherwise we make all the args a hash like this

`+helpers.createADB = async function ({javaVersion, udid, emPort, adbPort, suppressKillServer, remoteAdbHost})`

And then we make this a major version bump and refactor any code that uses this to the new format.

imurchie:
I think the change that was made, suggested by @dpgraham, is sufficient.

imurchie:
Published in `appium-android-driver@1.21.0`.

PR #243, opened 2017-06-05
patch to pass android package name to chromedriver

AnishKumarMB:
### Modification
Modified lib/commands/context.js setupNewChromedriver() method to accept the androidPackage
passed via opts.chromeOptions passed by the client in desiredCapabilities.

### Reviewers
@dpgraham

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -212,9 +212,10 @@ async function setupNewChromedriver (opts, curDeviceId, adbPort) {
     adbPort
   };
   let chromedriver = new Chromedriver(chromeArgs);
+  let appPackage = (opts.chromeOptions && opts.chromeOptions.androidPackage) || opts.appPackage;
   let caps = {
     chromeOptions: {
-      androidPackage: opts.appPackage,
+      androidPackage: appPackage,
     }
   };
   if (opts.chromeUseRunningApp) {


imurchie:
Is it possible to sign the CLA? Without it, it is not possible to accept the patch. :(

AnishKumarMB:
I have signed the CLA. Thanks.

dpgraham:
Published to https://www.npmjs.com/package/appium-android-driver version 1.21.1

lib/commands/context.js
```diff
@@ -212,9 +212,10 @@ async function setupNewChromedriver (opts, curDeviceId, adbPort) {
     adbPort
   };
   let chromedriver = new Chromedriver(chromeArgs);
+  let appPackage = (opts.chromeOptions && opts.chromeOptions.androidPackage) || opts.appPackage;
```
jlipps: if you read the code below, it apparently also allows `chromeAndroidPackage` to be on `opts`; should we handle that case as well?

AnishKumarMB: @jlipps 
Thank you for the comment.
As per https://sites.google.com/a/chromium.org/chromedriver/getting-started/getting-started---android
Android-specific Desired Capabilities setting androidPackage solved the issue, but as you said if chromeAndroidPackage can also be set by client then yes, I think same needs to be handled.

PR #245, opened 2017-06-14
Add method to get current package

imurchie:
Add method for getting the current package. Also change tests to use these methods instead of bypassing and going to the `adb` instance. 

Route is in https://github.com/appium/appium-base-driver/pull/123

Second step in https://github.com/appium/appium/issues/5493


imurchie:
Published in `appium-android-driver@1.22.0`.

PR #247, opened 2017-06-19
Fix screenshoting with exec-out

mykola-mokhnach:
Trying to address the issue visible in https://gist.github.com/ajaj91/a457b8b378481f9ddee39434f3343785:

> [AndroidDriver] Cannot get screenshot data with 'adb exec-out' because of 'Command output is not a valid PNG file (fd504e47 !== 89504e47)'. Defaulting to 'adb shell' call

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -183,37 +183,39 @@ commands.sendSMS = async function (phoneNumber, message) {
   await this.adb.sendSMS(phoneNumber, message);
 };
 
-async function storeScreenshotDataWithAdbShell (adb, opts, localFile) {
-  let pngDir = opts.androidScreenshotPath || '/data/local/tmp/';
-  const png = path.posix.resolve(pngDir, 'screenshot.png');
-  let cmd = ['/system/bin/rm', `${png};`, '/system/bin/screencap', '-p', png];
-  await adb.shell(cmd);
-  await adb.pull(png, localFile);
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
this still requires some fine-tuning

mykola-mokhnach:
It still does not work as expected. Probably, because of an issue with teen_process implementation

lib/commands/actions.js
```diff
@@ -185,64 +183,61 @@ commands.sendSMS = async function (phoneNumber, message) {
...
+    } catch (err) {
+      log.warn(`Could not rotate screenshot due to error: ${err}`);
     }
-    let b64data = (await B.promisify(image.getBuffer).call(image, jimp.MIME_PNG))
-                  .toString('base64');
-    return b64data;
-  } finally {
-    await fs.unlink(localFile);
   }
+  return (await B.promisify(image.getBuffer).call(image, jimp.MIME_PNG)).toString('base64');
```
jlipps: while nice and compact, this is a bit less readable than:

```js
const getBuffer = B.promisify(image.getBuffer, {context: image});
const imgBuffer = await getBuffer(jimp.MIME_PNG);
return imgBuffer.toString('base64');
```

mykola-mokhnach: thanks, changed

mykola-mokhnach:
@jlipps yep, it finally works with updated teen_process

PR #248, opened 2017-06-27
Do not try to pull zero-size screenshots

mykola-mokhnach:
Do not try to pull/handle zero-sized screenshots. The screencap utility may return such if one tries to take a screenshot where this is prohibited by Android API. And there is a [complain](https://github.com/appium/appium/issues/8697) about unexpected freeze after one tries to pull zero-size screenshot from the device.


imurchie:
I'm working on the CI problem elsewhere. Merging this.

PR #252, opened 2017-07-17
Preset Shared preferences

vrunoa:
Add possibility to preset apps shared preferences. We can set specific shared preferences for apps right before launching it. I have a functional test running locally, let me know if there's any app on appium repos where I can I add this test. 

This implementation will take the sharedPreferences object capability and create an xml that it will be push to the app shared_prefs space before running the application.


imurchie:
Most of the tests use https://github.com/appium/android-apidemos

lib/driver.js
```diff
@@ -435,6 +442,43 @@ class AndroidDriver extends BaseDriver {
     }
   }
 
+  async setSharedPreferences () {
+    let sharedPrefs = this.opts.sharedPreferences;
```
jlipps: does the builder mutate `sharedPrefs`? if so it will also mutate `this.opts.sharedPreferences`. in that case we'd want to clone rather than assign. but i don't know how the builder works.

lib/driver.js
```diff
@@ -435,6 +442,43 @@ class AndroidDriver extends BaseDriver {
...
+      if (await fs.exists(localPath)) {
+        await fs.unlink(localPath);
+      }
+    } catch (e) {
+      log.warn(`Error trying to remove temporary file ${localPath}`);
+    }
+    return true;
+  }
+
+  getPrefsBuilder () {
```
jlipps: is there any reason to create and reuse this rather than just create a new instance when we call `setSharedPreferences`? don't we only call this one per session anyway?

lib/driver.js
vrunoa: @jlipps no, it takes the sharedPrefs object and creates an _.xml_ file
https://github.com/vrunoa/shared-preferences-builder/blob/master/lib/index.js#L69-L82.

lib/driver.js
vrunoa: is the AndroidDriver created every session ? if that's the case then yes, we can just create a new instance

lib/driver.js
vrunoa: btw, let me know if you guys want to have the shared-preferences-builder repo inside appium. I can transfer the repo to you guys.

jlipps: it's fine where it is for now

lib/driver.js
jlipps: yes, one new instance of AndroidDriver for every session. so i think we should just simplify this.

lib/driver.js
```diff
@@ -473,10 +473,7 @@ class AndroidDriver extends BaseDriver {
   }
 
   getPrefsBuilder () {
```
jlipps: i think we can just delete this getPrefsBuilder method now that it doesn't do anything

vrunoa: @jlipps I tried, but couldnt find a way to mock `new SharedPrefsBuilder`. I was going to ask you guys today

jlipps: ah ok, if it's to make it testable than i'm fine with it. maybe just add a comment to that effect?

vrunoa: sure, maybe @imurchie knows how to mock `new SharedPrefsBuilder` with sinon. I tried some implementations but none of them worked 😞

imurchie: No, new objects are difficult to mock.

vrunoa: ok, i'll add a comment on `getPrefsBuilder` then

lib/driver.js
```diff
@@ -325,6 +326,12 @@ class AndroidDriver extends BaseDriver {
     await this.grantPermissions();
     this.apkStrings[this.opts.language] = await helpers.pushStrings(
         this.opts.language, this.adb, this.opts);
+
+    // This must run after installing the apk, otherwise it would make the
+    // install to fail. And before running the app.
```
jlipps: grammar nit: you can say `otherwise it would cause the install to fail`, or `otherwise it would make the install fail`

imurchie:
Ack! This has gotten out of sync. Can you rebase and make sure linting passes? I think the problem is somewhere along the line you re-introduced `version` into `lib/driver.js`.

imurchie:
Published in `appium-android-driver@1.29.0`.

vrunoa:
thanks @imurchie! sorry I couldnt finish this implementation.

imurchie:
No problem. It was a simple fix.

PR #254, opened 2017-08-04
Add gpsEnabled capability

vrunoa:
Add possibility to enable/disable GPS on emulators before running apps by adding gpsEnabled capability.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -153,6 +153,16 @@ class AndroidDriver extends BaseDriver {
         await this.checkPackagePresent();
       }
 
+      // check if we have to enable/disable gps before running the application
+      if (this.opts.gpsEnabled) {
+        if (this.isEmulator()) {
+          log.info(`Trying to ${this.opts.gpsEnabled ? "enable": "disable"} gps location provider`);
+          await this.adb.toggleGPSLocationProvider(this.opts.gpsEnabled);
+        } else {
+          log.warn('Sorry! gpsEnabled capability is only available for emulators');
+        }
+      }
+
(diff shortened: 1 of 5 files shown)


lib/driver.js
```diff
@@ -153,6 +153,16 @@ class AndroidDriver extends BaseDriver {
         await this.checkPackagePresent();
       }
 
+      // check if we have to enable/disable gps before running the application
+      if (this.opts.gpsEnabled) {
```
jlipps: probably want `if (util.hasValue(this.opts.gpsEnabled)) {`, otherwise if someone sets it to `false` this will never pass

test/unit/commands/network-specs.js
```diff
@@ -112,21 +112,32 @@ describe('Network', () => {
...
       let sequence = [22, 22, 19];
       sandbox.stub(driver.adb, 'getApiLevel').returns(19);
+      sandbox.stub(driver, 'isEmulator').returns(false);
       sandbox.stub(driver.adb, 'keyevent');
       await driver.toggleLocationServices();
       driver.adb.keyevent.calledWithExactly(19).should.be.true;
       driver.toggleSetting.calledWith('LOCATION_SOURCE_SETTINGS', sequence).should.be.true;
     });
+    /*
+    it('should set gps for emulators', async () => {
```
jlipps: why's the code commented out?

vrunoa: still have to update the appium-adb version, waiting for a PR in there. I forgot to add it to this PR description

PR #261, opened 2017-08-17
Launch settings app

vrunoa:
Real-devices on Android 7+ seems to be failing to set wifi, data settings if the io.appium.settings app is not launched prior to starting the session. 
https://github.com/appium/appium/issues/8957

Would like to have @mayureshshirodkar confirmation to verify this is not happening on Android 6 or below devices. Right now i'm out of devices.  😄

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -367,7 +367,7 @@ helpers.pushSettingsApp = async function (adb) {
   // if the app is not launched prior to start the session on android 7+
   // see https://github.com/appium/appium/issues/8957
   try {
-    if ((await adb.getApiLevel()) >= 24) {
+    if ((await adb.getApiLevel()) >= 23) {
       let startOpts = {
         pkg: "io.appium.settings",
         activity: ".Settings",
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -363,6 +363,24 @@ helpers.pushSettingsApp = async function (adb) {
   } catch (err) {
     // errors are expected there, since the app contains non-changeable permissons
   }
+  // lauch io.appium.settings app due to settings failing to be set
+  // if the app is not launched prior to start the session on android 7+
+  // see https://github.com/appium/appium/issues/8957
+  try {
+    if ((await adb.getApiLevel()) >= 24) {
+      let startOpts = {
+        pkg: "io.appium.settings",
+        activity: ".",
+        action: "android.intent.action.MAIN",
+        category: "android.intent.category.LAUNCHER",
+        flags: "0x10200000",
+        stopApp: false
+      };
+      await adb.startApp(startOpts);
+    }
+  } catch (err) {
+    logger.warn(`Failed to launch settings app: "${err.message}".`);
+  }
 };
 
 helpers.pushUnlock = async function (adb) {
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -368,7 +368,7 @@ helpers.pushSettingsApp = async function (adb) {
   // if the app is not launched prior to start the session on android 7+
   // see https://github.com/appium/appium/issues/8957
   try {
-    if ((await adb.getApiLevel()) >= 23) {
+    if (!(await adb.processExists(SETTINGS_HELPER_PKG_ID))) {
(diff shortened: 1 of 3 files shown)


vrunoa:
It would be great if we can confirm this is also happening on lower android real devices;

* [ ] 4.4
* [ ] 5.0
* [ ] 5.1

@mayureshshirodkar want to tackle this ?

lib/android-helpers.js
```diff
@@ -363,6 +363,24 @@ helpers.pushSettingsApp = async function (adb) {
...
+    if ((await adb.getApiLevel()) >= 23) {
+      let startOpts = {
+        pkg: "io.appium.settings",
+        activity: ".Settings",
+        action: "android.intent.action.MAIN",
+        category: "android.intent.category.LAUNCHER",
+        flags: "0x10200000",
+        stopApp: false
+      };
+      await adb.startApp(startOpts);
```
mykola-mokhnach: does it make any difference if the app is already running?

lib/android-helpers.js
```diff
@@ -363,6 +363,24 @@ helpers.pushSettingsApp = async function (adb) {
   } catch (err) {
     // errors are expected there, since the app contains non-changeable permissons
   }
+  // lauch io.appium.settings app due to settings failing to be set
+  // if the app is not launched prior to start the session on android 7+
+  // see https://github.com/appium/appium/issues/8957
+  try {
+    if ((await adb.getApiLevel()) >= 23) {
+      let startOpts = {
+        pkg: "io.appium.settings",
```
mykola-mokhnach: use **SETTINGS_HELPER_PKG_ID** constant

test/functional/settings-e2e-specs.js
```diff
@@ -0,0 +1,43 @@
...
+
+chai.should();
+chai.use(chaiAsPromised);
+
+let defaultCaps = _.defaults({
+  androidInstallTimeout: 90000,
+  browserName: 'chrome'
+}, DEFAULT_CAPS);
+
+describe.skip('toggle wifi tests', () => {
```
mykola-mokhnach: why to skip here? We only want to skip this test for emulator - right?

lib/android-helpers.js
vrunoa: good point, I could check if the app is already running and skip it.

test/functional/settings-e2e-specs.js
vrunoa: @mykola-mokhnach yes. But I didnt knew if we have real device on the appium ci. How can we make sure this runs only on real-devices ?

mayureshshirodkar:
@vrunoa I have checked on below OS with Android real device.
4.4.2
5.0.1
5.1
6.0.1
I am able to replicate the issue on all of these.

test/functional/settings-e2e-specs.js
mykola-mokhnach: there should be an env variable to check that. @imurchie knows more for sure

imurchie: Generally speaking with set the `REAL_DEVICE` environment variable when running on real devices. So we skip like https://github.com/appium/appium-android-driver/blob/master/test/functional/webview-browser-tester-e2e-specs.js#L26

you can also skip based on `process.env.TRAVIS` to not run in CI.

imurchie: We really ought to set `process.env.REAL_DEVICE` as false when `process.env.TRAVIS` is true.

test/functional/settings-e2e-specs.js
```diff
@@ -0,0 +1,49 @@
...
+let defaultCaps = _.defaults({
+  androidInstallTimeout: 90000,
+  browserName: 'chrome'
+}, DEFAULT_CAPS);
+
+describe('toggle wifi tests', () => {
+  let driver;
+
+  describe('functional', () => {
+    before(() => {
```
imurchie: `this` is only the mocha context if it is in a normal function. So this needs to be 
```js
before(function () {
```

imurchie:
Published in `appium-android-driver@1.25.0`.

PR #265, opened 2017-08-22
change the adbPort parameter to an adb instance

babyrjw:
change the adbPort parameter to an adb instance


imurchie:
This PR just needs to add a change to the `appium-chromedriver` version in `package.json` (https://github.com/appium/appium-android-driver/blob/master/package.json#L35) to `^3.0.0`.

imurchie:
Functionality added in #269

PR #266, opened 2017-08-22
Log error but do not fail if mock location setting fails

imurchie:
Do not fail the entire startup process for a failing mock location setup.


imurchie:
See https://github.com/appium/appium/issues/9059

imurchie:
Published in `appium-android-driver@1.25.1`.

PR #270, opened 2017-09-13
Fix parsing of webview process name

imurchie:
Rather than parsing the process name by looking at placement in relation to the headers (which is problematic, since there are sometimes no headers for particular fields, and also sometimes there are no entries for fields), use the absolute position in the line, in relation to the header.


imurchie:
Related to part of https://github.com/appium/appium/issues/8958

lib/webview-helpers.js
```diff
@@ -80,29 +81,60 @@ async function procFromWebview (adb, webview) {
...
+  let header = lines[0];
+
   // the column order may not be identical on all androids
   // dynamically locate the pid and name column.
-  let pidColumn = header.indexOf("PID");
-  let pkgColumn = header.indexOf("NAME") + 1;
+  let pidIndex = header.indexOf('PID');
+  let pkgIndex = header.indexOf('NAME');
+
+  function slice (line, index) {
```
jlipps: i eventually figured out what this function does, but a little comment describing what its purpose is would be helpful.

imurchie: Fixed!

PR #272, opened 2017-09-21
Add support to force upgrade settings app

imurchie:
Force-upgrade settings app. We don't care about its data, so there is no reason to do a real upgrade, which fails in some cases (such as when the device has settings app version 1 on it).

Also add a flag to the function, to aid in testing.

### DIFF
--- a/test/functional/android-helper-e2e-specs.js
+++ b/test/functional/android-helper-e2e-specs.js
@@ -5,6 +5,7 @@ import helpers from '../../lib/android-helpers';
 import ADB from 'appium-adb';
 import { app } from './desired';
 import { MOCHA_TIMEOUT } from './helpers';
+import { exec } from 'teen_process';
 
 
 let opts = {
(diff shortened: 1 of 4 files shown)


test/functional/android-helper-e2e-specs.js
```diff
@@ -54,4 +52,19 @@ describe('android-helpers e2e', () => {
...
     });
   });
+  describe('pushSettingsApp', function () {
+    const settingsPkg = 'io.appium.settings';
+    it('should be able to upgrade from settings v1 to latest', async function () {
+      await adb.uninstallApk(settingsPkg);
+
+      // get and install old version of settings app
+      await exec('npm', ['install', `${settingsPkg}@2.0.0`]);
+      await helpers.pushSettingsApp(adb, true).should.eventually.be.rejected;
```
mykola-mokhnach: I thought that _eventually_ makes _await_ redundant, isn't it?

imurchie: Kind of. But without the `await` the execution would just go on to the next command without waiting.

imurchie: Essentialy `eventually` just puts a `catch` on the promise that is returned, with an assertion inside that that block gets run. We need to `await` still so we don't move on until the promise is fulfilled (hopefully rejected).

PR #273, opened 2017-09-22
Always assume the file to be pushed contains binary data

mykola-mokhnach:
Improve the way files are pushed and pulled to/from the device


imurchie:
Published in `appium-android-driver@1.28.1`.

PR #274, opened 2017-10-03
Update exec-out screenshot

imurchie:
Add error handling to try to be able to trace problems like https://github.com/appium/appium/issues/9369

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -283,7 +283,10 @@ commands.getScreenshot = async function () {
     try {
       image = await this.getScreenshotDataWithAdbShell(this.adb, this.opts);
     } catch (e) {
-      log.errorAndThrow(`Cannot get screenshot data because of '${e.message}'`);
+      const err = `Cannot get screenshot data because of '${e.message}'. ` +
+                  `Make sure the 'LayoutParams.FLAG_SECURE' is not set for ` +
+                  `the current view`;
+      log.errorAndThrow(err);
     }
   }
   if (apiLevel < 23) {
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -250,11 +250,19 @@ helpers.getScreenshotDataWithAdbShell = async function (adb, opts) {
 };
 
 helpers.getScreenshotDataWithAdbExecOut = async function (adb) {
-  let {stdout} = await exec(adb.executable.path, ['exec-out', '/system/bin/screencap -p'],
-                            {encoding: 'binary', isBuffer: true});
-  if (!stdout.length) {
-    throw new Error('The size of the taken screenshot equals to zero.');
+  let {stdout, stderr, code} = await exec(adb.executable.path,
+                                    adb.executable.defaultArgs
+                                      .concat(['exec-out', '/system/bin/screencap', '-p']),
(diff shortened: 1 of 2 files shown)


lib/commands/actions.js
```diff
@@ -250,11 +250,19 @@ helpers.getScreenshotDataWithAdbShell = async function (adb, opts) {
...
+                                      .concat(['exec-out', '/system/bin/screencap', '-p']),
+                                    {encoding: 'binary', isBuffer: true});
+  // if there is an error, throw
+  if (code || stderr.length) {
+    throw new Error(`Screenshot returned error, code: '${code}', stderr: '${stderr.toString()}'`);
+  }
+  // if we don't get anything at all, throw
   if (!stdout.length) {
-    throw new Error('The size of the taken screenshot equals to zero.');
+    throw new Error('Screenshot returned no data');
```
mykola-mokhnach: I'd also add an explanation text here like "Make sure the LayoutParams.FLAG_SECURE is not set for the current view". There is also the second place where this error might be show below in the _getScreenshotDataWithAdbShell_ method

lib/commands/actions.js
```diff
@@ -250,11 +250,19 @@ helpers.getScreenshotDataWithAdbShell = async function (adb, opts) {
...
 
 helpers.getScreenshotDataWithAdbExecOut = async function (adb) {
-  let {stdout} = await exec(adb.executable.path, ['exec-out', '/system/bin/screencap -p'],
-                            {encoding: 'binary', isBuffer: true});
+  let {stdout, stderr, code} = await exec(adb.executable.path,
+                                    adb.executable.defaultArgs
+                                      .concat(['exec-out', '/system/bin/screencap', '-p']),
+                                    {encoding: 'binary', isBuffer: true});
+  // if there is an error, throw
+  if (code || stderr.length) {
```
mykola-mokhnach: I assume exec throws an error automatically if code !== 0

imurchie: It should.

lib/commands/actions.js
imurchie: That message can be handled a little lower. And the error handling for the `adb shell` path is handled in `appium-adb`. Since that is bypassed for `exec-out` it needs to be handled here.

lib/commands/actions.js
mykola-mokhnach: which means the interpreter will never reach this block of code if code !== 0

lib/commands/actions.js
```diff
@@ -250,11 +250,19 @@ helpers.getScreenshotDataWithAdbShell = async function (adb, opts) {
 };
 
 helpers.getScreenshotDataWithAdbExecOut = async function (adb) {
-  let {stdout} = await exec(adb.executable.path, ['exec-out', '/system/bin/screencap -p'],
-                            {encoding: 'binary', isBuffer: true});
+  let {stdout, stderr, code} = await exec(adb.executable.path,
+                                    adb.executable.defaultArgs
+                                      .concat(['exec-out', '/system/bin/screencap', '-p']),
```
mykola-mokhnach: one can use spreading operator instead of concat

lib/commands/actions.js
imurchie: Perhaps.

PR #275, opened 2017-10-05
Add initial orientation functionality

imurchie:
Make `orientation`  capability work. Resolves https://github.com/appium/appium/issues/9391


lib/driver.js
```diff
@@ -278,6 +278,12 @@ class AndroidDriver extends BaseDriver {
         await this.startAUT();
       }
     }
+
+    if (util.hasValue(this.opts.orientation)) {
```
mykola-mokhnach: Does this also work for UIA2?

PR #276, opened 2017-10-06
Use adb rotate for setting emus initial orientation

vrunoa:
[WORK in PROGRESS]
Still want to add some unit & functional test, but I would like a quick feedback of this implementation @mykola-mokhnach @imurchie

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -377,12 +377,13 @@ helpers.setMockLocationApp = async function (adb, app) {
   }
 };
 
-helpers.pushSettingsApp = async function (adb) {
+helpers.pushSettingsApp = async function (adb, throwError = false) {
   logger.debug("Pushing settings apk to device...");
+
   try {
-    await adb.installOrUpgrade(settingsApkPath, SETTINGS_HELPER_PKG_ID);
+    await adb.installOrUpgrade(settingsApkPath, SETTINGS_HELPER_PKG_ID, true);
   } catch (err) {
-    logger.warn(`Ignored error while installing Appium Settings helper: "${err.message}". ` +
+    logger.warn(`Ignored error while installing Appium Settings helper: '${err.message}'. ` +
                 `Expect some Appium features may not work as expected unless this problem is fixed.`);
   }
   try {
@@ -406,7 +407,10 @@ helpers.pushSettingsApp = async function (adb) {
       await adb.startApp(startOpts);
     }
   } catch (err) {
-    logger.warn(`Failed to launch settings app: "${err.message}".`);
+    logger.warn(`Failed to launch settings app: ${err.message}`);
+    if (throwError) {
+      throw err;
+    }
   }
 };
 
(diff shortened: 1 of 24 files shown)


lib/commands/actions.js
```diff
@@ -43,6 +43,15 @@ commands.setOrientation = async function (orientation) {
   return await this.bootstrap.sendAction("orientation", params);
 };
 
+commands.setEmuOrientation = async function (adb, orientation) {
```
imurchie: This should be on the `extensions` or `helpers` object, as it is not technically a command.

Also, since it eventually gets called in this context of the driver, there is no need to pass in `adb`. `this.adb` will be the same here as in the calling code.

lib/commands/actions.js
```diff
@@ -43,6 +43,15 @@ commands.setOrientation = async function (orientation) {
...
+  orientation = orientation.toUpperCase();
+  let currentOrientation = await this.getOrientation();
+  if (currentOrientation !== orientation) {
+    await adb.rotate();
+    await this.setEmuOrientation(adb, orientation);
```
imurchie: Is this because more than one rotation may be needed? A comment might clarify.

vrunoa: @imurchie I think I confuse this with ROTATION, does getOrientation() only returns `landscape` or `portarit` ?

test/functional/driver-e2e-specs.js
```diff
@@ -161,3 +161,23 @@ describe('close', function () {
...
+    await driver.deleteSession();
+  });
+  it('check emulator is rotating', async function () {
+    let caps = _.defaults({
+      androidInstallTimeout: 90000,
+      avd: 'Pixel_API_26',
+      orientation: 'landscape'
+    }, DEFAULT_CAPS);
+    await driver.createSession(caps);
+    let orientation  = await driver.getOrientation();
```
mykola-mokhnach: extra space here

lib/android-helpers.js
```diff
@@ -574,6 +574,15 @@ helpers.getChromePkg = function (browser) {
   return {pkg, activity};
 };
 
+helpers.setEmuOrientation = async function (orientation) {
+  orientation = orientation.toUpperCase();
+  let currentOrientation = await this.getOrientation();
+  if (currentOrientation !== orientation) {
+    await this.adb.rotate();
+    await this.setEmuOrientation(orientation);
```
mykola-mokhnach: is this a recursive call?

test/functional/driver-e2e-specs.js
```diff
@@ -161,3 +161,26 @@ describe('close', function () {
...
+    let caps = _.defaults({
+      androidInstallTimeout: 90000,
+      avd: 'Pixel_API_26',
+      orientation: 'landscape'
+    }, DEFAULT_CAPS);
+    caps.appPackage = 'io.appium.android.apis';
+    caps.appActivity = 'io.appium.android.apis.app.HelloWorld';
+    caps.intentCategory = 'appium.android.intent.category.SAMPLE_CODE';
+    await driver.createSession(caps);
+    let orientation  = await driver.getOrientation();
```
mykola-mokhnach: extra space here

vrunoa:
Made a mess with merge 🤦‍♂️  Going to start another PR

PR #277, opened 2017-10-06
Handle incompatible unlock helper during install

imurchie:
If installing the unlock helper fails because an old version is there (as in https://github.com/appium/appium/issues/9394), uninstall the old version and install the new one.


lib/android-helpers.js
```diff
@@ -416,7 +419,18 @@ helpers.pushSettingsApp = async function (adb, throwError = false) {
 
 helpers.pushUnlock = async function (adb) {
   logger.debug("Pushing unlock helper app to device...");
-  await adb.install(unlockApkPath, false);
+  try {
+    await adb.install(unlockApkPath, false);
```
mykola-mokhnach: there is installOrUpgrade method with force argument, which does this as well

PR #278, opened 2017-10-16
Fix handling of settings and unlock app pushing

imurchie:
Update logic for making sure we can install over the unlock and settings app.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -384,7 +384,7 @@ helpers.pushSettingsApp = async function (adb, throwError = false) {
   logger.debug("Pushing settings apk to device...");
 
   try {
-    await adb.installOrUpgrade(settingsApkPath, SETTINGS_HELPER_PKG_ID, true);
+    await adb.installOrUpgrade(settingsApkPath, SETTINGS_HELPER_PKG_ID);
   } catch (err) {
     logger.warn(`Ignored error while installing Appium Settings helper: '${err.message}'. ` +
                 `Expect some Appium features may not work as expected unless this problem is fixed.`);
@@ -419,7 +419,19 @@ helpers.pushSettingsApp = async function (adb, throwError = false) {
 
 helpers.pushUnlock = async function (adb) {
   logger.debug("Pushing unlock helper app to device...");
-  await adb.installOrUpgrade(unlockApkPath, UNLOCK_HELPER_PKG_ID, true);
+
+  // the unlock helpers does not have the correct versioning, so
+  // `adb.installOrUpgrade` will not work
+  try {
+    await adb.install(unlockApkPath, false);
+  } catch (err) {
+    if (!err.message.includes('INSTALL_FAILED_UPDATE_INCOMPATIBLE')) {
+      throw err;
+    }
+    logger.debug(`Incompatible unlock helper found, remove old version and re-installing`);
+    await adb.uninstallApk(UNLOCK_HELPER_PKG_ID);
+    await adb.install(unlockApkPath, false);
+  }
 };
 
 // pushStrings method extracts string.xml and converts it to string.json and pushes
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -380,15 +380,23 @@ helpers.setMockLocationApp = async function (adb, app) {
   }
 };
 
+helpers.installHelperApp = async function (adb, apkPath, packageId, appName) {
+  try {
+    await adb.installOrUpgrade(apkPath, packageId);
+  } catch (err) {
+    logger.warn(`Ignored error while installing Appium ${appName} helper: ` +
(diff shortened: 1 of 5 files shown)


imurchie:
Related to https://github.com/appium/appium/issues/9458

lib/android-helpers.js
```diff
@@ -419,7 +419,19 @@ helpers.pushSettingsApp = async function (adb, throwError = false) {
...
 helpers.pushUnlock = async function (adb) {
   logger.debug("Pushing unlock helper app to device...");
-  await adb.installOrUpgrade(unlockApkPath, UNLOCK_HELPER_PKG_ID, true);
+
+  // the unlock helpers does not have the correct versioning, so
+  // `adb.installOrUpgrade` will not work
+  try {
+    await adb.install(unlockApkPath, false);
+  } catch (err) {
+    if (!err.message.includes('INSTALL_FAILED_UPDATE_INCOMPATIBLE')) {
```
mykola-mokhnach: shouldn't we check if err.message exists first?

lib/android-helpers.js
```diff
@@ -419,7 +419,19 @@ helpers.pushSettingsApp = async function (adb, throwError = false) {
...
+      throw err;
+    }
+    logger.debug(`Incompatible unlock helper found, remove old version and re-installing`);
```
jlipps: `removing`

lib/android-helpers.js
```diff
@@ -419,7 +419,19 @@ helpers.pushSettingsApp = async function (adb, throwError = false) {
 
 helpers.pushUnlock = async function (adb) {
   logger.debug("Pushing unlock helper app to device...");
-  await adb.installOrUpgrade(unlockApkPath, UNLOCK_HELPER_PKG_ID, true);
+
+  // the unlock helpers does not have the correct versioning, so
+  // `adb.installOrUpgrade` will not work
+  try {
+    await adb.install(unlockApkPath, false);
```
mykola-mokhnach: I'd think we can use the same strategy to install both settings and unlock helpers: 
installOrUpgrade first, uninstall and install if this fails and then just warn instead of throwing exception.
These helpers are not so important in general, that we must stop the whole session because of an error there.

Also, I'd be happy if we could customise the error message that is thrown (if you still would like it to be there). It should be more helpful and should suggest users to uninstall the corresponding tool manually from the device in order to solve the problem.

jlipps: Good ideas

lib/android-helpers.js
imurchie: Sure.

lib/android-helpers.js
imurchie: I'm fine with putting a warning and not throwing an exception, so the session can continue.

As for using the same logic for both, that is not possible. The unlock helper will _never_ be upgraded by the `installOrUpgrade` function, since the version of the app is never updated. Every published version of the `appium-unlock` helper comes out as version 1 when using `aapt`.

mykola-mokhnach: that's sad. Can we improve this tool to have appropriate versioning?
@vrunoa @vmaxim ping

vmaxim: @mykola-mokhnach I get correct `versionCode` from `AndroidManifest.xml` with `aapt`:
```
C:\1\unlock_apk>type AndroidManifest.xml | findStr version*
<?xml version="1.0" encoding="utf-8"?>
      android:versionCode="555"
      android:versionName="1.555">

C:\1\unlock_apk>%android_home%\build-tools\26.0.0\aapt dump badging bin\unlock_apk-debug.apk | findStr version*
package: name='io.appium.unlock' versionCode='555' versionName='1.555' platformBuildVersionName='4.4.2-1456859'```

Comment (vmaxim) [reply lib/android-helpers.js]: @mykola-mokhnach appium/unlock_apk/pull/8

Comment (mykola-mokhnach) [reply lib/android-helpers.js]: cool, thanks @vmaxim 

Now we need to think about how do we deal here with upgrade from older builds, which don't have the version code set

File: lib/android-helpers.js [hunk 4]
```diff
@@ -419,7 +419,13 @@ helpers.pushSettingsApp = async function (adb, throwError = false) {
 
 helpers.pushUnlock = async function (adb) {
   logger.debug("Pushing unlock helper app to device...");
-  await adb.installOrUpgrade(unlockApkPath, UNLOCK_HELPER_PKG_ID, true);
+
+  try {
```
mykola-mokhnach: we could extract settings and unlock helper install into a separate method to avoid unnecessary code duplication

Also, in the warning message we can state the possibility to uninstall the tool manually from the device in order to fix the problem.

imurchie:
Published in `appium-android-driver@1.31.0`.

PR #279, opened 2017-10-20
Updated shared-preferences-builder

dpgraham:
* Version 0.0.3 was broken on Node version 5 because it didn't have 'use strict'


dpgraham:
Published to 1.32.0

PR #282, opened 2017-10-26
Make sure the platformVersion is of String type

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/9531


imurchie:
Published in `appium-android-driver@1.32.1`.

PR #283, opened 2017-10-31
Update appium-adb module

mykola-mokhnach:
now _adb.getApiLevel_ call always returns an integer


imurchie:
Published in `appium-android-driver@1.32.4`.

PR #284, opened 2017-11-10
Add support for setting device time

anthonypdawson:
Implementing setting device time for the Android platform across this repo, appium-base-driver and python-client

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -40,7 +40,7 @@ commands.getDeviceTime = async function () {
 };
 
 commands.setDeviceTime = async function (dateTime) {
-  log.info('Attempting to set device date and time')
+  log.info('Attempting to set device date and time');
   try {
     let out = await this.adb.shell(['date', dateTime]);
     return out.trim();
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -39,6 +39,16 @@ commands.getDeviceTime = async function () {
   }
 };
 
+commands.setDeviceTime = async function (dateTime) {
+  log.info('Attempting to set device date and time')
+  try {
+    let out = await this.adb.shell(['date', dateTime]);
+    return out.trim();
+  } catch (err) {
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -39,6 +39,16 @@ commands.getDeviceTime = async function () {
   }
 };
 
+commands.setDeviceTime = async function (dateTime) {
```
mykola-mokhnach: Please add proper docstring

lib/commands/general.js
```diff
@@ -39,6 +39,16 @@ commands.getDeviceTime = async function () {
...
+  log.info('Attempting to set device date and time');
+  try {
+    let out = await this.adb.shell(['date', dateTime]);
+    return out.trim();
+  } catch (err) {
+    log.errorAndThrow(`Could not set device date and time: ${err}`);
```
mykola-mokhnach: err.message

lib/commands/general.js
```diff
@@ -39,6 +39,16 @@ commands.getDeviceTime = async function () {
   }
 };
 
+commands.setDeviceTime = async function (dateTime) {
+  log.info('Attempting to set device date and time');
```
mykola-mokhnach: I assume a better place for such stuff would be appium-adb module. WDYT @imurchie ?

anthonypdawson: I don't mind making that change, these are my first PRs.  I was basically working alongside getDeviceTime

imurchie: Yes, this ought to be done in the `appium-adb` module. We should not really ever be calling `adb.shell` outside of there, if possible (I know we break that all over the place, but shouldn't :( ).

mykola-mokhnach:
So, as a conclusion:
- The time setting functionality should be moved to appium-adb module
- The command itself should be available via `mobile:` interface or as a capability value

PR #285, opened 2017-11-11
Implemented broadcastIntent endpoint

dpgraham:
(see https://github.com/appium/appium/issues/3172)

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -339,6 +339,10 @@ commands.getSystemBars = async function () {
   return parseWindows(out);
 };
 
+commands.broadcastIntent = async function (intent) {
+  await this.adb.broadcast(intent);
+};
+
 Object.assign(extensions, commands, helpers);
 export { commands, helpers };
 export default extensions;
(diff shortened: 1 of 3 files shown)


lib/commands/general.js
```diff
@@ -339,6 +339,10 @@ commands.getSystemBars = async function () {
   return parseWindows(out);
 };
 
+commands.broadcastIntent = async function (intent) {
```
mykola-mokhnach: sometimes this might to be enough to just have intent name, since broadcasting requires additional arguments to be provided

dpgraham: We would have to update `appium-adb` to include arguments to with `broadcast` (https://developer.android.com/studio/command-line/adb.html#IntentSpec) which we could easily do but I wonder, do you foresee any security issues?

mykola-mokhnach: I don't think so. If we provide arguments as a simple array then these are going to be escaped automatically by the underlying `spawn` call.

mykola-mokhnach:
@dpgraham Shall we close this PR because of https://github.com/appium/appium-android-driver/pull/293?

dpgraham:
Yeah #293 covers this.

PR #288, opened 2017-11-28
Add  a possibility to pull/push files from/to application containers

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/6692


mykola-mokhnach:
I'll extract common stuff into separate methods, but I don't think there's enough to extract into appium-support. It's just about single regexp parsing

mykola-mokhnach:
Closing for now, as I want to do some more tesings on newer Android versions

mykola-mokhnach:
Should be better now

PR #291, opened 2017-12-11
Keep track of args used to start the app, for restart

imurchie:
Currently if an app is opened after startup, and then backgrounded, the original app is opened and the new app is waited for.

So, keep track of the arguments used to open an app, and use them to reopen when backgrounding.


lib/commands/general.js
```diff
@@ -131,17 +132,37 @@ commands.background = async function (seconds) {
...
+      pkg: appPackage,
+      activity: appActivity,
+      waitPkg: appPackage,
+      waitActivity: appActivity,
+      stopApp: false
+    };
+  }
+  args = await util.filterObject(args);
+  log.debug(`Bringing application back to foreground with arguments: ${JSON.stringify(args)}`);
+  return this.adb.startApp(args);
```
mykola-mokhnach: await?

lib/commands/general.js
```diff
@@ -131,17 +132,37 @@ commands.background = async function (seconds) {
...
-    category: this.opts.intentCategory,
-    flags: this.opts.intentFlags,
-    waitPkg: appPackage,
-    waitActivity: appActivity,
-    optionalIntentArguments: this.opts.optionalIntentArguments,
-    stopApp: false
-  });
+
+  let args;
+  if (this.opts.startActivityArgs && this.opts.startActivityArgs[`${appPackage}/${appActivity}`]) {
```
mykola-mokhnach: `!_.isEmpty(this.opts.startActivityArgs)`

imurchie: I'm not sure what that would gain us. This is just to make sure the variable has something in it before accessing.

PR #292, opened 2017-12-11
Always assume that "'press', 'moveTo', 'tap', 'longPress'" actions require absolute coordinates

mykola-mokhnach:
It is a breaking change, but at some point of time this should be done anyway. Addresses https://github.com/appium/appium/issues/7486

### DIFF
--- a/test/functional/commands/touch/drag-e2e-specs.js
+++ b/test/functional/commands/touch/drag-e2e-specs.js
@@ -3,6 +3,7 @@ import chaiAsPromised from 'chai-as-promised';
 import _ from 'lodash';
 import AndroidDriver from '../../../..';
 import DEFAULT_CAPS from '../../desired';
+import B from 'bluebird';
 
 
 chai.should();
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
@imurchie would it be possible to merge this PR to beta after 1.7.2 is published?

test/functional/commands/touch/drag-e2e-specs.js
```diff
@@ -71,8 +72,16 @@ describe('apidemo - touch', function () {
                       {element: endEle.ELEMENT, x: 5, y: 5}},
                       {action: "release", options: {}}];
       await driver.performTouch(gestures);
-      let element3 = await driver.findElement("id", "io.appium.android.apis:id/drag_result_text");
-      await driver.getText(element3.ELEMENT).should.eventually.equal("Dropped!");
+      let element3;
+      const expectedText = 'Dropped!';
+      for (let retries = 0; retries < 3; ++retries) {
```
imurchie: There is an easier way for all this
```js
await retryInterval(3, 500, async () => {
  const el = await driver.findElement("id", "io.appium.android.apis:id/drag_result_text");
  (await driver.getText(el.ELEMENT)).should.eql(expectedText);
});
```

PR #296, opened 2017-12-19
Implements: Keep access to chromedriver after app relaunch

neofreko:
**Current Behavior:**
Automation steps:
1. launch hybrid app
2. switch to webview context
3. close app
4. launch hybrid app
5. switch to webview context
6. query webview element

Result:
Error: chrome is not reachable

**New Behavior:**
Automation steps:

1. launch hybrid app
2. switch to webview context
3. close app
4. launch hybrid app
5. switch to webview context
6. query webview element

Result:
No error thrown. Element can be found.

**Implementation:**

No need to throw error about current instance of chromedriver. The code can already decide when to reuse or recreate chromedriver session.
Reset current context when app is closed thus switchContext can react properly when app relaunched.


imurchie:
Published in `appium-android-driver@1.36.1`.

PR #300, opened 2018-01-09
Add application management primitives

mykola-mokhnach:
This might be useful while multiple applications are affected in a single test or for some complicated scenarios where application upgrade/restart is checked.

### DIFF
--- a/lib/commands/apps-management.js
+++ b/lib/commands/apps-management.js
@@ -1,0 +1,94 @@
+import _ from 'lodash';
+import waitForCondition from 'asyncbox';
+import {fs, utils} from 'appium-support';
+import log from '../logger';
+
+let commands = {};
+const APP_STATE_NOT_INSTALLED = 0;
+const APP_STATE_NOT_RUNNING = 1;
+const APP_STATE_RUNNING_IN_FOREGROUND = 2;
+const APP_STATE_RUNNING_IN_BACKGROUND = 3;
+
+
+function extractMandatoryOptions (opts = {}, namesList) {
+  const result = {};
+  for (const name of namesList) {
+    if (!utils.hasValue(opts[name])) {
+      throw new Error(`'${name}' option is mandatory.`);
+    }
+    result[name] = opts[name];
+  }
+  return result;
+}
+
+commands.mobileInstallApp = async function (opts = {}) {
+  const {app} = extractMandatoryOptions(opts, ['app']);
+  const dstPath = await this.helpers.configureApp(app, '.apk');
+  log.info(`Installing '${dstPath}' to the device...`);
+  if (!await fs.exists(dstPath)) {
+    log.errorAndThrow(`The application at '${dstPath}' does not exist or is not accessible`);
+  }
+  let installArgs = [];
+  if (utils.hasValue(opts.installArgs)) {
+    installArgs = _.isArray(opts.installArgs) ? opts.installArgs : [opts.installArgs];
+  }
+  const installTimeout = _.isNumber(opts.installlTimeout) ? parseInt(opts.installlTimeout, 10) : 2 * 60 * 1000;
+  try {
+    await this.adbExec(['install', ...installArgs, dstPath], {timeout: installTimeout});
+    log.info(`Installation of '${dstPath}' succeeded`);
+  } finally {
+    if (dstPath !== app) {
+      await fs.rimraf(dstPath);
+    }
+  }
+};
+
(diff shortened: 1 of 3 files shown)


lib/commands/apps-management.js
```diff
@@ -0,0 +1,94 @@
...
+  for (const name of namesList) {
+    if (!utils.hasValue(opts[name])) {
+      throw new Error(`'${name}' option is mandatory.`);
+    }
+    result[name] = opts[name];
+  }
+  return result;
+}
+
+commands.mobileInstallApp = async function (opts = {}) {
```
jlipps: don't we already have logic for installing / uninstalling apps, either in the driver or in appium-adb? i wouldn't want to duplicate it unless necessary

lib/commands/apps-management.js
```diff
@@ -0,0 +1,94 @@
+import _ from 'lodash';
+import waitForCondition from 'asyncbox';
+import {fs, utils} from 'appium-support';
+import log from '../logger';
+
+let commands = {};
+const APP_STATE_NOT_INSTALLED = 0;
```
jlipps: are we making up these constants or do they map to something in android land?

lib/commands/execute.js
```diff
@@ -14,14 +14,19 @@ extensions.execute = async function (script, args) {
 
 extensions.executeMobile = async function (mobileCommand, opts = {}) {
   const mobileCommandsMapping = {
-    shell: async (x) => await this.mobileShell(x),
+    shell: 'mobileShell',
+    insstallApp: 'mobileInstallApp',
```
jlipps: `installApp`

lib/commands/apps-management.js
mykola-mokhnach: we have some parts of this logic here and there, but not everything in one place.

lib/commands/apps-management.js
mykola-mokhnach: I just wanted to make them similar the corresponding iOS constants.

lib/commands/apps-management.js
```diff
@@ -0,0 +1,94 @@
...
+const APP_STATE_NOT_RUNNING = 1;
+const APP_STATE_RUNNING_IN_FOREGROUND = 2;
+const APP_STATE_RUNNING_IN_BACKGROUND = 3;
+
+
+function extractMandatoryOptions (opts = {}, namesList) {
```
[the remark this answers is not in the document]
mykola-mokhnach: I don't find the method universal enough. We could move it into helpers though if it is needed in more than one module (which is not the case for now).

lib/commands/apps-management.js
jlipps: ok makes sense. my other question is, why make these "mobile" methods when appium endpoints already exist?

mykola-mokhnach: These endpoints work with the application under test only and won't allow one to handle any other package on the device including built-in packages

jlipps: why does this method take an 'app' option then, if it is only to be used to install the AUT, whose info we presumably already have?

mykola-mokhnach: it can be a path to any apk or an URL. Don't you like the argument name?

jlipps: I think I'm just confused about the point of this. It looked to me like you were adding a new command for clients, like:

```js
await driver.execute("mobile: installApp", {app: '/path/to/my.apk'});
```

in the context of an existing session, why would someone use this to install anything other than a separate app? if the point is to do a reinstall only, should we enforce same package as AUT? i'm just trying to make sure this is different from the existing endpoint.

mykola-mokhnach: One of possible scenarios - we need to deploy some helper app for the automation along with the main AUT. Reinstall and upgrade are also possible usage scenarios.

jlipps: why would someone not just use `driver.installApp` for those purposes as well?

mykola-mokhnach: because it affects the whole session

mykola-mokhnach:
@imurchie do you have any comments to this PR?

mykola-mokhnach:
@jlipps My intention was to make these commands more flexible by adding more options. But you are right - it would be more practical to extend existing endpoints. I will remove the redundant stuff from the PR

mykola-mokhnach:
I'll combine this with https://github.com/appium/appium-base-driver/pull/179

PR #302, opened 2018-01-11
Reset context after reset

neofreko:
**Problem: context is not reset on reset command**
After invoking `driver.reset` on webview context, appium will keep telling us that chrome is not reachable although the app is up on the foreground. Apparently, switchContext was not invoked since new context is the same as last context (before reset).
Current workaround is by invoking `closeApp` before calling `reset`.

**Solution:**
Similar to `closeApp`, after `reset` we need to reset curContext as well.

### DIFF
--- a/test/unit/commands/general-specs.js
+++ b/test/unit/commands/general-specs.js
@@ -310,6 +310,7 @@ describe('General', () => {
       helpers.installApkRemotely.calledWithExactly(driver.adb, driver.opts)
         .should.be.true;
       driver.grantPermissions.calledOnce.should.be.true;
+      driver.startAUT.calledOnce.should.be.true;
     });
     it('should do fast reset if fullReset is false', async () => {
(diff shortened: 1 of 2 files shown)


test/unit/commands/general-specs.js
```diff
@@ -181,7 +181,7 @@ describe('General', () => {
         .should.be.true;
     });
     it('should throw an error if APK does not exist', async () => {
-      await driver.installApp('non/existent/app.apk').should.be.rejectedWith(/does not exist or is not accessible/);
+      await driver.installApp('non/existent/app.apk').should.be.rejectedWith(/Could not find app apk at/);
```
imurchie: This change should be removed. It is broken. See https://travis-ci.org/appium/appium-android-driver/jobs/327567864#L6469

imurchie:
Thanks! Published in `appium-android-driver@1.37.1`.

PR #303, opened 2018-01-11
Improve screen recording flow

mykola-mokhnach:
I've changed the logic here, so now it's much simpler to get the result of the screen record. there is also a possibility to upload the file to a remote location using ftp/http protocols and protection from OOM issues. 

Related base driver PR: https://github.com/appium/appium-base-driver/pull/168

I'll also take care about finishing the java client implementation (the PR is there, but it has been waiting for almost a year).


imurchie:
I don't understand the need for this. The methods need to be added to the clients, to give access to this functionality, rather than clutter up the `mobile: ` space with duplicate functionality.

mykola-mokhnach:
> The methods need to be added to the clients

Where exactly they should be added?

jlipps:
I think what @imurchie is saying is that [/start_recording_screen](https://github.com/appium/appium-base-driver/blob/master/lib/mjsonwp/routes.js#L323) and [/stop_recording_screen](https://github.com/appium/appium-base-driver/blob/master/lib/mjsonwp/routes.js#L326) already exist, so we don't need to make `mobile:` versions of them. instead, we need to teach the appium clients (java, python, etc...) how to make calls to the existing endpoints.

mykola-mokhnach:
@jlipps I checked the current API endpoint in base driver and I don't find them very flexible and cross-platform. For example,  I'd like the stopRecording endpoint to return the taken video immediately. Also, some parameters, which are currently marked as required for start_recording are optional in reality.

Don't you mind if I change the endpoints in the base drivers to make them more flexible?

mykola-mokhnach:
Also, I could add such endpoints to XCTest driver, since iOS simulator also allow to record a video.

lib/commands/recordscreen.js
```diff
@@ -1,50 +1,209 @@
...
+function toReadableSizeString (bytes) {
+  if (bytes >= 1048576) {
+    return `${parseFloat(bytes / 1048576.0).toFixed(2)} MB`;
+  } else if (bytes >= 1024) {
+    return `${parseFloat(bytes / 1024.0).toFixed(2)} KB`;
+  }
+  return `${bytes} B`;
+}
+
+async function uploadRecordedMedia (adb, pathOnDevice, remotePath = null, uploadOptions = {}) {
```
mykola-mokhnach: all these helper functions can be also reused in iOS. Perhaps, we could move them to some common lib. the only question is to what lib

imurchie: The only place would be `appium-support`.

mayuayla:
Thank you so much for this code update.   Happy that there are amazing dedicated people improving Appium.   Looking forward to using this with the Python client.   
-Mayu

PR #305, opened 2018-01-12
Make sure we return a context

imurchie:
When closing apps and resetting, the context is set to null, and it remains so. If it is null when someone asks for the current context, that means it is NATIVE_APP.

Resolves https://github.com/appium/appium/issues/9933


imurchie:
Published in `appium-android-driver@1.37.2`.

PR #306, opened 2018-01-12
Stop making package-lock.json file

imurchie:
See appium/appium#9968


jlipps:
consider these all approved from my perspective

PR #307, opened 2018-01-13
Fix TypeError by Chrome driver detection

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/9978


imurchie:
Published in `appium-android-driver@1.37.3`.

PR #308, opened 2018-01-17
Add rect for window

KazuCocoa:
This implementation is for Android driver.
As same as : https://github.com/appium/appium-xcuitest-driver/pull/607

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -78,11 +78,13 @@ commands.getWindowSize = function () {
 
 // For W3C
 commands.getWindowRect = function () {
-  const rect = this.getWindowSize();
-  rect.x = 0;
-  rect.y = 0;
-
-  return rect;
+  const { width, height } = this.getWindowSize();
+  return {
+    width,
+    height,
+    x: 0,
+    y: 0
+  };
 };
 
 commands.getCurrentActivity = async function () {
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -78,7 +78,11 @@ commands.getWindowSize = function () {
 
 // For W3C
 commands.getWindowRect = function () {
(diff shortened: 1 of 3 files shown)


lib/commands/general.js
```diff
@@ -76,6 +76,15 @@ commands.getWindowSize = function () {
   return this.bootstrap.sendAction('getDeviceSize');
 };
 
+// For W3C
+commands.getWindowRect = function () {
```
mykola-mokhnach: is the endpoint present in NO_PROXY list?

imurchie: This probably does need to be added, or else it won't work in Chrome (unless Chromedriver handles this endpoint?).

mykola-mokhnach: then there should be isWebContext verification?

imurchie: Window size should be the native window size, no?

KazuCocoa: Yes, it should be window size on mobile devices so far.

test/unit/commands/general-specs.js
```diff
@@ -120,6 +120,13 @@ describe('General', function () {
       await driver.getWindowSize().should.be.equal('size');
     });
   });
+  describe('getWindowRect', function () {
+    it('should get window size', async () => {
+      sandbox.stub(driver.bootstrap, 'sendAction')
+        .withArgs('getDeviceSize').returns({width: 300, height: 400});
+      await driver.getWindowRect().toString().should.be.equal({width: 300, height: 400, x: 0, y: 0}.toString());
```
mykola-mokhnach: toString() will return something like `[object Object]`. You need to wrap objects into JSON.stringify to reach proper results

PR #310, opened 2018-01-18
Refactor file upload primitives used in screen recording

mykola-mokhnach:
Please do not merge until https://github.com/appium/appium-support/pull/61 is tagged


mykola-mokhnach:
@KazuCocoa Would you like to create wrappers for these in ruby/python client?

PR #311, opened 2018-01-18
Update lock logic, so it's in sync with iOS

mykola-mokhnach:
the endpoint in base driver requires lock to have the required `seconds` parameter. the previous implementation was ignoring it. Now this is fixed and the server will wait for given number of seconds before to unlock the screen if seconds > 0. The same behaviour is implemented for iOS.

FYI @vrunoa

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -121,16 +121,26 @@ commands.doDrag = async function (dragOpts) {
   }
 };
 
-commands.lock = async function () {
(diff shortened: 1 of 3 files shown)


imurchie:
Once the unit test is fixed, this can get in.

mykola-mokhnach:
Natürlich

mykola-mokhnach:
@imurchie tests passed. Can we merge it?

PR #313, opened 2018-01-19
Don't force `wipe-data`, if the `reboot` capability is provided.

vmaxim:
Data can be wiped from the user side via `avdAgs` capability.


imurchie:
Is this in response to something? Or just because it can be done in a different way?

vmaxim:
It is great then we can shutdown emulator after session was deleted. But with implicit `wipe-data` combo it's almost useless - for example when you need to modify system image (eg Chrome+Android O).

imurchie:
Ok.

PR #315, opened 2018-01-22
Reset context to NATIVE_APP instead of null

imurchie:
Rather than reset to `null`, which means native and causes much problems (see https://travis-ci.org/appium/appium-uiautomator2-driver/jobs/331143446 for instance), use `'NATIVE_APP'`.

Fixes the build for https://github.com/appium/appium-uiautomator2-driver/pull/134

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -209,7 +209,7 @@ commands.reset = async function () {
   // reset context since we don't know what kind on context we will end up after app launch.
   this.curContext = NATIVE_WIN;
 
-  return await this.startAUT();
+  return await this.isChromeSession ? this.startChromeSession() : this.startAUT();
 };
 
 commands.startAUT = async function () {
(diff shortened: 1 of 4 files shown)


lib/commands/general.js
```diff
@@ -239,7 +240,7 @@ commands.reset = async function () {
     await this.adb.stopAndClear(this.opts.appPackage);
   }
   // reset context since we don't know what kind on context we will end up after app launch.
-  this.curContext = null;
+  this.curContext = NATIVE_WIN;
```
mykola-mokhnach: what's the diff between NATIVE_WIN and NATIVE_APP?

imurchie: `NATIVE_WIN` is a constant with the value `'NATIVE_APP'` (see https://github.com/appium/appium-android-driver/blob/0a1727dea8d6a7e6d0acfd8c61f77882c9d57a85/lib/webview-helpers.js#L5).

vmaxim: This changes introduce new issue with the following flow `Start Session -> switch to WebView -> Reset -> switch to Native -> Find Element` : `Selenium error: chrome not reachable`

imurchie: Updated to recreate the chrome session using the current logic.

PR #316, opened 2018-01-23
if we are pushing a media file, broadcast the rescan intent

jlipps:
should help with pushing photos

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -13,7 +13,7 @@ const dragStepsPerSec = 40;
 const CONTAINER_PATH_MARKER = '@';
 // https://regex101.com/r/PLdB0G/2
 const CONTAINER_PATH_PATTERN = new RegExp(`^${CONTAINER_PATH_MARKER}([^/]+)/(.+)`);
-const ANDROID_PICTURE_PATH = '/mnt/sdcard/Pictures';
+const ANDROID_SDCARD_PATH_RE = new RegExp('^/mnt/sdcard/');
 const ANDROID_MEDIA_RESCAN_INTENT = 'android.intent.action.MEDIA_SCANNER_SCAN_FILE';
 
 let commands = {}, helpers = {}, extensions = {};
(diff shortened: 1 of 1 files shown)


lib/commands/actions.js
```diff
@@ -13,6 +13,8 @@ const dragStepsPerSec = 40;
 const CONTAINER_PATH_MARKER = '@';
 // https://regex101.com/r/PLdB0G/2
 const CONTAINER_PATH_PATTERN = new RegExp(`^${CONTAINER_PATH_MARKER}([^/]+)/(.+)`);
+const ANDROID_PICTURE_PATH = '/mnt/sdcard/Pictures';
```
mykola-mokhnach: is it the only path, that matches media pattern?

lib/commands/actions.js
```diff
@@ -229,6 +231,13 @@ commands.pushFile = async function (remotePath, base64Data) {
     } else {
       // adb push creates folders and overwrites existing files.
       await this.adb.push(localFile, remotePath);
+
+      // if we have a media file, broadcast media scan intent
+      if (remotePath.indexOf(ANDROID_PICTURE_PATH) === 0) {
```
mykola-mokhnach: there is `startsWith` method

jlipps:
@mykola-mokhnach i made the check a little more general since it seems different vendors can use different paths

lib/commands/actions.js
```diff
@@ -13,6 +13,8 @@ const dragStepsPerSec = 40;
 const CONTAINER_PATH_MARKER = '@';
 // https://regex101.com/r/PLdB0G/2
 const CONTAINER_PATH_PATTERN = new RegExp(`^${CONTAINER_PATH_MARKER}([^/]+)/(.+)`);
+const ANDROID_SDCARD_PATH_RE = new RegExp('^/mnt/sdcard/');
```
mykola-mokhnach: can it be simply /sdcard/ thou? The scanning based on target path anyway looks quite unstable to me? Can we, for example, analyse the type of the source file, for example using the system `file` utility (if present) or some node package to identify whether the particular file is a media file?

jlipps: sure, we could do that. or we could assume that if someone is pushing a file to the device it's most likely going to be a media file and just always do the broadcast. it probably doesn't hurt to do the broadcast every time anyway, since it's a relatively instant operation.

lib/commands/actions.js
```diff
@@ -229,6 +231,13 @@ commands.pushFile = async function (remotePath, base64Data) {
     } else {
       // adb push creates folders and overwrites existing files.
       await this.adb.push(localFile, remotePath);
+
+      // if we have a media file, broadcast media scan intent
+      if (ANDROID_SDCARD_PATH_RE.test(remotePath)) {
+        log.info("After pushing media file, broadcasting media scan intent");
+        await this.adb.shell(['am', 'broadcast', '-a',
```
mykola-mokhnach: should we ignore possible exceptions due to non-zero return code?

jlipps:
ok @mykola-mokhnach @imurchie i think this is now ready to go

jlipps:
published in 2.0.3

PR #319, opened 2018-01-29
Update and refactor application installation helpers

mykola-mokhnach:
The PR depends on https://github.com/appium/appium-adb/pull/290

PR #323, opened 2018-01-31
fix for issue #10086

Jiurong-Yang:
A simple fix for the timeout issue for appium-android-driver when installing app. You can reference the issue number for more information. Just trying to get one that passes the CI test.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -273,8 +273,8 @@ helpers.reinstallRemoteApk = async function (adb, localApkPath, pkg,
 
 helpers.installApkRemotely = async function (adb, opts) {
   let {app, appPackage, fastReset, androidInstallTimeout} = opts;
-  if (androidInstallTimeout ==null){
-	androidInstallTimeout = REMOTE_INSTALL_TIMEOUT
+  if (androidInstallTimeout ==null) {
+    androidInstallTimeout = REMOTE_INSTALL_TIMEOUT;
   }
   if (!app || !appPackage) {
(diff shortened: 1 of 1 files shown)


jlipps:
@Jiurong-Yang the restriction is there precisely to _keep_ Appium open source, to prevent anyone's employer from claiming that code they submitted to Appium is actually copyrighted. It's a pain, but it's good for the project.

lib/android-helpers.js
```diff
@@ -273,7 +273,9 @@ helpers.reinstallRemoteApk = async function (adb, localApkPath, pkg,
 
 helpers.installApkRemotely = async function (adb, opts) {
   let {app, appPackage, fastReset, androidInstallTimeout} = opts;
-
+  if (androidInstallTimeout ==null) {
```
jlipps: it's better to say `_.isNull(androidInstallTimeout)`, but that may not do exactly what we want. Probably more like if it's null or undefined.

Jiurong-Yang: Can you elaborate on this? Does .isNull and == null have different behavior? To be honestly I've never written a line of Java in my life so I'm not very familiar to all the syntax and language specific nuances. So I'm always open to suggestions

mykola-mokhnach:
I'd rather close it in favour of https://github.com/appium/appium-android-driver/pull/319, since this part of code is completely reworked there

Jiurong-Yang:
Closing because another contributor is already working on it.

PR #324, opened 2018-01-31
Fix application management endpoints

mykola-mokhnach:
Also added some missing async/await pairs. For good.


mykola-mokhnach:
@imurchie Please do not forget to raise the major component version for the upcoming tag

PR #330, opened 2018-02-07
Fix handling of chrome

imurchie:
Move chromedriver setup functions out of `lib/driver.js` so they are available in uiautomator2 driver. Fixes https://github.com/appium/appium/issues/10146

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -184,9 +184,9 @@ helpers.isChromedriverContext = function (viewName) {
 };
 
 helpers.shouldDismissChromeWelcome = function shouldDismissChromeWelcome () {
-  return !_.isUndefined(this.opts.chromeOptions) &&
-    _.isArray(this.opts.chromeOptions.args) &&
-    this.opts.chromeOptions.args.indexOf('--no-first-run') !== -1;
+  return !!this.opts.chromeOptions &&
+         _.isArray(this.opts.chromeOptions.args) &&
+         this.opts.chromeOptions.args.includes('--no-first-run');
 };
 
 helpers.dismissChromeWelcome = async function dismissChromeWelcome () {
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import logger from '../logger';
+import log from '../logger';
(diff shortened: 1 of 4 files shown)


lib/commands/context.js
```diff
@@ -183,6 +183,66 @@ helpers.isChromedriverContext = function (viewName) {
   return _.includes(viewName, WEBVIEW_WIN) || viewName === CHROMIUM_WIN;
 };
 
+helpers.shouldDismissChromeWelcome = function shouldDismissChromeWelcome () {
+  return !_.isUndefined(this.opts.chromeOptions) &&
```
mykola-mokhnach: shall we check if it is a plain object instead?

lib/commands/context.js
```diff
@@ -183,6 +183,66 @@ helpers.isChromedriverContext = function (viewName) {
...
+    _.isArray(this.opts.chromeOptions.args) &&
+    this.opts.chromeOptions.args.indexOf('--no-first-run') !== -1;
```
mykola-mokhnach: includes?

lib/commands/find.js
```diff
@@ -19,9 +19,6 @@ helpers.doFindElementOrEls = async function (params) {
 // mult: multiple elements or just one?
 // context: finding an element from the root context? or starting from another element
 helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
-  // throws error if not valid, uses this.locatorStrategies
```
mykola-mokhnach: why this is removed?

lib/commands/context.js
```diff
@@ -183,6 +183,66 @@ helpers.isChromedriverContext = function (viewName) {
...
+    // DO NOTHING, THIS DEVICE DIDNT LAUNCH THE SIGNIN DIALOG
+    // IT MUST BE A NON GMS DEVICE
+    log.warn(`This device didnt show Chrome SignIn dialog, ${e.message}`);
+  }
+};
+
+helpers.startChromeSession = async function startChromeSession () {
+  log.info("Starting a chrome-based browser session");
+  let opts = _.cloneDeep(this.opts);
+  opts.chromeUseRunningApp = false;
```
mykola-mokhnach: where this is used?

PR #331, opened 2018-02-08
Move to real find methods, not findByElOrEls

imurchie:
Unit tests should (as they do) test `findElOrEls`, but the rest of the time we ought to use the base driver methods that are actually called by the protocol.

This also relieves us of the need of validating the locators, which is done already (https://github.com/appium/appium-base-driver/blob/master/lib/basedriver/commands/find.js#L19), and leads to weird logs with multiple validation lines.

### DIFF
--- a/test/functional/commands/find/from-el-e2e-specs.js
+++ b/test/functional/commands/find/from-el-e2e-specs.js
@@ -12,31 +12,30 @@ const alv = 'android.widget.ListView';
 
 describe('Find - from element', function () {
   let driver;
+  let el;
   before(async function () {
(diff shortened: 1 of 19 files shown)


test/functional/commands/find/from-el-e2e-specs.js
```diff
@@ -12,31 +12,30 @@ const alv = 'android.widget.ListView';
 
 describe('Find - from element', function () {
   let driver;
+  let el;
```
mykola-mokhnach: I rather rename it to parentElement

PR #332, opened 2018-02-12
Fix the returned value in prepareAVDArgs helper

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10177

PR #333, opened 2018-02-13
Do not include empty avd args

imurchie:
If there are no `avdArgs` then don't include them.

Fixes https://github.com/appium/appium/issues/10177


mykola-mokhnach:
Have you seen https://github.com/appium/appium-android-driver/pull/332 ?

mykola-mokhnach:
However this solution looks more elegant %)

imurchie:
Hmm. No. I looked for an alert that a PR had come in about it. I guess Github was slow or something.

PR #335, opened 2018-02-19
Fix background when using appWaitPkg/Activity

Kignuf:
When app is started using an activity but redirects to another one (or
another package) one may use appWaitPackage or appWaitActivity either
as capabilities which will be used during the initial launch, or as parameters
of the startActivity command.

Using the background command in such circumstances would lead to ADB
permission denial error (activity is not exported).

This PR fixes that issue when:
* the activity is the original session activity (launchApp command)
* the app was started using startActivity command

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -124,11 +124,12 @@ commands.background = async function (seconds) {
   if (this.opts.startActivityArgs && this.opts.startActivityArgs[`${appPackage}/${appActivity}`]) {
     // the activity was started with `startActivity`, so use those args to restart
     args = this.opts.startActivityArgs[`${appPackage}/${appActivity}`];
-  } else if (appPackage === this.opts.appPackage && appActivity === this.opts.appActivity) {
+  } else if (appPackage === this.opts.appPackage && appActivity === this.opts.appActivity ||
+    appPackage === this.opts.appWaitPackage && this.opts.appWaitActivity.split(',').indexOf(appActivity) >= 0) {
     // the activity is the original session activity, so use the original args
     args = {
-      pkg: appPackage,
-      activity: appActivity,
+      pkg: this.opts.appPackage,
+      activity: this.opts.appActivity,
       action: this.opts.intentAction,
       category: this.opts.intentCategory,
       flags: this.opts.intentFlags,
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -124,11 +124,12 @@ commands.background = async function (seconds) {
   if (this.opts.startActivityArgs && this.opts.startActivityArgs[`${appPackage}/${appActivity}`]) {
     // the activity was started with `startActivity`, so use those args to restart
     args = this.opts.startActivityArgs[`${appPackage}/${appActivity}`];
-  } else if (appPackage === this.opts.appPackage && appActivity === this.opts.appActivity) {
+  } else if (appPackage === this.opts.appPackage && appActivity === this.opts.appActivity ||
```
jlipps: i think it would be good to add parentheses for grouping to make it visually clear the boolean operator precedence

lib/commands/general.js
```diff
@@ -124,11 +124,12 @@ commands.background = async function (seconds) {
...
+    appPackage === this.opts.appWaitPackage && this.opts.appWaitActivity.split(',').indexOf(appActivity) >= 0) {
```
mykola-mokhnach: this might fail if  this.opts.appWaitActivity is undefined

Kignuf:
Thanks a lot for the reviews !

I've updated the code to add parenthesis and make sure we don't crash if appWaitActivity is undefined.
Let me know if that's ok for you.

lib/commands/general.js
```diff
@@ -124,8 +124,10 @@ commands.background = async function (seconds) {
   if (this.opts.startActivityArgs && this.opts.startActivityArgs[`${appPackage}/${appActivity}`]) {
     // the activity was started with `startActivity`, so use those args to restart
     args = this.opts.startActivityArgs[`${appPackage}/${appActivity}`];
-  } else if (appPackage === this.opts.appPackage && appActivity === this.opts.appActivity ||
-    appPackage === this.opts.appWaitPackage && this.opts.appWaitActivity.split(',').indexOf(appActivity) >= 0) {
+  } else if ((appPackage === this.opts.appPackage && appActivity === this.opts.appActivity) ||
+    (appPackage === this.opts.appWaitPackage &&
+      this.opts.appWaitActivity &&
```
mykola-mokhnach: you could use `(this.opts.appWaitActivity || '').split(',').includes(appActivity)` instead

Kignuf:
Thanks again for your advices, should I squash the last 2 commits ?

mykola-mokhnach:
nah, lets wait for CI resulsts

PR #336, opened 2018-02-21
Added new Capability

vinodbenure:
Hi

I am trying to add new capability(user profile) for Android. This capability is related to the appium [issue](https://github.com/appium/appium/issues/8305). However I need help in below issues:

1. The idea is to make this capability as optional.

2. I have added the new capability 'userProfile'  in "**desired-caps.js**" of _appium-android-driver_ package.
        `userProfile: {
    isNumber: true
  },`

3. If user specifies this capability then this capability value should be added to defaultArgs of executable so that each and every adb command goes with this parameter.

    For eg. `executable.defaultArgs.push("--user", opts.userProfile);`

4. However, I want to make sure that this capability is added only if the the user profile available on the device. We can get the list of available users by using below adb command.
            `adb shell pm list users`

    Once we get the list of users and we can parse the values from output of the above command.

5. If the user profile is not available on the device then we can initialize the device with the traditional way without using this capability.

6. I am little confused about where add the this piece of code to. Whether to add this to "**driver.js**", "**android-helpers.js**" of _appium-android-driver_ package or  add this to "**system-calls.js**" of _appium-adb_ package .

Waiting for suggestions

### DIFF
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -154,6 +154,9 @@ let commonCapConstraints = {
   unlockKey: {
     isString: true
   },
+  userProfile: {
+    isNumber: true
+  },
 };
 
 let uiautomatorCapConstraints = {


mykola-mokhnach:
low-level adb interactions should be placed into appium-adb module

vinodbenure:
@mykola-mokhnach 
I can add the below code **system-calls.js** of _appium-adb_ package.

> systemCallMethods.getAvailableUserProfiles = async function () {
  log.debug("Getting list of user profiles...");
  try {
    let {stdout} = await exec(this.executable.path, [' shell pm list users']);
    // expecting adb pm list users to return output as
    //Users:
    //UserInfo{0:Owner:13} running
    let startingIndex = stdout.indexOf("UserInfo");
    if (startingIndex === -1) {
      throw new Error(`Unexpected output while trying to list of User Profile. output was: ${stdout}`);
    } else {
      // slicing ouput we care about.
      stdout = stdout.slice(startingIndex);
      let userProfiles = [];
      for (let line of stdout.split("\n")) {
        if (line.trim() !== "" &&
            line.indexOf("Users") === -1 &&
            line.indexOf("no devices") === -1) {
          let fullUserInfo = line.split(":");
          let lineInfo = fullUserInfo[0].split("{");
          // value is either integer value of user profile or "no devices/emulators found"
          userProfiles.push({userProfile: lineInfo[1], userValue: lineInfo[1]});
        }
      }
      log.debug(`${userProfiles.length} User Profile(s) available`);
      return userProfiles;
    }
  } catch (e) {
    log.errorAndThrow(`Error while getting list of User Profiles. Original error: ${e.message}`);
  }
};
>

However, I want to call this method if the 'userProfile' capability value is specified in the desired capabilities. Do you have any suggestion to where to place the below condition and call the above method?
`if (opts.userProfile) {      
 // call the above method and check whether opts.userProfile is available,
 if available execute below code
      this.executable.defaultArgs.push("-u", opts.userProfile);
    }`

imurchie:
Thanks for this!

I would suggest that the user profile number be passed in when creating the instance of `appium-adb`. This would mean modifying `ADB.createADB()` (https://github.com/appium/appium-adb/blob/master/lib/adb.js#L63-L67), and adding a method to do the check if the option is there (since constructors cannot call `async` code), on the model of the call to `await adb.getAdbWithCorrectAdbPath();`.

Something like:
```js
// lib/adb.js
class ADB {
  // existing code
  async setUserProfile (userProfile) {
    const profiles = await this.getAvailableUserProfiles();
    if (!profiles.includes(userProfile)) {
      log.warn(`User profile '${userProfile}' not available on device.`);
      return;
    }
    this.executable.defaultArgs.push('-u', userProfile);
  }
}

ADB.createADB = async function (opts) {
  let adb = new ADB(opts);
  await adb.getAdbWithCorrectAdbPath();
  if (util.hasValue(opts.userProfile)) {
    await adb.setUserProfile(opts.userProfile);
  }
  return adb;
};
```

Then the calls to `createADB` in `appium-android-driver` and `appium-uiautomator2-driver` would need to be adjusted when the cap is set.

vinodbenure:
@imurchie 
Thanks for helping.  After making the changes as you suggested, On my local machine I am getting below issue.

`[debug] [ADB] Trying to find a connected android device
[debug] [ADB] Getting connected devices...
[ADB] Error: Error while getting connected devices. Original error: Command 'C\:\\AndroidSDK\\platform-tools\\adb.exe -P 5037 --user 0 devices' exited with code 1
    at Object.wrappedLogger.errorAndThrow (../../lib/logging.js:69:13)
    at ADB.callee$0$0$ (..\..\../lib/tools/system-calls.js:170:9)`

As you can clearly see in the log that the new capability "**--user**" is getting appended with adb path and hence it becomes an invalid command when the adb path is combined with the "**devices**" option.

I want to pass the new capability "--user" as an argument to adb. However after making the changes, its getting appended to the adb default path. Any idea how to get this working? And not append the path to the default adb path?

imurchie:
I don't understand. `--user 0` is getting added to the default arguments:
```
C\:\\AndroidSDK\\platform-tools\\adb.exe -P 5037 --user 0 devices
```
It appears the `--user` is not a valid argument for _all_ `adb` commands, so this won't work. At the moment there is no distinction between what arguments get sent to what `adb` commands... all the arguments get sent to all the commands.

So to move this forward, you need to figure out what commands `--user 0` will work as an argument for. Is `adb devices` the only one that fails?

Once that is determined, I am happy to help figure out a good way to distinguish the commands in the `appium-adb` package, so the correct arguments get sent to the correct commands.

vinodbenure:
@imurchie 
Thanks for guiding me with this.

As you suggested. I have checked the commands and arguments for adb. `--user 0` will not work with all the  adb commands. This argument will work only at profile level and not at device level. For e.g. device serial number/udid is used for all the commands which is valid and similarly we can not use the `--user`.

I have figured out where we can use the `--user` argument. Lets consider the android device has two profiles(_Most of the devices will have only one profile by default_). This multiple profiles feature is provided by Google. And the numeric value for the default user profile is always **0**.

Lets assume that there are two profiles with values **0** and **1**. The AUT is installed on the both profiles on the device. If we don't use `--user` argument, all the commands sent through adb will point to profile **0**.

By default appium will use the below command if we don't specify the `--user` argument
`adb -s deviceSerialNumner shell am start -n com.package.name/.ActivityName`

And this will launch the AUT from the user profile **0**. 

But what if we want to launch the AUT from user profile **1** ?. We can add the `--user` argument to the above command and launch the app from profile **1**.

`adb -s deviceSerialNumner shell am start --user 1 -n com.package.name/.ActivityName`


I think adding the user profile argument only for launch command is sufficient. Because once the AUT is on foreground, UIAutomator can identify the elements on the app irrespective of the user profile.

You can guide me where need to make changes to get this working. Thanks

vinodbenure:
@imurchie  any update on this?

vinodbenure:
@imurchie  I am waiting for your suggestion. This feature is very essential for us to run our scripts in second profile. Can you please guide?

mykola-mokhnach:
https://github.com/appium/appium-adb/pull/318

no-need-to-dlmbf0

PR #338, opened 2018-02-22
Add 'bugreport' and 'server' log types

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/4446 and https://github.com/appium/appium/issues/5032

PR #339, opened 2018-02-23
Use base driver logging methods

imurchie:
Use logging commands from https://github.com/appium/appium-base-driver/pull/191. Re-implements functionality from https://github.com/appium/appium-android-driver/pull/338

### DIFF
--- a/test/unit/commands/log-specs.js
+++ b/test/unit/commands/log-specs.js
@@ -1,0 +1,38 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import _ from 'lodash';
+import sinon from 'sinon';
+import ADB from 'appium-adb';
+import os from 'os';
+import AndroidDriver from '../../..';
+
+
+chai.should();
+chai.use(chaiAsPromised);
+
+describe('commands - logging', function () {
+  let driver;
+  before(function () {
+    driver = new AndroidDriver();
+    driver.adb = new ADB();
+  });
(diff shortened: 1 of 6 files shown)


test/unit/commands/log-specs.js
```diff
@@ -0,0 +1,38 @@
...
+  let driver;
+  before(function () {
+    driver = new AndroidDriver();
+    driver.adb = new ADB();
+  });
+  describe('getLogTypes', function () {
+    it('should get log types', async function () {
+      const types = await driver.getLogTypes();
+      // all the types should be returned
+      _.xor(['logcat', 'bugreport', 'server'], types).should.eql([]);
```
mykola-mokhnach: should we so this verification in the same way it is done for ios driver?

imurchie: If we wanted them to be the same, I would go for this way (since it is order independent).

mykola-mokhnach: we could also import log type names instead of hardcoding them here

imurchie: I personally would rather they be hardcoded here. Otherwise it doesn't really test anything other than the same data is passed around. And it wouldn't catch an erroneous edit of the type, for instance. When a new type is added it is easy enough to add here.

That all being said, this doesn't need to be an e2e test, and that would shorten the feedback cycle, since the unit tests should be run on commit.

PR #341, opened 2018-03-15
Tune log records format to make clients happy

mykola-mokhnach:
Fixes https://github.com/appium/appium/issues/10355#issuecomment-373435982


jlipps:
yeah, the java client will barf if logs aren't the standard format. LGTM

PR #342, opened 2018-03-16
Tune log entry format for bugreport output

mykola-mokhnach:
Selenium client does not like log items, which are not compatible to LogRecord class


imurchie:
This looks ok but the test needs fixing.

mykola-mokhnach:
fixed the unit test

imurchie:
Published in `appium-android-driver@2.1.3`.

PR #343, opened 2018-03-19
Add a possibility to broadcast logcat output via websocket

mykola-mokhnach:
Based on https://github.com/appium/appium-adb/pull/306

### DIFF
--- a/lib/commands/log.js
+++ b/lib/commands/log.js
@@ -1,9 +1,12 @@
 import log from '../logger';
 import os from 'os';
 import _ from 'lodash';
-
+import WebSocket from 'ws';
+import { DEFAULT_WS_PATHNAME_PREFIX } from 'appium-base-driver';
 
 let commands = {}, helpers = {}, extensions = {};
+
+const WEBSOCKET_ENDPOINT = (sessionId) => `${DEFAULT_WS_PATHNAME_PREFIX}/session/${sessionId}/appium/device/logcat`;
 
 // https://github.com/SeleniumHQ/selenium/blob/0d425676b3c9df261dd641917f867d4d5ce7774d/java/client/src/org/openqa/selenium/logging/LogEntry.java
 function toLogRecord (timestamp, level, message) {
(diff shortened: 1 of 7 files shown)


jlipps:
I'm all for websocket output in general. Here are some things to think about:

1. websockets are not part of the webdriver spec, so anything we do via websockets i'd like to also make sure to provide via the spec
2. are we going to take on the burden of adding a websocket client to the appium bindings?
3. let's make sure we architect a single point of entry for any websocket connections. we may add more things to websockets in the future. in fact we have had websocket log output in the past (maybe it even still exists?). i think having a single websocket server mediated by appium itself, with differently-tagged messages, is going to be more maintainable long-term, rather than adding servers in separate drivers.

jlipps:
code LGTM

mykola-mokhnach:
@jlipps It is more like POC for now, so we can try how this works and whether it is stable. Also, I'm totally open to any ideas you have on how this implementation can be optimised better, because there are no specs at all and we are kind'a pioneering the area. 
Regarding the client-side implementation I wouldn't case too much for now, since these two methods I added are anyway available via `execute: mobile` interface and implementing a simple websocket client itself is one-line solution for almost any language.

lib/commands/log.js
```diff
@@ -45,6 +51,83 @@ extensions.supportedLogTypes = {
   },
 };
 
+function assignUpgradeEventListener () {
+  if (IS_UPGRADE_LISTENER_ASSIGNED && EXPRESS_SERVER.listenerCount('upgrade') !== 0) {
+    return;
+  }
+
+  // https://github.com/websockets/ws/pull/885
+  EXPRESS_SERVER.on('upgrade', (request, socket, head) => {
```
mykola-mokhnach: I'm thinking about moving this part into base driver. Also it might be useful to move all websocket-related endpoints into some common root, for example `/wd/hub/ws/:sessionId:/logcat`, where `/wd/hub/ws/` will be the common prefix for all websockets. WDYT @jlipps ?

jlipps: i agree with this. if we're going to add WS stuff, let's do it in one place and provide a nice set of convenience methods for drivers to use to hook into it.

lib/server.js
```diff
@@ -2,12 +2,12 @@ import log from './logger';
 import { server as baseServer, routeConfiguringFunction } from 'appium-base-driver';
 import AndroidDriver from './driver';
 
-
 async function startServer (port, host) {
   let d = new AndroidDriver();
   let router = routeConfiguringFunction(d);
   let server = baseServer(router, port, host);
   log.info(`AndroidDriver server listening on http://${host}:${port}`);
+  d.server = server;
```
imurchie: This will work for the running the `appium-android-driver` tests, but this server does not exist for Appium itself, where the server is created here: https://github.com/appium/appium/blob/master/lib/main.js#L124

I'm not sure how this is intended to work.

mykola-mokhnach: :( what would be the best way to get this server instance in the driver then?

mykola-mokhnach: https://github.com/appium/appium/pull/10400

mykola-mokhnach:
@KazuCocoa Would you like to create Ruby wrapper around this feature?

KazuCocoa:
🆗
I'll do in ruby_lib like xcuitest's mobile commands.

PR #344, opened 2018-03-21
Tune pushStrings helper

mykola-mokhnach:
This improves the flow of `pushStrings` helper and performs temporary resources cleanup. also it will avoid errors like https://gist.github.com/tfcardoso/355aede1f8bab333ab76e5b0a341cd79#file-cpuinfo-L177 in the log


dpgraham:
Published to 2.1.4

PR #345, opened 2018-03-22
Make it possible to record screen on Android P emulators

mykola-mokhnach:
It looks like Google has finally fixed compatibility issues since Android P and from now on one can record screen on emulator the same way he does it on real devices.

### DIFF
--- a/lib/commands/recordscreen.js
+++ b/lib/commands/recordscreen.js
@@ -16,6 +16,8 @@ const DEFAULT_RECORDING_TIME_SEC = MAX_RECORDING_TIME_SEC;
 const PROCESS_SHUTDOWN_TIMEOUT_SEC = 5;
 const SCREENRECORD_BINARY = 'screenrecord';
 const DEFAULT_EXT = '.mp4';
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
@KazuCocoa FYI

lib/commands/recordscreen.js
```diff
@@ -263,11 +271,18 @@ commands.startRecordingScreen = async function (options = {}) {
...
  *                   parameter is empty or null or an empty string.
  * @throws {Error} If there was an error while getting the name of a media file
- *                 or the file content cannot be uploaded to the remote location.
+ *                 or the file content cannot be uploaded to the remote location
+ *                 or screen recording is not supported on the device under test.
  */
 commands.stopRecordingScreen = async function (options = {}) {
   const {remotePath, user, pass, method} = options;
 
+  const apiLevel = await this.adb.getApiLevel();
```
imurchie: This validity check could be extracted out, since it is repeated twice (and in the future might change in both places, as different versions of things support different things).

mykola-mokhnach: fixed

PR #346, opened 2018-03-26
Replace `kill -0` with `ps`

mykola-mokhnach:
It looks like this way of terminated process detection is more precise on Android.

### DIFF
--- a/lib/commands/recordscreen.js
+++ b/lib/commands/recordscreen.js
@@ -42,14 +42,23 @@ async function finishScreenCapture (adb, pids) {
     return true;
   }
   try {
+    // Wait until the process is terminated
     await waitForCondition(async () => {
       try {
-        await adb.shell(['kill', '-0', ...pids]);
-      } catch (ign) {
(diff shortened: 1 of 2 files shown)


lib/commands/recordscreen.js
```diff
@@ -42,14 +42,23 @@ async function finishScreenCapture (adb, pids) {
     return true;
   }
   try {
+    // Wait until the process is terminated
     await waitForCondition(async () => {
       try {
-        await adb.shell(['kill', '-0', ...pids]);
-      } catch (ign) {
+        const output = await adb.shell(['ps']);
+        for (const line of output.split('\n')) {
```
imurchie: No need to split, if you use a multiline regexp, like (for pid 8479) [/\b8479\b[^\n]+\bscreenrecord$/m](https://regex101.com/r/0AlSOn/1)

mykola-mokhnach: Thanks, changed

PR #350, opened 2018-04-11
Do not try to change connection state if it is already in correct state

mykola-mokhnach:
The previous approach was not really effective if it was only necessary to change the state of the particular system, for example WiFi only without touching the others via setConnection API.

### DIFF
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -40,11 +40,11 @@ commands.setNetworkConnection = async function (type) {
   const shouldEnableDataConnection = (type & DATA_MASK) !== 0;
 
   const currentState = await this.getNetworkConnection();
-  const isArirplaneModeEnabled = (currentState & AIRPLANE_MODE_MASK) !== 0;
+  const isAirplaneModeEnabled = (currentState & AIRPLANE_MODE_MASK) !== 0;
   const isWiFiEnabled = (currentState & WIFI_MASK) !== 0;
   const isDataEnabled = (currentState & DATA_MASK) !== 0;
 
-  if (shouldEnableAirplaneMode !== isArirplaneModeEnabled) {
+  if (shouldEnableAirplaneMode !== isAirplaneModeEnabled) {
     await this.wrapBootstrapDisconnect(async () => {
       await this.adb.setAirplaneMode(shouldEnableAirplaneMode);
     });
@@ -55,8 +55,8 @@ commands.setNetworkConnection = async function (type) {
     log.info('Not changing airplane mode, since it already has the correct state');
   }
 
-  if (shouldEnableAirplaneMode) {
-    log.info('Not changing data connection and Wi-Fi states, because airplane mode is enabled');
+  if (shouldEnableAirplaneMode && shouldEnableWifi === isWiFiEnabled) {
+    log.info('Not changing data connection, because airplane mode is enabled');
     return;
   }
   if (shouldEnableWifi === isWiFiEnabled && shouldEnableDataConnection === isDataEnabled) {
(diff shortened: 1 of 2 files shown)


lib/commands/network.js
```diff
@@ -31,25 +35,48 @@ commands.isWifiOn = async function () {
...
-      await this.setWifiState(wifi);
-      await this.adb.setDataState(data, this.isEmulator());
+      await this.adb.broadcastAirplaneMode(shouldEnableAirplaneMode);
     });
+  } else {
+    log.info('Not changing airplane mode, since it already has the correct state');
   }
 
+  if (shouldEnableAirplaneMode) {
+    log.info('Not changing data connection and Wi-Fi states, because airplane mode is enabled');
```
vmaxim: Wi-Fi can be enabled in airplane mode on real devices.

mykola-mokhnach: I'd consider such possibility as android bug then :)

vmaxim: It's not a bug it's a feature (:

lib/commands/network.js
```diff
@@ -31,25 +35,48 @@ commands.isWifiOn = async function () {
...
 
-  await this.wrapBootstrapDisconnect(async () => {
-    await this.adb.setAirplaneMode(airplaneMode);
-  });
-  await this.wrapBootstrapDisconnect(async () => {
-    await this.adb.broadcastAirplaneMode(airplaneMode);
-  });
-  if (!airplaneMode) {
+  const currentState = await this.getNetworkConnection();
+  const isArirplaneModeEnabled = (currentState & AIRPLANE_MODE_MASK) !== 0;
```
SrinivasanTarget: `isAirplaneModeEnabled`

lib/commands/network.js
mykola-mokhnach: Ok, I'll change that condition. I don't fly much though :)

lib/commands/network.js
mykola-mokhnach: corrected

lib/commands/network.js
```diff
@@ -31,25 +35,51 @@ commands.isWifiOn = async function () {
...
+  }
+
+  await this.wrapBootstrapDisconnect(async () => {
+    if (shouldEnableWifi !== isWiFiEnabled) {
+      await this.setWifiState(shouldEnableWifi);
+    } else {
+      log.info(`Not changing Wi-Fi state, since it is already ` +
+               `${shouldEnableWifi ? 'enabled' : 'disabled'}`);
+    }
+    if (shouldEnableDataConnection !== isDataEnabled) {
```
vmaxim: There is no point to toggle Data connection in Airplane mode.

mykola-mokhnach: this is handled in the above condition:

```
if (shouldEnableAirplaneMode && shouldEnableWifi === isWiFiEnabled) {
```

vmaxim: shouldEnableAirplaneMode = true
shouldEnableWifi = true
shouldEnableDataConnection = true
isWiFiEnabled = false
isDataEnabled = false

mykola-mokhnach: ok I got it

lib/commands/network.js
```diff
@@ -31,25 +35,50 @@ commands.isWifiOn = async function () {
...
-      await this.setWifiState(wifi);
-      await this.adb.setDataState(data, this.isEmulator());
+      await this.adb.broadcastAirplaneMode(shouldEnableAirplaneMode);
     });
+  } else {
+    log.info(`Not changing airplane mode, since it is already ` +
+             `${shouldEnableAirplaneMode ? 'enabled' : 'disabled'}`);
   }
 
+  if (shouldEnableWifi === isWiFiEnabled && shouldEnableDataConnection === isDataEnabled) {
```
vmaxim: This ```if``` statement is redundant.

PR #351, opened 2018-04-20
Check other apps versions before reinstalling them

mykola-mokhnach:
Usually third party apps only need to be reinstalled if there are new versions of them, which speeds up session initialization.


dpgraham:
Lgtm

dpgraham:
Good timing on this. I'll publish this and get it into 1.8.0

PR #352, opened 2018-04-21
Return the current connection state instead of undefined

mykola-mokhnach:
this method is expected to return the resulting connection state rather than undefined.

### DIFF
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -58,7 +58,7 @@ commands.setNetworkConnection = async function (type) {
 
   if (shouldEnableWifi === isWiFiEnabled && shouldEnableDataConnection === isDataEnabled) {
     log.info('Not changing data connection/Wi-Fi states, since they are already set to expected values');
-    return;
+    return await this.getNetworkConnection();
   }
 
   await this.wrapBootstrapDisconnect(async () => {


lib/commands/network.js
```diff
@@ -58,7 +58,7 @@ commands.setNetworkConnection = async function (type) {
 
   if (shouldEnableWifi === isWiFiEnabled && shouldEnableDataConnection === isDataEnabled) {
     log.info('Not changing data connection/Wi-Fi states, since they are already set to expected values');
-    return;
+    return await this.getNetworkConnection();
```
jlipps: is there a reason we do the call to `getNetworkConnection` again? shouldn't we theoretically know what the state is now, without having to query it again?

mykola-mokhnach: yep, this might be simplified

mykola-mokhnach: I've checked it again and I think it would be better to keep it as it. Theoretically we might return just the expected connection state here and also at the end of the method, but we'd prefer to know if the connection state has really been changed. Also, the `getNetworkConnection` call itself is cheap

mykola-mokhnach: ok, I've gotten some better solution.

PR #355, opened 2018-04-24
Fix tap by element coordinates

mykola-mokhnach:
The same fix as for https://github.com/appium/appium-uiautomator2-driver/pull/163


imurchie:
I'm confused. Why is the ability to set an element and the offset being removed?

mykola-mokhnach:
Because such feature is not supported by the bootstrap server itself:

```java
  @Override
  public AndroidCommandResult execute(final AndroidCommand command)
      throws JSONException {
    if (command.isElementCommand()) {
      try {
        final AndroidElement el = command.getElement();
        el.click();
        return getSuccessResult(true);
      } catch (final UiObjectNotFoundException e) {
        return new AndroidCommandResult(WDStatus.NO_SUCH_ELEMENT,
            e.getMessage());
      } catch (final Exception e) { // handle NullPointerException
        return getErrorResult("Unknown error");
      }
    } else {
      final Hashtable<String, Object> params = command.params();
      Point coords = new Point(Double.parseDouble(params.get("x").toString()),
          Double.parseDouble(params.get("y").toString()) );

      try {
        coords = PositionHelper.getDeviceAbsPos(coords);
      } catch (final UiObjectNotFoundException e) {
        return new AndroidCommandResult(WDStatus.NO_SUCH_ELEMENT,
            e.getMessage());
      } catch (final InvalidCoordinatesException e) {
        return new AndroidCommandResult(WDStatus.INVALID_ELEMENT_COORDINATES,
            e.getMessage());
      }

      final boolean res = UiDevice.getInstance().click(coords.x.intValue(), coords.y.intValue());
      return getSuccessResult(res);
    }
  }
```

As you can see here it can only perform click on element or click by absolute coordinates.

imurchie:
Ok. That makes sense, then.

imurchie:
It's still possible to do what we purport to do (element with offset) by converting to absolute coordinates (i.e., finding the coordinates of the element and then adding the offset). Since this is what is purported to be done.

mykola-mokhnach:
@imurchie I think we could merge this solution for now, since it is not ideal, but fixes the original problem, so clients could properly run their tests with Appium@beta. In the meantime one could refactor the code and present something better. Same for https://github.com/appium/appium-uiautomator2-driver/pull/163

mykola-mokhnach:
@imurchie the code above is already applied in https://github.com/appium/appium-android-driver/blob/20c9e4a64626897eb74d47852f62fbccf7cba663/lib/commands/touch.js#L196 method.

PR #356, opened 2018-04-26
Add a possibility to include stderr output into adb:shell call

mykola-mokhnach:
Was kindly requested by @tfcardoso, who even tried to implement this functionality himself and almost succeeded.

### DIFF
--- a/lib/commands/shell.js
+++ b/lib/commands/shell.js
@@ -1,6 +1,7 @@
 import log from '../logger';
 import _ from 'lodash';
-import { util } from 'appium-support';
+import { exec } from 'teen_process';
+import { quote } from 'shell-quote';
 
 let commands = {};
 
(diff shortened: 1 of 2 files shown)


lib/commands/shell.js
```diff
@@ -9,19 +10,38 @@ commands.mobileShell = async function (opts = {}) {
...
+  const adbArgs = [
+    ...this.adb.executable.defaultArgs,
+    'shell',
+    command,
+    ...(_.isArray(args) ? args : [args])
+  ];
+  log.debug(`Running '${this.adb.executable.path} ${quote(adbArgs)}'`);
+  try {
+    const {stdout, stderr} = await exec(this.adb.executable.path, adbArgs, {timeout});
+    if (includeStderr) {
```
jlipps: what's the harm in including `stderr` in the object even if it's not requested?

mykola-mokhnach: I want to keep the backward compatibility

jlipps:
it seems like this is something we might plausibly want in other places. why not make it an optional arg to the appium-adb methods, and then simply supply the correct parameter here?

mykola-mokhnach:
I really don't want to touch that method, since too many things rely on it

jlipps:
OK. if we find we need a similar bit of functionality in the future, we should move it to appium-adb. In any case, adding a default param at the end of the args should be safe. But I'm fine to move forward with this, as-is, for now.

PR #357, opened 2018-04-29
Fix passing suppressKillServer option while creating ADB instance

mykola-mokhnach:
Fixed by @myso-kr in https://github.com/appium/appium/issues/10610


mykola-mokhnach:
I agree, but this would be a topic for more serious refactoring. Care to create a separate issue for it?

jlipps:
@mykola-mokhnach https://github.com/appium/appium/issues/10627

PR #361, opened 2018-05-05
Add missing await

mykola-mokhnach:
I have created a small helper tool for finding missing await calls. It is pretty dumb yet, but managed to find some issues so far.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -70,9 +70,9 @@ helpers.switchContext = async function (name) {
     // else simply suspend proxying to the latter
     if (this.opts.recreateChromeDriverSessions) {
       log.debug("recreateChromeDriverSessions set to true; killing existing chromedrivers");
-      this.stopChromedriverProxies();
+      await this.stopChromedriverProxies();
     } else {
-      this.suspendChromedriverProxy();
+      await this.suspendChromedriverProxy();
     }
   } else {
     throw new Error(`Didn't know how to handle switching to context '${name}'`);
@@ -125,7 +125,7 @@ helpers.startChromedriverProxy = async function (context) {
     // one stopped unexpectedly
     cd.on(Chromedriver.EVENT_CHANGED, (msg) => {
       if (msg.state === Chromedriver.STATE_STOPPED) {
-        this.onChromedriverStop(context);
+        await this.onChromedriverStop(context);
       }
     });
     // save the chromedriver object under the context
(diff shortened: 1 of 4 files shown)


lib/commands/context.js
```diff
@@ -225,7 +225,7 @@ helpers.startChromeSession = async function startChromeSession () {
   this.chromedriver = await this.setupNewChromedriver(opts, this.adb.curDeviceId, this.adb);
   this.chromedriver.on(Chromedriver.EVENT_CHANGED, (msg) => {
     if (msg.state === Chromedriver.STATE_STOPPED) {
-      this.onChromedriverStop(CHROMIUM_WIN);
+      await this.onChromedriverStop(CHROMIUM_WIN);
```
jlipps: this one won't work will it? since it's not in an async function?

lib/commands/context.js
```diff
@@ -125,7 +125,7 @@ helpers.startChromedriverProxy = async function (context) {
     // one stopped unexpectedly
     cd.on(Chromedriver.EVENT_CHANGED, (msg) => {
       if (msg.state === Chromedriver.STATE_STOPPED) {
-        this.onChromedriverStop(context);
+        await this.onChromedriverStop(context);
```
jlipps: this await is not in an async function

lib/commands/context.js
mykola-mokhnach: why? it is defined as async. Perhaps, the name of it is not the best one:

```
// Handle an out-of-band Chromedriver stop event
helpers.onChromedriverStop = async function (context) {
```

lib/commands/context.js
mykola-mokhnach: you are right. eslint does not like it anyway, since it is invoked in a callback. shall I define the callback function as asyncchonous as well or remove await?

jlipps:
for future situations like these, my approach would be:

1. in almost every case, an `async` function should be awaited. even if that means making the context function async.
2. in a case where an async function is not awaited, we should have a comment or it should be obvious that we are wanting to trigger the async method without caring about it finishing
3. any time we have an async function that we are not awaiting, we need to be sure that any errors thrown inside it are handled (otherwise we could get unhandled rejections)
4. for anonymous async functions which aren't themselves awaited, the same rule applies--everything inside should either be guaranteed to not throw an error, or there should be a try/catch that wraps the whole thing

PR #362, opened 2018-05-07
Use activateApp call to restore an app from the background

mykola-mokhnach:
Try to use `activateApp` endpoint while restoring the app from background and fall back to the old approach if this fails.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -100,27 +100,33 @@ commands.background = async function (seconds) {
   await this.adb.goToHome();
   await B.delay(seconds * 1000);
 
-  let args;
+  let args = null;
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -104,31 +104,31 @@ commands.background = async function (seconds) {
...
-      waitPkg: appPackage,
-      waitActivity: appActivity,
-      stopApp: false
-    };
+    try {
+      log.debug(`Activating app '${appPackage}' in order to restore it`);
+      await this.activateApp(appPackage);
+      return true;
+    } catch (ign) {}
+    args = ((appPackage === this.opts.appPackage && appActivity === this.opts.appActivity) ||
```
jlipps: i'm not sure this more readable than a simple if/else conditional

PR #364, opened 2018-05-14
Pass chrome bundle id into CD when possible

imurchie:
Update `appium-chromedriver` and pass in the bundle id of the chrome process, so that webview handling works as expected.

Also, recent changes to `appium-android-bootstrap` necessitate changing the find method, and some tests.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -15,9 +15,38 @@ import { default as unlocker, PIN_UNLOCK, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGE
 
 
 const PACKAGE_INSTALL_TIMEOUT = 90000; // milliseconds
-const CHROME_BROWSERS = ["Chrome", "Chromium", "Chromebeta", "Browser",
-                         "chrome", "chromium", "chromebeta", "browser",
-                         "chromium-browser", "chromium-webview"];
+const CHROME_BROWSERS = [
+  'chrome', 'chromium', 'chromebeta',
+  'Chrome', 'Chromium', 'Chromebeta',
+  'chromium-browser', 'chromium-webview',
+  'browser', 'Browser',
+];
+const CHROME_BROWSER_PACKAGE_ACTIVITY = {
+  chromium: {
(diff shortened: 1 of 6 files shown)


lib/android-helpers.js
```diff
@@ -15,9 +15,38 @@ import { default as unlocker, PIN_UNLOCK, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGE
 
 
 const PACKAGE_INSTALL_TIMEOUT = 90000; // milliseconds
-const CHROME_BROWSERS = ["Chrome", "Chromium", "Chromebeta", "Browser",
-                         "chrome", "chromium", "chromebeta", "browser",
-                         "chromium-browser", "chromium-webview"];
+const CHROME_BROWSERS = [
```
mykola-mokhnach: I'd rather make the verification case-insenitive.

lib/android-helpers.js
```diff
@@ -15,9 +15,38 @@ import { default as unlocker, PIN_UNLOCK, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGE
...
+  'chrome', 'chromium', 'chromebeta',
+  'Chrome', 'Chromium', 'Chromebeta',
+  'chromium-browser', 'chromium-webview',
+  'browser', 'Browser',
+];
+const CHROME_BROWSER_PACKAGE_ACTIVITY = {
```
mykola-mokhnach: do we still need the array above with this mapping?

lib/commands/context.js
```diff
@@ -213,14 +213,18 @@ helpers.startChromeSession = async function startChromeSession () {
...
-    "org.chromium.webview_shell"];
+  const knownPackages = [
+    'org.chromium.chrome.shell',
+    'com.android.chrome',
+    'com.chrome.beta',
+    'org.chromium.chrome',
+    'org.chromium.webview_shell',
+  ];
 
   if (!_.includes(knownPackages, this.opts.appPackage)) {
```
mykola-mokhnach: I think that positive conditions are easier to read

PR #366, opened 2018-05-18
Add adbExecTimeout cap

imurchie:
Add a capability `adbExecTimeout` to customize the amount of time we will wait for `adb` to run. Necessary on some slow systems where timeouts happen a lot (see, for instance, the Espresso driver on Travis).

The signature for `helpers.createADB()` is too long. Change to taking an options object rather than a list of arguments. That is a breaking change.


stride83:
as same , if the capability can have one property which can retry some times when adbExecTimeout exception throw out?
the server version is 1.15.1

PR #367, opened 2018-05-23
Extract basic adb creation into helper function

imurchie:
See https://github.com/appium/appium/issues/10627

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -52,6 +52,27 @@ const UNLOCK_HELPER_PKG_ACTIVITY = ".Unlock";
 
 let helpers = {};
 
+helpers.createBaseADB = async function (opts = {}) {
+  // filter out any unwanted options sent in
+  // this list should be updated as ADB takes more arguments
+  const {
+    javaVersion,
+    adbPort,
+    suppressKillServer,
+    remoteAdbHost,
+    clearDeviceLogsOnStart,
+    adbExecTimeout,
+  } = opts;
+  return await ADB.createADB({
+    javaVersion,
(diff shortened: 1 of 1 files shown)


lib/android-helpers.js
```diff
@@ -135,14 +156,7 @@ helpers.getDeviceInfoFromCaps = async function (opts = {}) {
...
-  const adb = await ADB.createADB({
-    javaVersion: opts.javaVersion,
-    adbPort: opts.adbPort,
-    remoteAdbHost: opts.remoteAdbHost,
-    suppressKillServer: opts.suppressKillServer,
-    clearDeviceLogsOnStart: opts.clearDeviceLogsOnStart,
-    adbExecTimeout: opts.adbExecTimeout,
-  });
+  const adb = await helpers.createBaseADB(opts);
   let udid = opts.udid;
```
mykola-mokhnach: can we also assign these local vars in one-liner, like it's done below?

lib/android-helpers.js
```diff
@@ -52,6 +52,27 @@ const UNLOCK_HELPER_PKG_ACTIVITY = ".Unlock";
 
 let helpers = {};
 
+helpers.createBaseADB = async function (opts = {}) {
+  // filter out any unwanted options sent in
```
mykola-mokhnach: is this really necessary to filter them? unknown options are anyway going to be ignored

lib/android-helpers.js
imurchie: It is one line. The `emPort` should not be initialized from the `opts`.

PR #371, opened 2018-05-28
Add warning for apppackage, appactivity and appwaitpackages

KazuCocoa:
This is just a suggestion.

----

I got feedback from some Appium users in Japan who faced wrong package/activity name.
We can see `Bad component name: com.android.vending` if we inserted a space in `appActivity` for example. For users, it's difficult to estimate the case.

I know the error message by adb command is a common thing for Android developers though. But in the case, users should take much time to dig into adb command world.

```
error: [MJSONWP] Encountered internal error running command: Error: Error occured while starting App. Original error: Error executing adbExec. Original error: 'Command '/Users/itonozomi/Library/Android/sdk/platform-tools/adb -P 5037 -s emulator-5554 shell am start -W -n 'com.android.vending /.AssetBrowserActivity' -S -a android.intent.action.MAIN -c android.intent.category.LAUNCHER -f 0x10200000' exited with code 255'; Stderr: 'Exception occurred while executing:
error: [MJSONWP] java.lang.IllegalArgumentException: Bad component name: com.android.vending
error: [MJSONWP] 	at android.content.Intent.parseCommandArgs(Intent.java:6417)
```

For Appium users, I'd like to try to show more helpful error message than simple adb command error message. I won't insert this kind of messages too much so only insert in initialising.I found it follows https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html . So, roughly, I've added warning messages using `RegExp` like the below.

## What I did
So, what about adding warning message?
(This is warning message. So, if some illegal name happens, Appium should work leaving some warning messages.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -236,10 +236,25 @@ helpers.createADB = async function (opts = {}) {
 };
 
 helpers.validatePackageActivityName = function (name) {
-  let result = /([^a-zA-Z0-9_\.])+/.exec(name);
+  let result = /([^a-zA-Z0-9_.*])+/.exec(name);
   if (result) {
     logger.warn(`'${name}' is expected to only include latin letters, digits, underscore and dot characters.`);
-    logger.warn(`'${name}' <- the first position is ${result.index}. '${result}' are included.`);
+    logger.warn(`'${name.substring(0, result.index + 1)}' <- the first position is ${result.index}.`);
+  }
+};
+
+helpers.validatePackageActivityNames = function (appPackage, appActivity, appWaitPackage, appWaitActivity) {
+  if (appWaitPackage) {
+    this.validatePackageActivityName(appWaitPackage);
+  }
+  if (appWaitActivity) {
+    this.validatePackageActivityName(appWaitActivity);
+  }
+  if (appPackage) {
+    this.validatePackageActivityName(appPackage);
+  }
+  if (appActivity) {
+    this.validatePackageActivityName(appActivity);
   }
 };
 
@@ -250,16 +265,9 @@ helpers.getLaunchInfo = async function (adb, opts) {
     return;
   }
 
-  if (!appWaitPackage) {
-    this.validatePackageActivityName(appWaitPackage);
-  }
-  if (!appWaitActivity) {
-    this.validatePackageActivityName(appWaitActivity);
-  }
+  this.validatePackageActivityNames(appPackage, appActivity, appWaitPackage, appWaitActivity);
 
   if (appPackage && appActivity) {
-    this.validatePackageActivityName(appPackage);
-    this.validatePackageActivityName(appActivity);
     return;
   }
 
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -239,11 +239,13 @@ helpers.validatePackageActivityName = function (name) {
   let result = /([^\w.*])+/.exec(name);
   if (result) {
     logger.warn(`'${name}' is expected to only include latin letters, digits, underscore and dot characters.`);
-    logger.warn(`'${name.substring(0, result.index + 1)}' <- the first position is ${result.index}.`);
+    logger.warn(`'${name.substring(0, result.index + 1)}' <- the first non-matching character occurrence is at index ${result.index}.`);
   }
 };
 
-helpers.validatePackageActivityNames = function (appPackage, appActivity, appWaitPackage, appWaitActivity) {
+helpers.validatePackageActivityNames = function (opts) {
+  let {appPackage, appActivity, appWaitPackage, appWaitActivity} = opts;
+
   if (appWaitPackage) {
     this.validatePackageActivityName(appWaitPackage);
   }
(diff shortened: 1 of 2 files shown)


lib/android-helpers.js
```diff
@@ -242,6 +242,13 @@ helpers.getLaunchInfo = async function (adb, opts) {
     return;
   }
   if (appPackage && appActivity) {
+    if (!new RegExp(/^[a-zA-Z0-9_\.]+$/).test(appPackage)) {
+      logger.warn(`appPackage, "${appPackage}", should be ^[a-zA-Z0-9_\.]+$`);
```
mykola-mokhnach: I'd rather say `'${appPackage}' is expected to only include latin letters, digits, underscore and dot characters.`

mykola-mokhnach: also, it might be handy to have a following note about where the first occurrence of an unexpected character has been detected, like

```
blabla$ <- at position 9
```

like compilers illustrate this

lib/android-helpers.js
```diff
@@ -259,6 +266,10 @@ helpers.getLaunchInfo = async function (adb, opts) {
   }
   if (!appWaitActivity) {
     appWaitActivity = appActivity;
+  } else {
+    if (!new RegExp(/^[a-zA-Z0-9_\.]+$/).test(appWaitActivity)) {
```
mykola-mokhnach: should appWaitPackage have this verification?

lib/android-helpers.js
```diff
@@ -235,20 +235,31 @@ helpers.createADB = async function (opts = {}) {
...
-    }
-    if (!new RegExp(/^[a-zA-Z0-9_\.]+$/).test(appActivity)) {
-      logger.warn(`appActivity, "${appActivity}", should be ^[a-zA-Z0-9_\.]+$`);
-    }
 
+  if (!appWaitPackage) {
+    this.validatePackageActivityName(appWaitPackage);
+  }
+  if (!appWaitActivity) {
+    this.validatePackageActivityName(appWaitActivity);
```
KazuCocoa: I forgot with ”*” case.

lib/android-helpers.js
KazuCocoa: Yeah, since users can set value in appWaitpackage by themselves.

lib/android-helpers.js
```diff
@@ -235,13 +235,31 @@ helpers.createADB = async function (opts = {}) {
   return adb;
 };
 
+helpers.validatePackageActivityName = function (name) {
+  let result = /([^a-zA-Z0-9_\.])+/.exec(name);
```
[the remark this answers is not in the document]
imurchie: Yes, with no capture the parentheses are unnecessary. Also, `^a-zA-Z0-9_` can be reduced to `\W` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#character-classes).

lib/android-helpers.js
```diff
@@ -235,12 +235,38 @@ helpers.createADB = async function (opts = {}) {
   return adb;
 };
 
+helpers.validatePackageActivityName = function (name) {
+  let result = /([^a-zA-Z0-9_.*])+/.exec(name);
+  if (result) {
+    logger.warn(`'${name}' is expected to only include latin letters, digits, underscore and dot characters.`);
+    logger.warn(`'${name.substring(0, result.index + 1)}' <- the first position is ${result.index}.`);
```
KazuCocoa: 📝 
`'bar ' <- the first position is 3. ' , ' are included.`

lib/android-helpers.js
```diff
@@ -235,12 +235,38 @@ helpers.createADB = async function (opts = {}) {
   return adb;
 };
 
+helpers.validatePackageActivityName = function (name) {
+  let result = /([^\w.*])+/.exec(name);
+  if (result) {
+    logger.warn(`'${name}' is expected to only include latin letters, digits, underscore and dot characters.`);
+    logger.warn(`'${name.substring(0, result.index + 1)}' <- the first position is ${result.index}.`);
```
mykola-mokhnach: ... the first non-matching character occurrence is at index ...

lib/android-helpers.js
```diff
@@ -235,12 +235,38 @@ helpers.createADB = async function (opts = {}) {
...
+  }
+};
+
+helpers.validatePackageActivityNames = function (appPackage, appActivity, appWaitPackage, appWaitActivity) {
```
mykola-mokhnach: you can pass `opts` in arguments and unpack it then. It would be easier to read

lib/android-helpers.js
```diff
@@ -235,12 +235,40 @@ helpers.createADB = async function (opts = {}) {
...
+  if (result) {
+    logger.warn(`'${name}' is expected to only include latin letters, digits, underscore and dot characters.`);
+    logger.warn(`'${name.substring(0, result.index + 1)}' <- the first non-matching character occurrence is at index ${result.index}.`);
+  }
+};
+
+helpers.validatePackageActivityNames = function (opts) {
+  let {appPackage, appActivity, appWaitPackage, appWaitActivity} = opts;
+
+  if (appWaitPackage) {
```
mykola-mokhnach:
```
for (key of ['appPackage', 'appActivity', 'appWaitPackage', 'appWaitActivity']) {
  if (!opts[key]){
    continue;
  }
  this.validatePackageActivityName(opts[key]);
}
```

KazuCocoa:
thanks for the review!

PR #372, opened 2018-05-29
Add an option to install testOnly APKs

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10758


jlipps:
got it

PR #373, opened 2018-06-06
Fix implicit wait for finding elements

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10809


imurchie:
Published in `appium-android-driver@3.0.2`.

PR #375, opened 2018-06-08
Install settings app on AVD for Locales

caguilar187:
Related to [9444](https://github.com/appium/appium/issues/9944) 

This error occurs when trying to change the language on an emulator.


KazuCocoa:
cc @imurchie 
Settings app should be available on both emulators and real devices since the app has some features to change device preferences like locale.

PR #376, opened 2018-06-09
fix issue in find element error handling, so implicit wait can work

jlipps:
cc @mykola-mokhnach: unfortunately we can't do `instanceof` checks where the LHS could be an instance of a class from a different version of a module as the RHS.

### DIFF
--- a/lib/commands/find.js
+++ b/lib/commands/find.js
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import { errors } from 'appium-base-driver';
+import { errors, isErrorType } from 'appium-base-driver';
 
 
 let helpers = {}, extensions = {};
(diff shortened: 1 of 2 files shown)


lib/commands/find.js
```diff
@@ -35,7 +35,17 @@ helpers.findElOrEls = async function (strategy, selector, mult, context = '') {
...
+      // if the error that comes back is from a proxied request, we need to
+      // unwrap it to its actual protocol error first
+      if (_.isFunction(err.getActualError)) {
+        err = err.getActualError(); // eslint-disable-line no-ex-assign
+      }
+
+      // now we have to inspect the error to determine if it is a no such
+      // element error, based on the shape of the error object from
+      // appium-base-driver
+      if (err.jsonwpCode === 7 || err.error === "no such element") {
```
mykola-mokhnach: why should we use magic numbers here? I'd rather use `errors.NoSuchElementError.error()/code()`

mykola-mokhnach:
@jlipps Thanks for pointing that out. I didn't know

lib/commands/find.js
jlipps: right you are! will fix

jlipps:
@mykola-mokhnach yeah i learned this the hard way from the early days of writing basedriver. can you imagine the crazy bugs that result from this? it's why we have [isErrorType()](https://github.com/appium/appium-base-driver/blob/master/lib/protocol/errors.js#L747)... actually come to think of it, we could just export this function and use it

mykola-mokhnach:
yep, I was actually wondering why such function exists. It might be useful to add a docstring there with a detailed purpose explanation, otherwise this limitation will stay in your head only ;)

Also, you say it is safe to use `instanceof` in scope of a single package, so https://github.com/appium/appium-base-driver/blob/165543e16dc8ad2748d9ce0316b5ec0b5856eaae/lib/jsonwp-proxy/proxy.js#L164 should work?

jlipps:
@mykola-mokhnach that will only be safe it is guaranteed that `err` is only ever instantiated from within basedriver. In other words, at some point there is a `new ProxyRequestError(...)` which results in the error which is checked here. And that instantiation must also be in basedriver. Can we guarantee this?

jlipps:
published in 3.0.3

PR #377, opened 2018-06-09
tweak android date format

KazuCocoa:
addressed : https://github.com/appium/appium/issues/10832

I made sure the format worked with emulator OS 4, 5, 6, 7, 8 and real device with 6.

Wha do you think?

----

- iOS returns JS's `Date` (both real devices and simulators):
https://github.com/appium/appium-ios-driver/blob/17ac721c3c1cbefce28445a036daf1a50f4c6d80/lib/commands/general.js#L20 
- Android's `shell date` can format.

```
1|generic_x86:/ $ date --help
usage: date [-u] [-r FILE] [-d DATE] [+DISPLAY_FORMAT] [-D SET_FORMAT] [SET]

Set/get the current date/time. With no SET shows the current date.

Default SET format is "MMDDhhmm[[CC]YY][.ss]", that's (2 digits each)
month, day, hour (0-23), and minute. Optionally century, year, and second.
Also accepts "@UNIXTIME[.FRACTION]" as seconds since midnight Jan 1 1970.

-d	Show DATE instead of current time (convert date format)
-D	+FORMAT for SET or -d (instead of MMDDhhmm[[CC]YY][.ss])
-r	Use modification time of FILE instead of current date
-u	Use UTC instead of current timezone

+FORMAT specifies display format string using strftime(3) syntax:

%% literal %             %n newline              %t tab
%S seconds (00-60)       %M minute (00-59)       %m month (01-12)
%H hour (0-23)           %I hour (01-12)         %p AM/PM
%y short year (00-99)    %Y year                 %C century
%a short weekday name    %A weekday name         %u day of week (1-7, 1=mon)
%b short month name      %B month name           %Z timezone name
%j day of year (001-366) %d day of month (01-31) %e day of month ( 1-31)
%N nanosec (output only)

%U Week of year (0-53 start sunday)   %W Week of year (0-53 start monday)
%V Week of year (1-53 start monday, week < 4 days not part of this year)

%D = "%m/%d/%y"    %r = "%I : %M : %S %p"   %T = "%H:%M:%S"   %h = "%b"
%x locale date     %X locale time           %c locale date/time
```

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -28,7 +28,8 @@ commands.doSendKeys = async function (params) {
 commands.getDeviceTime = async function () {
   log.info('Attempting to capture android device date and time');
   try {
-    let out = await this.adb.shell(['date']);
+    // format: Sat Jun 09 2018 11:17:42 GTM+0900 (JST)
+    let out = await this.adb.shell(['date', '\"+%a\ %b\ %d\ %Y\ %X\ GTM%z\ \(%Z\)\"']);
     return out.trim();
   } catch (err) {
     log.errorAndThrow(`Could not capture device date and time: ${err}`);
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -28,13 +28,12 @@ commands.doSendKeys = async function (params) {
 /**
  * Get device time
  * @return {String} Return value following iso-8601 format like `2018-06-09T16:21:54+0900`
- * @throws {Error} If the adb command doesn't return device date, error throes
+ * @throws {Error} If the adb command doesn't return device date
  */
 commands.getDeviceTime = async function () {
   log.info('Attempting to capture android device date and time');
   try {
-    let out = await this.adb.shell(['date', '+%Y-%m-%dT%T%z']);
-    return out.trim();
+    return (await this.adb.shell(['date', '+%Y-%m-%dT%T%z'])).trim();
   } catch (err) {
     log.errorAndThrow(`Could not capture device date and time: ${err}`);
   }
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -25,10 +25,14 @@ commands.doSendKeys = async function (params) {
   return await this.bootstrap.sendAction('setText', params);
 };
 
+/**
+ * Get device time
+ * @return {String} Return value following iso-8601 format like `2018-06-09T16:21:54+0900`
+ * @throws {Error} If the adb command doesn't return device date, error throes
+ */
 commands.getDeviceTime = async function () {
   log.info('Attempting to capture android device date and time');
   try {
-    // format: 2018-06-09T16:21:54+0900
     let out = await this.adb.shell(['date', '+%Y-%m-%dT%T%z']);
     return out.trim();
   } catch (err) {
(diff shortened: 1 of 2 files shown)


KazuCocoa:
In Ruby, both formats can parse as same, so this change doesn't affect Ruby client though.

```rb
# "Sat Jun  9 10:31:40 JST 2018"
Date.parse "Sat Jun  9 10:31:40 JST 2018"
=> #<Date: 2018-06-09 ((2458279j,0s,0n),+0s,2299161j)>

# "Sat Jun 09 2018 10:30:13 GMT+0900 (JST)"
Date.parse "Sat Jun 09 2018 10:30:13 GMT+0900 (JST)"
=> #<Date: 2018-06-09 ((2458279j,0s,0n),+0s,2299161j)>

```

mykola-mokhnach:
The initial proposal was to be able to provide a custom format string as function argument, so we could alter this is base driver first. Also, I'd love to see the default format as iso-8601, since its parsing is for sure supported by majority of the corresponding clients.

KazuCocoa:
yeah.
Android returns `"Sat Jun  9 10:31:40 JST 2018"` by default and JS can't parse the string like `new Date("Sat Jun  9 10:31:40 JST 2018")` nor `Date.parse`. So, I customised `shell date` format first to make the return value manageable by JS. (And I thought the format we could get by `new Data()` was the most JS friendly)
I'll update the format to `iso-8601`, anyway.

lib/commands/general.js
```diff
@@ -28,8 +28,8 @@ commands.doSendKeys = async function (params) {
 commands.getDeviceTime = async function () {
   log.info('Attempting to capture android device date and time');
   try {
-    // format: Sat Jun 09 2018 11:17:42 GTM+0900 (JST)
-    let out = await this.adb.shell(['date', '\"+%a\ %b\ %d\ %Y\ %X\ GTM%z\ \(%Z\)\"']);
+    // format: 2018-06-09T16:21:54+0900
```
mykola-mokhnach: it probably makes sense to mention this in the docstring, that the resulting date string format conforms to ISO-8601 by default.

mykola-mokhnach:
@KazuCocoa Would you like to include the optional format argument into this PR or create a new one?

KazuCocoa:
I'd like to create another PR for the option to keep change small.

And I'll add docstring later :+1:

lib/commands/general.js
```diff
@@ -25,10 +25,15 @@ commands.doSendKeys = async function (params) {
   return await this.bootstrap.sendAction('setText', params);
 };
 
+/**
+ * Get device time
+ * @return {String} Return value following iso-8601 format like `2018-06-09T16:21:54+0900`
+ * @throws {Error} If the adb command doesn't return device date, error throes
```
mykola-mokhnach: it is enough to keep "If the adb command doesn't return a device date" part

lib/commands/general.js
```diff
@@ -25,10 +25,15 @@ commands.doSendKeys = async function (params) {
...
+ */
 commands.getDeviceTime = async function () {
   log.info('Attempting to capture android device date and time');
   try {
-    let out = await this.adb.shell(['date']);
+    let out = await this.adb.shell(['date', '+%Y-%m-%dT%T%z']);
```
mykola-mokhnach: `return (await this.adb.shell(['date', '+%Y-%m-%dT%T%z'])).trim()`

KazuCocoa:
thanks for the review!

PR #381, opened 2018-06-19
Retry installation of Unicode IME helper

mykola-mokhnach:
See https://discuss.appium.io/t/unicodeime-debug-apk-failure-install-failed-already-exists-attempt-to-re-install-io-appium-android-ime-without-first-uninstalling/21270 for more details.

Uninstalling the app from the device first sometimes helps to resolve the problem.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -11,8 +11,8 @@ import Bootstrap from 'appium-android-bootstrap';
 import B from 'bluebird';
 
 import ADB from 'appium-adb';
-import { default as unlocker, PIN_UNLOCK, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGERPRINT_UNLOCK } from './unlock-helpers';
-
+import { default as unlocker, PIN_UNLOCK, PASSWORD_UNLOCK,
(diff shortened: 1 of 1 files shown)


lib/android-helpers.js
```diff
@@ -400,13 +401,18 @@ helpers.installOtherApks = async function (otherApps, adb, opts) {
 helpers.initUnicodeKeyboard = async function (adb) {
   logger.debug('Enabling Unicode keyboard support');
   logger.debug("Pushing unicode ime to device...");
-  await adb.install(unicodeIMEPath, {replace: false});
+  try {
+    await adb.install(unicodeIMEPath, {replace: false});
+  } catch (err) {
+    await adb.uninstallApk(UNICODE_IME_PKG_ID);
```
jlipps: I think a log message here saying we're retrying would be helpful for reading the logs during debugging

mykola-mokhnach: uninstallApk does write logs anyway, so I thought it might too much

jlipps: sure, i just meant for us to understand why we'd see an install, then an uninstall, then another install. presumably the existing logs don't tell us that there's a retry going on. but it's not a big deal either way.

PR #397, opened 2018-07-25
add comma for package validation to relax the logic

KazuCocoa:
https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/caps.md allows to use `,` in `appWaitActivity`.

https://github.com/appium/appium/issues/11101#issuecomment-407682259

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -243,12 +243,12 @@ helpers.validatePackageActivityNames = function (opts) {
       continue;
     }
 
-    const match = /([^\w.*])+/.exec(name);
+    const match = /([^\w.*,])+/.exec(name);
     if (!match) {
       continue;
     }
 
-    logger.warn(`'${name}' is expected to only include latin letters, digits, underscore, dot and asterisk characters.`);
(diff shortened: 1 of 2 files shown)


lib/android-helpers.js
```diff
@@ -243,12 +243,12 @@ helpers.validatePackageActivityNames = function (opts) {
...
     }
 
-    const match = /([^\w.*])+/.exec(name);
+    const match = /([^\w.*,])+/.exec(name);
     if (!match) {
       continue;
     }
 
-    logger.warn(`'${name}' is expected to only include latin letters, digits, underscore, dot and asterisk characters.`);
+    logger.warn(`'${name}' is expected to only include latin letters, digits, underscore, dot, comma and asterisk characters.`);
```
imurchie: While this code is being updated, it might be nice to add what capability is being objected. Right now it just shows the value. Something like
```js
logger.warn(`Capability '${key}' is expected to only include latin letters, digits, underscore, dot, comma and asterisk characters.`);
logger.warn(`Current value '${name}' has non-matching character at index ${match.index}: '${name.substring(0, match.index + 1)}'`);
```

PR #418, opened 2018-09-06
Add get geolocation handler

mykola-mokhnach:
This should works for both real devices and emulators as long as Appium Settings is installed there.

### DIFF
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -2,6 +2,7 @@ import log from '../logger';
 import _ from 'lodash';
 import { errors } from 'appium-base-driver';
 import B from 'bluebird';
+import { util } from 'appium-support';
 
 let commands = {}, helpers = {}, extensions = {};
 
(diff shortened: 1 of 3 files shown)


lib/commands/network.js
```diff
@@ -127,6 +128,15 @@ commands.setGeoLocation = async function (location) {
   return await this.adb.setGeoLocation(location, this.isEmulator());
 };
 
+commands.getGeoLocation = async function () {
+  const {latitude, longitude, altitude} = await this.adb.getGeoLocation();
+  return {
+    latitude: parseFloat(latitude),
+    longitude: parseFloat(longitude),
+    altitude: parseFloat(util.hasValue(altitude) ? altitude : 0),
```
KazuCocoa: ✅

test/unit/commands/network-specs.js
```diff
@@ -149,6 +149,18 @@ describe('Network', function () {
...
+  describe('getGeoLocation', function () {
+    it('should get location', async function () {
+      adb.getGeoLocation.returns({
+        latitude: '1.1',
+        longitude: '2.2',
+      });
+      const {latitude, longitude, altitude} = await driver.getGeoLocation();
+      isNaN(latitude).should.be.false;
+      isNaN(longitude).should.be.false;
+      isNaN(altitude).should.be.false;
```
dpgraham: I think we should be checking if these are numbers. This will be a false-positive if latitude, longitude or altitude is `undefined` or `null`.

mykola-mokhnach: how about Number.isNaN?

dpgraham: Yeah I can't find anything for checking that the type is a number for chai-as-promised.

This should be okay as-is.

PR #420, opened 2018-09-11
Return the current location from setGeoLocation

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/7400

### DIFF
--- a/test/unit/commands/network-specs.js
+++ b/test/unit/commands/network-specs.js
@@ -143,7 +143,7 @@ describe('Network', function () {
     });
   });
   describe('setGeoLocation', function () {
-    it('should set location', async function () {
+    it('should return location in use after setting', async function () {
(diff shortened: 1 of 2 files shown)


test/unit/commands/network-specs.js
```diff
@@ -145,8 +145,16 @@ describe('Network', function () {
   describe('setGeoLocation', function () {
     it('should set location', async function () {
```
imurchie: This is really not test that it "should set location" so much as that it returns the geo location.

mykola-mokhnach: which name would be better?

imurchie: Something like?
```js
it('should return location in use after setting', async function () {
```

mykola-mokhnach: fixed

PR #429, opened 2018-09-23
Do not install unlock helper if skipUnlock is set to true

mykola-mokhnach:
This will help to save time for session init if `skipUnlock` capability is truth

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -632,7 +632,7 @@ helpers.initDevice = async function (adb, opts) {
   if (opts.unicodeKeyboard) {
     defaultIME = await helpers.initUnicodeKeyboard(adb);
   }
-  if (_.isUndefined(opts.unlockType)) {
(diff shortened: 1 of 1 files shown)


lib/android-helpers.js
```diff
@@ -632,7 +632,7 @@ helpers.initDevice = async function (adb, opts) {
   if (opts.unicodeKeyboard) {
     defaultIME = await helpers.initUnicodeKeyboard(adb);
   }
-  if (_.isUndefined(opts.unlockType)) {
+  if (_.isUndefined(opts.unlockType) && !opts.skipUnlock) {
```
imurchie: Do we log somewhere if the user gives an unlock type _and_ sets `skipUnlock` to true?

jlipps: > (who is gonna do such unlogical stuff though?)

you're still asking that question after working on appium for so long?

mykola-mokhnach: Yes I still do believe in people...

PR #437, opened 2018-10-03
Fix resource strings handling

mykola-mokhnach:
This fixes the result of `getStrings` API call, since client libs expect `Map<String, String>` as the resulting type and we might put a list there if the particular resource value has plural options

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -159,16 +159,28 @@ commands.getStrings = async function (language) {
     log.info(`No language specified, returning strings for: ${language}`);
   }
 
+  // Clients require the resulting mapping to have both keys
+  // and values of type string
+  const preprocessStringsMap = (mapping) => {
+    const result = {};
+    for (const [key, value] of _.toPairs(mapping)) {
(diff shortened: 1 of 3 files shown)


lib/commands/general.js
```diff
@@ -159,16 +159,28 @@ commands.getStrings = async function (language) {
     log.info(`No language specified, returning strings for: ${language}`);
   }
 
+  // Clients require the resulting mapping to have both keys
+  // and values of type string
+  const preprocessStringsMap = (mapping) => {
+    const result = {};
+    for (const [key, value] of _.toPairs(mapping)) {
```
imurchie: This could be done with `_.reduce`, and not produce an extra array. Something like
```js
const preprocessStringsMap = function preprocessStringsMap (mapping) {
  return _.reduce(mapping, function (result, key, value) {
    result[key] = _.isString(value) ? value : JSON.stringify(value);
    return result;
  }, {});
};
```

mykola-mokhnach: np, changed

PR #455, opened 2018-10-23
Deflake some tests

imurchie:


test/functional/commands/keyboard/keyboard-e2e-specs.js
```diff
@@ -138,6 +147,21 @@ describe('keyboard', function () {
       driver = new AndroidDriver();
       await driver.createSession(defaultAsciiCaps);
 
+      // on Travis the device is sometimes not unlocked
+      await waitForCondition(async function () {
+        const keyguardEls = await driver.findElements('id', 'com.android.systemui:id/keyguard_header');
```
jlipps: why not use isLocked?

PR #458, opened 2018-10-31
Update Enter button locator for Samsung devices

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/11605


mykola-mokhnach:
No it does not. Is it possible to simulator pin lock in the emulator?

dpgraham:
I think it does, but I won't let that hold up merging this PR.

PR #461, opened 2018-11-05
Add hidden api policy

KazuCocoa:
I forgot to add below for Android P..

- It needs for changing animation scale so far since `java.lang.NoSuchMethodException: setAnimationScales [class [F]` can happen.
    - It only needs when animation scale changes
    - But considering safe, I have called `setDefaultHiddenApiPolicy` every time

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -249,6 +249,11 @@ class AndroidDriver extends BaseDriver {
 
     if (this.opts.disableWindowAnimation) {
       if (await this.adb.isAnimationOn()) {
+        if (parseInt(this.caps.platformVersion, 10) >= 28) {
+          log.warn('Relaxing hidden api policy to manage animation scale');
+          await this.adb.setHiddenApiPolicy('1');
+        }
+
         log.info('Disabling window animation as it is requested by "disableWindowAnimation" capability');
         await this.adb.setAnimationState(false);
         this._wasWindowAnimationDisabled = true;
@@ -408,6 +413,11 @@ class AndroidDriver extends BaseDriver {
     if (this._wasWindowAnimationDisabled) {
       log.info('Restoring window animation state');
       await this.adb.setAnimationState(true);
+
+      if (parseInt(this.caps.platformVersion, 10) >= 28) {
+        log.warn('Restoring hidden api policy to the device default configuration');
+        await this.adb.setDefaultHiddenApiPolicy();
+      }
     }
     if (this.opts.reboot) {
       let avdName = this.opts.avd.replace('@', '');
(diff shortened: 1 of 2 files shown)


lib/driver.js
```diff
@@ -249,6 +249,11 @@ class AndroidDriver extends BaseDriver {
 
     if (this.opts.disableWindowAnimation) {
       if (await this.adb.isAnimationOn()) {
+        if (parseInt(this.caps.platformVersion, 10) >= 28) { // API level 28 is Android P
```
mykola-mokhnach: it is safe to call adb.getApiVersion

lib/driver.js
```diff
@@ -409,6 +414,14 @@ class AndroidDriver extends BaseDriver {
       log.info('Restoring window animation state');
       await this.adb.setAnimationState(true);
     }
+
+    // This was necessary to change animation scale over Android P
+    // Calls below everytime to keep the api policy safe for the security
+    if (parseInt(this.caps.platformVersion, 10) >= 28) {
```
mykola-mokhnach: do we want to call it inside `if (this._wasWindowAnimationDisabled) {` block?

mykola-mokhnach:
I would assume uia2 driver and espresso also need this change

mykola-mokhnach:
It is strange why travis cannot checkout properly from the branch %/

lib/driver.js
mykola-mokhnach: I'm just concerned about performance. Each new adb command execution will slow down session deletion

KazuCocoa: I see. I can agree with you 👍

lib/driver.js
```diff
@@ -233,13 +234,14 @@ class AndroidDriver extends BaseDriver {
...
 
-    // set actual device name, udid, platform version, screen size, model and manufacturer details
+    // set actual device name, udid, platform version, screen size, model, manufacturer details and device API level.
     this.caps.deviceName = this.adb.curDeviceId;
     this.caps.deviceUDID = this.opts.udid;
     this.caps.platformVersion = await this.adb.getPlatformVersion();
     this.caps.deviceScreenSize = await this.adb.getScreenSize();
     this.caps.deviceModel = await this.adb.getModel();
     this.caps.deviceManufacturer = await this.adb.getManufacturer();
+    this.caps.deviceApiLevel = await this.adb.getApiLevel();
```
mykola-mokhnach: It is safe to call `getApiLevel` without assigning it to a cap, since the value there is cached per session and does not invoke extra command line calls.

Also, I'd say the caps should be in sync in both uia1 and uia2 drivers.

PR #462, opened 2018-11-05
Update hide keyboard algorithm

mykola-mokhnach:
Added the verification of keyboard hidden state.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -5,7 +5,7 @@ import B from 'bluebird';
 import log from '../logger';
 import { NATIVE_WIN } from '../webview-helpers';
 import moment from 'moment';
-
+import { waitForCondition } from 'asyncbox';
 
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -68,14 +68,22 @@ commands.isKeyboardShown = async function () {
...
-  }
-
-  if (canCloseKeyboard) {
-    return await this.back();
+  if (isKeyboardShown) {
+    if (canCloseKeyboard) {
+      // Press escape
+      await this.adb.keyevent(111);
+    }
+    try {
```
jlipps: I don't understand: if we can't close the keyboard, we still wait and wait until it is just closed on its own?

mykola-mokhnach: yes, it is just in case we are in the middle of some running animation

PR #477, opened 2018-12-02
Calls helpers.ensureDeviceLocale if the opts has either language or locale

KazuCocoa:
https://appium.slack.com/archives/C02956V3H/p1543739923001400

`language` or `locale` is mandatory. `localeScript` is option.
We can reduce warning messages for no `language` and no `locale` case.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -635,7 +635,9 @@ helpers.initDevice = async function (adb, opts) {
     await helpers.setMockLocationApp(adb, SETTINGS_HELPER_PKG_ID);
   }
 
-  await helpers.ensureDeviceLocale(adb, opts.language, opts.locale, opts.localeScript);
+  if (opts.language || opts.locale) {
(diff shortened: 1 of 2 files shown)


lib/android-helpers.js
```diff
@@ -635,7 +635,9 @@ helpers.initDevice = async function (adb, opts) {
     await helpers.setMockLocationApp(adb, SETTINGS_HELPER_PKG_ID);
   }
 
-  await helpers.ensureDeviceLocale(adb, opts.language, opts.locale, opts.localeScript);
+  if (opts.language || opts.locale) {
```
mykola-mokhnach: should it be `||` or `&&` ?

KazuCocoa: `||` is fine here since if users set only `language` (or `locale`), then we should show a message in `ensureDeviceLocale` to guide them.
We can skip calling the `ensureDeviceLocale` when they do not specify both language and locale in their capability.

PR #480, opened 2018-12-13
Introduce skipDeviceInitialization capability

AnnaWyrwal:
New capability is introduced, which is named 'skipDeviceInitialization' to allow user for omitting phase of device initialization (waiting for device, pushing settings app, permissions). It allows to improve startup performance when device was already used for tests and all required things was setup.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -632,16 +632,16 @@ helpers.initDevice = async function (adb, opts) {
     logger.info(`'skipDeviceInitialization' is set. Skipping device initialization.`);
   } else {
     await adb.waitForDevice();
-
     // pushSettingsApp required before calling ensureDeviceLocale for API Level 24+
     await helpers.pushSettingsApp(adb);
-    if (!opts.avd) {
-      await helpers.setMockLocationApp(adb, SETTINGS_HELPER_PKG_ID);
-    }
+  }
 
-    if (opts.language || opts.locale) {
-      await helpers.ensureDeviceLocale(adb, opts.language, opts.locale, opts.localeScript);
-    }
+  if (!opts.avd) {
+    await helpers.setMockLocationApp(adb, SETTINGS_HELPER_PKG_ID);
+  }
+
+  if (opts.language || opts.locale) {
+    await helpers.ensureDeviceLocale(adb, opts.language, opts.locale, opts.localeScript);
   }
 
   await adb.startLogcat();
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -628,7 +628,12 @@ helpers.verifyUnlock = async function (adb) {
 };
 
 helpers.initDevice = async function (adb, opts) {
+  if (opts.skipDeviceInitialization) {
+    logger.info(`'skipDeviceInitialization' is set. Skipping device initialization.`);
+    return await adb.defaultIME();
+  }
   await adb.waitForDevice();
+
   // pushSettingsApp required before calling ensureDeviceLocale for API Level 24+
   await helpers.pushSettingsApp(adb);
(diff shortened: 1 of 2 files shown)


lib/android-helpers.js
```diff
@@ -628,7 +628,12 @@ helpers.verifyUnlock = async function (adb) {
 };
 
 helpers.initDevice = async function (adb, opts) {
+  if (opts.skipDeviceInitialization) {
+    logger.info(`'skipDeviceInitialization' is set. Skipping device initialization.`);
+    return await adb.defaultIME();
+  }
   await adb.waitForDevice();
```
mykola-mokhnach: I think it is still necessary to wait for device and to start logcat even if the other stuff is skipped

mykola-mokhnach: I'm also not quite sure about the unicode keyboard. Isn't it reset on session deletion?

AnnaWyrwal: I thought so either, but I executed few tests with skips for all and it worked. I can perform additional tests (on other devices). Or just add that if it's really required.

AnnaWyrwal: It is, but what we skip is change of the keyboard:
```
  logger.debug(`Unsetting previous IME ${defaultIME}`);
  const appiumIME = `${SETTINGS_HELPER_PKG_ID}/.UnicodeIME`;
  logger.debug(`Setting IME to '${appiumIME}'`);
  await adb.enableIME(appiumIME);
  await adb.setIME(appiumIME);
```
the return value stays the same, so if we have valid keyboard at the beginning, then the behavior should be the same.

mykola-mokhnach: > And if we want to skip initialization, then we assume that device is ready for them and we don't want to change it's state.

It might not be what a user wants. Providing this option will render the `unicodeKeyboard` capability invalid, because the IME is not going to be changed to the unicode one.

mykola-mokhnach: I've checked waitForDevice implementation and everything it does is the execution of wait-for-device and ping commands. I would say it is safe to skip it

AnnaWyrwal: That's good point. I'll change that.

AnnaWyrwal: Done.

AnnaWyrwal: Great, thanks! So I left `waitForDevice` skipped (when `skipDeviceInitialization` is set) and restored `adb.startLogcat` to be executed independently.

lib/android-helpers.js
```diff
@@ -628,16 +628,22 @@ helpers.verifyUnlock = async function (adb) {
...
+
+    // pushSettingsApp required before calling ensureDeviceLocale for API Level 24+
+    await helpers.pushSettingsApp(adb);
+    if (!opts.avd) {
+      await helpers.setMockLocationApp(adb, SETTINGS_HELPER_PKG_ID);
+    }
 
-  if (opts.language || opts.locale) {
-    await helpers.ensureDeviceLocale(adb, opts.language, opts.locale, opts.localeScript);
+    if (opts.language || opts.locale) {
```
mykola-mokhnach: locale setting can also be preserved. It is only changed if `language` or `locale` options are present

AnnaWyrwal: Actually, now I see where I was wrong. I preserved all actions dependent on other caps. They can be easily skipped anyway, if they are not needed. Sorry for that. I'm just starting jumping into that project from the code side.

mykola-mokhnach: np, we have code review procedure exactly for this purpose

AnnaWyrwal:
Thanks. Sure, I'll prepare PR for documentation as well.

PR #488, opened 2019-01-10
fix: update node.js engine requirement to 7.6.0+

DekusDenial:
to avoid node.js version in which async function is not supported

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "mobile",
     "mobile testing"
   ],
-  "version": "4.8.0",
+  "version": "4.9.0",
   "author": "appium",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -19,9 +19,9 @@
   "bugs": {
     "url": "https://github.com/appium/appium-android-driver/issues"
   },
-  "engines": [
-    "node"
-  ],
+  "engines": {
+    "node": ">=7.6.0
```
imurchie: Missing `"`.

package.json
```diff
@@ -9,7 +9,7 @@
     "mobile",
     "mobile testing"
   ],
-  "version": "4.8.0",
+  "version": "4.9.0",
```
imurchie: We never manually bump versions. This is done as part of the release process.

DekusDenial:
@imurchie @jlipps 
I understand your concern and agree that it can be better documented. One suggestion is to have an explicit fail-fast node.js version check because even if you set the engines flag on the package.json, it's just shown as an advisory warning. To avoid seeing a bunch of `Error trying to install Chromedriver binary. Waiting and trying again. Unexpected token function` error before realizing the node.js version the env has isn't supporting async function. It's more straightforward from the perspective of someone who is not familiar with node.js.

Thanks for the feedback.

PR #493, opened 2019-01-23
Avoid throwing an error if Appium Settings fails to provide geolocation values

mykola-mokhnach:
Geolocation providers require some time to get the actual info. Also, one doesn't always care about getting the current geolocation value geolocation is being set.


mykola-mokhnach:
https://travis-ci.org/appium/appium-uiautomator2-driver/jobs/483266968#L6348

jlipps:
hopefully no one was relying on the exception behavior :-)

PR #494, opened 2019-01-28
Do not apply implicit wait if looking for multiple elements

mykola-mokhnach:
WebDrivers also don't not do this in browsers for a reason. Also, multiple elements lookup is used to verify that an element is NOT present on screen, which might create significant delays with the previous implementation if implicit wait is defined.


jlipps:
this is technically a breaking change so we should call it out appropriately in the changelog

mykola-mokhnach:
Closed as violating the spec

PR #495, opened 2019-01-29
stop chromedriver proxies when reset command is called. fixes #12083

Jonahss:
The same exact command is already being called on `closeApp`: https://github.com/appium/appium-android-driver/blob/master/lib/commands/general.js#L262

A side note: should both of these commands respect the `recreateChromeDriverSessions` desired capability? Then we should call `suspendChromedriverProxy` instead of `stopChromedriverProxies` like here:
https://github.com/appium/appium-android-driver/blob/master/lib/commands/context.js#L71

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -3,6 +3,7 @@ import log from '../logger';
 import Chromedriver from 'appium-chromedriver';
 import PortFinder from 'portfinder';
 import B from 'bluebird';
+import { util } from 'appium-support';
 import { errors } from 'appium-base-driver';
 import { default as webviewHelpers,
          NATIVE_WIN, WEBVIEW_BASE, WEBVIEW_WIN, CHROMIUM_WIN } from '../webview-helpers';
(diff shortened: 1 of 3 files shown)


Jonahss:
Oh, and should I add an e2e test for this case?

Jonahss:
Fixes #495

lib/commands/general.js
```diff
@@ -228,7 +228,7 @@ commands.reset = async function () {
   await androidHelpers.resetApp(this.adb, Object.assign({}, this.opts, {fastReset: true}));
   // reset context since we don't know what kind on context we will end up after app launch.
   this.curContext = NATIVE_WIN;
-
+  await this.stopChromedriverProxies();
```
KazuCocoa: What about calling `switchContext(NATIVE_WIN)`?
As you mentioned in your side note, if we set native window, the method calls _stop_ or _suspend_ chrome driver properly..?

Jonahss: Sure. I was just following the pattern from `closeApp` in the same file. If we call `switchContect` we get the added benefit of respecting the `recreateChromeDriverSessions` desired cap.
On `closeApp` are we ok leaving chromedriver still running if somebody set this cap?

Jonahss:
@KazuCocoa just pushed another try at this.
`switchContext()` didn't work because if fails if you try to switch to the context that's already the current context. `setContext()` does this check, so called that.

`setContext` compares the context you pass in to `this.curContext`. The issue I ran into is that the unit tests don't actually call the `createSession` method which sets `curContext = this.defaultContext()`. I moved the initial setting to the defaultContext into the constructor. I thought this was OK because it's not clear if there is ever a case where `this.curContext` makes sense to be `undefined`.

lib/commands/context.js
```diff
@@ -37,7 +37,7 @@ commands.getContexts = async function () {
 };
 
 commands.setContext = async function (name) {
-  if (name === null) {
+  if (name === null || name === undefined) {
```
KazuCocoa: what about `#hasValue` by `import { util } from 'appium-support';`?
https://github.com/appium/appium-support/blob/2fcf470576f0357e3200f8014f8b1898b326d79e/lib/util.js#L14

Jonahss: ooh that's much better. I love that.

Jonahss:
@KazuCocoa PR updated with suggestion for using `hasValue()`

Jonahss:
CI tests failed, so I want to take a look before merging. Looks related.

PR #498, opened 2019-02-04
screen off before wake up to clean text field, avoid no element crash

KazuCocoa:
When I tried to pin unlock:

- some devices approve after entering pin-code, 4 digits, automatically.
    - Then, clicking element raises an exception
- pin failed if pin text field already has texts before starting to input digits
- some devices have a different enter button for pin code such as Samsung devices

To avoid the error, I did:

- Call `KEYCODE_POWER` before wake up
    - If the text field has texts, the texts will be clean after wake up
    - If the screen already off, the screen will be on. Calling wakeup after that does not affect device state
- Calling `KEYCODE_NUMPAD_ENTER` when we tap enter key
    - It can work for Samsung devices and some other devices I can test
        - The keycode equals to clicking `id/key_enter`
- Calling `KEYCODE_NUMPAD_ENTER` after checking screen lock

---

I tested:

- emulators: API 20, 21, 26
- real devices: API 22, 25, Samsung device

They worked well than before. I'm not sure this can cover all of android devices though...
(I have no mush Chinese devices, for example..)

### DIFF
--- a/lib/unlock-helpers.js
+++ b/lib/unlock-helpers.js
@@ -8,9 +8,9 @@ const PASSWORD_UNLOCK = 'password';
 const PATTERN_UNLOCK = 'pattern';
 const FINGERPRINT_UNLOCK = 'fingerprint';
 const UNLOCK_TYPES = [PIN_UNLOCK, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGERPRINT_UNLOCK];
-const KEYCODE_NUMPAD_ENTER = '66';
-const KEYCODE_POWER = '26';
-const KEYCODE_WAKEUP = '224'; // Can work over API Level 20
+const KEYCODE_NUMPAD_ENTER = 66;
+const KEYCODE_POWER = 26;
+const KEYCODE_WAKEUP = 224; // Can work over API Level 20
 const UNLOCK_WAIT_TIME = 100;
 const HIDE_KEYBOARD_WAIT_TIME = 100;
 const INPUT_KEYS_WAIT_TIME = 100;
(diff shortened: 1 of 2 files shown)


lib/unlock-helpers.js
```diff
@@ -117,10 +122,12 @@ helpers.pinUnlock = async function (adb, driver, capabilities) {
       await driver.click(util.unwrapElement(el));
     }
   }
-  const el = await driver.findElOrEls('xpath', "//*[contains(@resource-id, 'id/key_enter')]", false);
-  await driver.click(util.unwrapElement(el));
-  // Waits a bit for the device to be unlocked
+  // Some devices accept the entered key automatically
   await sleep(UNLOCK_WAIT_TIME);
```
KazuCocoa: When I rush commands without this wait before pressKeyCode, rarely UI2 sever crashed.
Thus, I leave this and call `isScreenLocked` after this.

lib/unlock-helpers.js
```diff
@@ -9,6 +9,8 @@ const PATTERN_UNLOCK = 'pattern';
 const FINGERPRINT_UNLOCK = 'fingerprint';
 const UNLOCK_TYPES = [PIN_UNLOCK, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGERPRINT_UNLOCK];
 const KEYCODE_NUMPAD_ENTER = '66';
+const KEYCODE_POWER = '26';
+const KEYCODE_WAKEUP = '224'; // Can work over API Level 20
```
mykola-mokhnach: key codes may be integers

lib/unlock-helpers.js
mykola-mokhnach: Please add this note as the comment in the code

lib/unlock-helpers.js
KazuCocoa: Either worked fine. I defined as string because of the previous implementation.
I ensured this worked, including `KEYCODE_NUMPAD_ENTER`, worked as _int_ as well.

PR #504, opened 2019-02-15
move disabling animation before autoLaunch check

KazuCocoa:
As same as https://github.com/appium/appium-uiautomator2-driver/pull/291 .

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -261,7 +261,6 @@ class AndroidDriver extends BaseDriver {
       await this.initAUT();
     }
 
-
     // start UiAutomator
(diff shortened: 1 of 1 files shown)


lib/driver.js
```diff
@@ -261,6 +255,13 @@ class AndroidDriver extends BaseDriver {
       }
     }
 
+    // If the user sets autoLaunch to false, they are responsible for initAUT() and startAUT()
+    if (this.opts.autoLaunch) {
+      // set up app under test
+      await this.initAUT();
+    }
+
+
```
mykola-mokhnach: extra empty line here

PR #510, opened 2019-02-22
add skipLogcatCapture

KazuCocoa:
Add ability to disable logcat capture 
https://github.com/appium/appium/issues/12203

cc @sshock

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -416,7 +416,9 @@ class AndroidDriver extends BaseDriver {
     }
     // some cleanup we want to do regardless, in case we are shutting down
     // mid-startup
-    await this.adb.stopLogcat();
+    if (!this.opts.skipLogcatCapture) {
+      await this.adb.stopLogcat();
+    }
     if (this.useUnlockHelperApp) {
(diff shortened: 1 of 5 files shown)


sshock:
great, thanks for adding this!

mykola-mokhnach:
@KazuCocoa Please add the new cap into the documentation

lib/driver.js
```diff
@@ -416,7 +416,9 @@ class AndroidDriver extends BaseDriver {
     }
     // some cleanup we want to do regardless, in case we are shutting down
     // mid-startup
-    await this.adb.stopLogcat();
+    if (!this.opts.skipLogcatCapture) {
```
mykola-mokhnach: I think it's safe not to check this, since `stopLogcat` simply returns silently if no logs collections has been started

KazuCocoa: Okay, right. Will simplify here
https://github.com/appium/appium-adb/blob/84d736c6b64a66721b634e69494f8ffaa47cba81/lib/tools/adb-commands.js#L1119-L1128

PR #514, opened 2019-03-13
Add mobile commands for interaction with package permissions

mykola-mokhnach:
https://github.com/appium/appium/issues/12296

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -379,7 +379,7 @@ const PERMISSION_ACTION = {
 
 /**
  * @typedef {Object} ChangePermissionsOptions
- * @property {!string|Array<string>} permissions - The full name of the permission to be granted
+ * @property {!string|Array<string>} permissions - The full name of the permission to be changed
  * or comma-separated list of permissions. Mandatory argument.
  * @property {string} appPackage [this.opts.appPackage] - The application package to set change
  * permissions on. Defaults to the package name under test.
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -377,6 +377,21 @@ const PERMISSION_ACTION = {
   REVOKE: 'revoke',
 };
 
+/**
+ * @typedef {Object} ChangePermissionsOptions
+ * @property {!string|Array<string>} permissions - The full name of the permission to be granted
+ * or comma-separated list of permissions. Mandatory argument.
+ * @property {string} appPackage [this.opts.appPackage] - The application package to set change
+ * permissions on. Defaults to the package name under test.
+ * @property {string} action [grant] - One of `PERMISSION_ACTION` values
+ */
+
+/**
+ * Changes package permissions in runtime.
+ *
+ * @param {?ChangePermissionsOptions} opts - Available options mapping.
+ * @throws {Error} if there was a failure while settings permissions
+ */
 commands.mobileChangePermissions = async function (opts = {}) {
   const {
     permissions,
@@ -411,6 +426,22 @@ const PERMISSIONS_TYPE = {
   REQUESTED: 'requested',
 };
 
+/**
+ * @typedef {Object} GetPermissionsOptions
+ * @property {string} type [requested] - One of possible permission types to get.
+ * Can be any of `PERMISSIONS_TYPE` values.
+ * @property {string} appPackage [this.opts.appPackage] - The application package to set change
+ * permissions on. Defaults to the package name under test.
+ */
+
+/**
+ * Gets runtime permissions list for the given application package.
+ *
+ * @param {GetPermissionsOptions} opts - Available options mapping.
+ * @returns {Array<string>} The list of retrieved permissions for the given type
+ * (can also be empty).
+ * @throws {Error} if there was an error while getting permissions.
+ */
 commands.mobileGetPermissions = async function (opts = {}) {
   const {
     type = PERMISSIONS_TYPE.REQUESTED,
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -380,7 +380,7 @@ const PERMISSION_ACTION = {
 /**
  * @typedef {Object} ChangePermissionsOptions
  * @property {!string|Array<string>} permissions - The full name of the permission to be changed
- * or comma-separated list of permissions. Mandatory argument.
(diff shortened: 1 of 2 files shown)


KazuCocoa:
links:
- https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/commands/general.js#L99
- https://github.com/appium/appium-espresso-driver/blob/master/lib/commands/execute.js

lib/commands/general.js
```diff
@@ -372,6 +372,101 @@ commands.mobilePerformEditorAction = async function (opts = {}) {
...
+    action = PERMISSION_ACTION.GRANT,
+  } = opts;
+  if (!util.hasValue(permissions)) {
+    log.errorAndThrow(`'permissions' argument is required`);
+  }
+
+  let actionFunc;
+  switch (_.toLower(action)) {
+    case PERMISSION_ACTION.GRANT:
+      actionFunc = this.adb.grantPermission;
```
DudaGod: @mykola-mokhnach, @KazuCocoa, @dpgraham guys why we dont use `grantPermissions`? - `https://github.com/appium/appium-adb/blob/e6bdc234e7dcf61e0252aad3ac7ab47491fe3720/lib/tools/adb-commands.js#L232` for much better performance?

We can check here if it is an Array than use `grantPermissions` if not use `grantPermission`.

What do you think?

mykola-mokhnach: it is not used, because we cannot get a reliable result if one of the permissions fails to be set

DudaGod: But we used it for set all permissions... strange position.

Okay, how can you be sure that when you call `grantPermission` one by one they are not will be failed?

What would you say if I send PR for fix it? In my project I set only two permissions and it took 2 seconds... 

Example:
```
2019-03-21T17:47:51.489924461Z 2019-03-21 17:47:51:489 - [HTTP] --> POST /wd/hub/session/5f9c50e6-9331-4681-8dcb-01a017724866/execute
2019-03-21T17:47:51.490015748Z 2019-03-21 17:47:51:489 - [HTTP] {"args":[{"action":"grant","permissions":["android.permission.ACCESS_FINE_LOCATION","android.permission.ACCESS_COARSE_LOCATION"]}],"script":"mobile:changePermissions"}
2019-03-21T17:47:51.491257929Z 2019-03-21 17:47:51:491 - [debug] [MJSONWP (5f9c50e6)] Calling AppiumDriver.execute() with args: ["mobile:changePermissions",[{"action":"grant","permissions":["android.permission.ACCESS_FINE_LOCATION","android.permission.ACCESS_COARSE_LOCATION"]}],"5f9c50e6-9331-4681-8dcb-01a017724866"]
2019-03-21T17:47:51.501615331Z 2019-03-21 17:47:51:501 - [debug] [ADB] Running '/opt/android-sdk-linux/platform-tools/adb -P 5037 -s emulator-5554 shell pm grant ru.yandex.searchplugin android.permission.ACCESS_FINE_LOCATION'
2019-03-21T17:47:51.526565336Z 2019-03-21 17:47:51:526 - [HTTP] --> POST /wd/hub/session/5f9c50e6-9331-4681-8dcb-01a017724866/context
2019-03-21T17:47:51.526625910Z 2019-03-21 17:47:51:526 - [HTTP] {"name":"NATIVE_APP"}
2019-03-21T17:47:51.531258988Z 2019-03-21 17:47:51:531 - [debug] [MJSONWP (5f9c50e6)] Calling AppiumDriver.setContext() with args: ["NATIVE_APP","5f9c50e6-9331-4681-8dcb-01a017724866"]
2019-03-21T17:47:52.890431324Z 2019-03-21 17:47:52:890 - [debug] [ADB] Running '/opt/android-sdk-linux/platform-tools/adb -P 5037 -s emulator-5554 shell pm grant ru.yandex.searchplugin android.permission.ACCESS_COARSE_LOCATION'
2019-03-21T17:47:53.475705150Z 2019-03-21 17:47:53:475 - [debug] [MJSONWP (5f9c50e6)] Responding to client with driver.execute() result: null
2019-03-21T17:47:53.476994446Z 2019-03-21 17:47:53:476 - [HTTP] <-- POST /wd/hub/session/5f9c50e6-9331-4681-8dcb-01a017724866/execute 200 1987 ms - 76
2019-03-21T17:47:53.477153301Z 2019-03-21 17:47:53:477 - [HTTP] 
2019-03-21T17:47:53.478759706Z 2019-03-21 17:47:53:478 - [debug] [MJSONWP (5f9c50e6)] Responding to client with driver.setContext() result: null
2019-03-21T17:47:53.479436167Z 2019-03-21 17:47:53:479 - [HTTP] <-- POST /wd/hub/session/5f9c50e6-9331-4681-8dcb-01a017724866/context 200 1953 ms - 76
```

mykola-mokhnach: > What would you say if I send PR for fix it? In my project I set only two permissions and it took 2 seconds...

Combining this into single shell call won't help much. There is only a visible diff if we handle 5 and more permissions at the same time. The performance of `pm grant` call is not very good in general

> Okay, how can you be sure that when you call grantPermission one by one they are not will be failed?

each call to adb shell will return different exit code, although if all calls are combined then one does not know for sure which one of them has failed

mykola-mokhnach: btw, I think by granting FINE location permission the COARCE one should be assigned automatically

DudaGod: > Combining this into single shell call won't help much. There is only a visible diff if we handle 5 and more permissions at the same time. The performance of pm grant call is not very good in general

Okay, what if i need to set six permissions tomorrow and I don't want to use `autoGrantPermissons` because it take more time?

> each call to adb shell will return different exit code, although if all calls are combined then one does not know for sure which one of them has failed

I am not sure that there is a problem. Why one of the permissions should know that another failed? For save time and show an error to developer earlier? But then after I got an error, I will fix my permissions and will not see that error again and I will still have a better performance.

> btw, I think by granting FINE location permission the COARCE one should be assigned automatically

Unfortunately not =( My test failed if I request only FINE location.

mykola-mokhnach: test the performance for different situations with your application under test and create a PR if a better solution can be applied. We will be happy to review it.

PR #519, opened 2019-03-18
add ability for users to send in additional chromedriver args

jlipps:
So that folks can send in cli flags to chromedriver.

Also, noticed that unit tests are broken on master for me, something to do with unlock specs?

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -277,6 +277,7 @@ helpers.setupNewChromedriver = async function setupNewChromedriver (opts, curDev
     port: opts.chromeDriverPort,
     executable: opts.chromedriverExecutable,
     adb,
+    cmdArgs: opts.chromedriverArgs,
     verbose: !!opts.showChromedriverLog,
     executableDir: opts.chromedriverExecutableDir,
     mappingPath: opts.chromedriverChromeMappingFile,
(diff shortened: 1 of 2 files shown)


lib/commands/context.js
```diff
@@ -277,6 +277,7 @@ helpers.setupNewChromedriver = async function setupNewChromedriver (opts, curDev
     port: opts.chromeDriverPort,
     executable: opts.chromedriverExecutable,
     adb,
+    cmdArgs: opts.chromedriverArgs,
```
KazuCocoa: Do we need to update appium-chromedriver in package.json to accept this arg?

jlipps: @KazuCocoa appium-chromedriver already accepted it, we just had never used it before!

jlipps:
rebased on master to hopefully get CI passing due to unit test fixes there.

lib/desired-caps.js
```diff
@@ -76,6 +76,9 @@ let commonCapConstraints = {
   chromeDriverPort: {
     isNumber: true
   },
+  chromedriverArgs: {
```
mykola-mokhnach: it would be nice to have the cap documented elsewhere

imurchie: Ugh. **One** cap has snake case in "chromedriver"!

jlipps: yeah :-(

jlipps:
will add documentation in caps.md momentarily.

jlipps:
published in 4.11.0

PR #526, opened 2019-03-29
Make sure geolocation values are cast to float

mykola-mokhnach:
I was surprised, but in JS `parseFloat('0.0') === 0`. And Java requires these values to be of type [double](https://discuss.appium.io/t/appium-1-9-1-issues-setting-location-on-android-device/23955/36). @imurchie Do you have any better ideas of how the cast to float zero can be implemented?

### DIFF
--- a/lib/commands/network.js
+++ b/lib/commands/network.js
@@ -12,7 +12,8 @@ const DATA_MASK = 0b100;
 // to trick JSON generation and send a float value instead of an integer,
 // This allows strictly-typed clients, like Java, to properly
 // parse it. Otherwise float 0.0 is always represented as integer 0 in JS.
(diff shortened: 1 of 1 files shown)


imurchie:
No good way, no. And I'm surprised this works, too. I would expect them to come out as `0`, still. It might be that the Java client needs to handle ints.

mykola-mokhnach:
@imurchie I've changed zero to 1e-15. I hope this would be a good approximation

imurchie:
If you want a _very small number_, can you not use a constant like `Number.MIN_VALUE`?

mykola-mokhnach:
Changed to `Number.MIN_VALUE`

PR #527, opened 2019-03-31
Handle `mobile:` calls in non-native context

mykola-mokhnach:
Partially addresses https://github.com/appium/appium/issues/12407

It makes sense to always assume that scripts starting with `mobile:` magic always target the native context even if the current one is the web one.


dpgraham:
Agreed. Could this be put in base driver?

mykola-mokhnach:
> Agreed. Could this be put in base driver?

I don't think so, since the base driver should have abstractions over context drivers for all platforms in order to make this working.

PR #528, opened 2019-04-02
allow chrome automation even when AUT also has a webview

jlipps:
There was a bug, in the following situation:

1. User opens up AUT which has a webview
2. AUT triggers Chrome to open
3. User gets list of contexts. Now the context list correctly says [AUT, chrome]
4. User selects chrome context ("WEBVIEW_chrome")
5. User thinks they are automating the chrome context, but actually they're still automating the AUT webview (title, source, etc..., all come from the AUT webview)

This was because, whenever we set up chromedriver, we always used the AUT's android package for chromedriver to attach to.

This PR special cases the chrome webview so that we can correctly attach to its android package from chromedriver. It also makes it so this works even when the user isn't using the (extremely long) `extractChromeAndroidPackageFromContextName` capability.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -115,7 +115,13 @@ helpers.startChromedriverProxy = async function startChromedriverProxy (context)
   } else {
     let opts = _.cloneDeep(this.opts);
     opts.chromeUseRunningApp = true;
-    if (opts.extractChromeAndroidPackageFromContextName) {
+
+    // if requested, tell chromedriver to attach to the android package we have
+    // associated with the context name, rather than the package of the AUT.
+    // And turn this on by default for chrome--if chrome pops up with a webview
+    // and someone wants to switch to it, we should let chromedriver connect to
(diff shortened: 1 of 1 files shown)


lib/commands/context.js
```diff
@@ -333,7 +339,16 @@ helpers.setupNewChromedriver = async function setupNewChromedriver (opts, curDev
   if (opts.pageLoadStrategy) {
     caps.pageLoadStrategy = opts.pageLoadStrategy;
   }
+  if (caps.chromeOptions.androidPackage === 'chrome') {
```
KazuCocoa: Is it safer to compare as case insensitive?

jlipps: yeah, i was not sure where `WEBVIEW_chrome` comes from and if we can rely on it always being the same. I was assuming we could, but maybe it's a bad assumption.

jlipps: decided to make it case insensitive

PR #529, opened 2019-04-03
add chromedriverPorts cap which allows specifying multiple ports or a range

jlipps:
this helps support the multiple webview case, and would provide a supported 'workaround' for e.g. https://github.com/appium/appium/issues/11820

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -271,16 +271,64 @@ async function setupExistingChromedriver (chromedriver) {
   return chromedriver;
 }
 
+async function getChromedriverPort (opts) {
+  const getPort = B.promisify(PortFinder.getPort, {context: PortFinder});
+
+  // if the user didn't give us any specific information about chromedriver
+  // port ranges, just find any free port
+  if (!opts.chromedriverPorts) {
+    const port = await getPort();
+    log.debug(`A port was not given, using random free port: ${port}`);
+    return port;
+  }
+
+  // otherwise find the free port based on a list or range provided by the user
+  // which could look like [8000, 8001, [9000, 9500]], where each element
+  // in the array is either a single port or an array with two elements
+  // defining a port range
+  log.debug(`Finding a free port for chromedriver using spec ${JSON.stringify(opts.chromedriverPorts)}`);
+  let foundPort = null;
+  for (let potentialPort of opts.chromedriverPorts) {
+    let port, stopPort;
+    if (_.isArray(potentialPort)) {
+      ([port, stopPort] = potentialPort);
+    } else {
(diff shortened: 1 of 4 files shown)


jlipps:
oh, also deprecate `chromeDriverPort` in favor of the more standard `chromedriverPort`, for parity with all our other chromedriver caps

jlipps:
while i was in there i also removed a discrepancy where the bootstrapPort opt was only a CLI opt and not a desired cap. now it can be either like all the rest.

lib/commands/context.js
```diff
@@ -265,16 +265,64 @@ async function setupExistingChromedriver (chromedriver) {
...
+    return port;
+  }
+
+  // otherwise find the free port based on a list or range provided by the user
+  // which could look like [8000, 8001, [9000, 9500]], where each element
+  // in the array is either a single port or an array with two elements
+  // defining a port range
+  log.debug(`Finding a free port for chromedriver using spec ${JSON.stringify(opts.chromedriverPorts)}`);
+  let foundPort = null;
+  for (let potentialPort of opts.chromedriverPorts) {
```
KazuCocoa: const?

lib/commands/context.js
```diff
@@ -265,16 +265,64 @@ async function setupExistingChromedriver (chromedriver) {
   return chromedriver;
 }
 
+async function getChromedriverPort (opts) {
```
KazuCocoa: docstring?

lib/commands/context.js
```diff
@@ -265,16 +265,64 @@ async function setupExistingChromedriver (chromedriver) {
...
+}
+
 helpers.setupNewChromedriver = async function setupNewChromedriver (opts, curDeviceId, adb) {
-  // if a port wasn't given, pick a random available one
-  if (!opts.chromeDriverPort) {
-    const getPort = B.promisify(PortFinder.getPort, {context: PortFinder});
-    opts.chromeDriverPort = await getPort();
-    log.debug(`A port was not given, using random port: ${opts.chromeDriverPort}`);
+  if (opts.chromeDriverPort) {
+    log.warn(`The 'chromeDriverPort' capability is deprecated. Please use 'chromedriverPort' instead`);
```
KazuCocoa: 👍

jlipps:
ok, feedback addressed. this PR now also depends on https://github.com/appium/appium-base-driver/pull/314

KazuCocoa:
a note need to update after this:

- https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/caps.md

jlipps:
published in 4.14.0

PR #538, opened 2019-05-03
call tempDir from appium-support to respect --temp

KazuCocoa:
related to https://github.com/appium/appium-support/pull/112
To respect `--tmp` arg by the PR, `temp.path` should be `tempDir.path`.

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -179,7 +179,7 @@ commands.pullFile = async function pullFile (remotePath) {
                         `Original error: ${e.message}`);
     }
   }
-  const localFile = tempDir.path({prefix: 'appium', suffix: '.tmp'});
+  const localFile = await tempDir.path({prefix: 'appium', suffix: '.tmp'});
   try {
     await this.adb.pull(_.isString(tmpDestination) ? tmpDestination : remotePath, localFile);
     const data = await fs.readFile(localFile);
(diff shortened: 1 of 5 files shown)


imurchie:
Does this mean we can get rid of the [temp](https://www.npmjs.com/package/temp) dependency?

KazuCocoa:
ah, yeah.
They generate like `/var/folders/34/2222sh8n27d6rcp7jqlkw8km0000gn/T/appium201944-69102-55h9j9.beo9w.tmp`. `/var/folders/34/2222sh8n27d6rcp7jqlkw8km0000gn/T` are by `os.tmpdir()` (macOS case). The prefix, suffix and string combination is same as our appium-support lib's the same name method, `path`. I think we can remove the dependency.

I grepped this repo and found e2e test used it.
I'll replace them as well in this PR

lib/commands/actions.js
```diff
@@ -180,7 +179,7 @@ commands.pullFile = async function pullFile (remotePath) {
                         `Original error: ${e.message}`);
     }
   }
-  const localFile = temp.path({prefix: 'appium', suffix: '.tmp'});
+  const localFile = tempDir.path({prefix: 'appium', suffix: '.tmp'});
```
mykola-mokhnach: await is missing

PR #542, opened 2019-05-13
Use existing chrome if noReset is true

imurchie:
If `noReset` is set to `true`, use an existing Chrome. Without this a temporary profile is used for each session, so everything is new.

If `noReset` is `false` or not defined, it sill stay as `false`.


clample:
@imurchie I tested this locally with `noReset` set to `true` - neither Appium nor Chromedriver are starting Chrome now. If you don't manually open Chrome then the test fails since Chromedriver times out attempting to connect to Chrome.
[chrome-not-launching.txt](https://github.com/appium/appium-android-driver/files/3176438/chrome-not-launching.txt)

I think it would be more user friendly if Appium starts Chrome in this case.

PR #546, opened 2019-06-14
Adding new capability to specify the chromedriver url 

shridharkalagi:
Part of this [PR](https://github.com/appium/appium/pull/12777). Fixes issue [https://github.com/appium/appium/issues/10651](https://github.com/appium/appium/issues/10651)

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -3,13 +3,18 @@ import log from '../logger';
 import Chromedriver from 'appium-chromedriver';
 import PortFinder from 'portfinder';
 import B from 'bluebird';
-import { util } from 'appium-support';
-import { errors } from 'appium-base-driver';
-import { default as webviewHelpers,
-         NATIVE_WIN, WEBVIEW_BASE, WEBVIEW_WIN, CHROMIUM_WIN } from '../webview-helpers';
+import path from 'path';
+import { util, zip, tempDir } from 'appium-support';
+import { errors, downloadApp } from 'appium-base-driver';
+import {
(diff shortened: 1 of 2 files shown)


shridharkalagi:
OK I will update the PR to put this behind `--relaxed-security` flag

saikrishna321:
FYI https://github.com/appium/appium/pull/12778

mykola-mokhnach:
I don't like the idea at all from security perspective. We could think about providing a driver version, so then the script could get it from the official google drivers repository. But downloading and executing random executables on the server side - not today.

umutuzgur:
This is really dangerous. This is not even relaxed security but no security. We should just make Appium update the chromedriver upon server boot up. Maybe we can pass a server flag to sync chromedrivers and we download them from googles repo

jlipps:
@shridharkalagi thanks for this change, and it was fun to help you with it at the appiumconf workshop. after hearing the feedback from the other maintainers, i agree that it's an unreasonable security risk to include this as a feature. my apologies for not seeing the security implications earlier!

so far i like @mykola-mokhnach's idea of including a cap that would download a new (known good) version of chromedriver if it's not the one that appium downloaded on install. that way the version would be modulated by session rather than by server. cloud services could download all the versions to have them available if necessary. it would take some thinking how to do this the right way but all the pieces are already there. (maybe the `--chromedriver-version` install flag could be modified to `--chromedriver-versions`, and then the client can pick any chromedriver from the list that is already installed on the server?)

PR #548, opened 2019-06-30
Add server feature to enable automated Chromedriver downloads

mykola-mokhnach:
Based on https://github.com/appium/appium-chromedriver/pull/131

Please do not merge until the above PR is published

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -333,15 +333,12 @@ async function getChromedriverPort (portSpec) {
 }
 
 helpers.isChromedriverAutodownloadEnabled = function isChromedriverAutodownloadEnabled () {
-  try {
-    this.ensureFeatureEnabled(CHROMEDRIVER_AUTODOWNLOAD_FEATURE);
-    log.debug('Automated Chromedriver download is enabled');
+  if (this.isFeatureEnabled(CHROMEDRIVER_AUTODOWNLOAD_FEATURE)) {
     return true;
-  } catch (e) {
-    log.debug(`Automated Chromedriver download is disabled. ` +
-      `Use '${CHROMEDRIVER_AUTODOWNLOAD_FEATURE}' server feature to enable it`);
-    return false;
   }
+  log.debug(`Automated Chromedriver download is disabled. ` +
+    `Use '${CHROMEDRIVER_AUTODOWNLOAD_FEATURE}' server feature to enable it`);
+  return false;
 };
 
 helpers.setupNewChromedriver = async function setupNewChromedriver (opts, curDeviceId, adb) {
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -368,7 +368,8 @@ helpers.setupNewChromedriver = async function setupNewChromedriver (opts, curDev
     bundleId: opts.chromeBundleId,
     useSystemExecutable: opts.chromedriverUseSystemExecutable,
     disableBuildCheck: opts.chromedriverDisableBuildCheck,
-    isAutodownloadEnabled: this.isChromedriverAutodownloadEnabled(),
+    isAutodownloadEnabled: (this || {}).isChromedriverAutodownloadEnabled
+      ? this.isChromedriverAutodownloadEnabled() : undefined,
   });
 
   // make sure there are chromeOptions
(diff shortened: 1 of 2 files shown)


lib/commands/context.js
```diff
@@ -353,7 +356,18 @@ helpers.setupNewChromedriver = async function setupNewChromedriver (opts, curDev
     bundleId: opts.chromeBundleId,
     useSystemExecutable: opts.chromedriverUseSystemExecutable,
     disableBuildCheck: opts.chromedriverDisableBuildCheck,
-  });
+  };
+
+  try {
+    this.ensureFeatureEnabled(CHROMEDRIVER_AUTODOWNLOAD_FEATURE);
```
umutuzgur: Could we just turn this into a helper method. Like

```javascript
isChromedriverAutoDownloadEnabled() {
try {
  this.ensureFeatureEnabled(CHROMEDRIVER_AUTODOWNLOAD_FEATURE);
  return true;
} catch (e) {
  log.debug(`Automated Chromedriver download is disabled. ` +
    `Use '${CHROMEDRIVER_AUTODOWNLOAD_FEATURE}' server feature to enable it`);
  return false;
}
}

Comment (umutuzgur) [reply lib/commands/context.js]: Then we can keep the old style the way it is and create the `Chromedriver` like the way it was before

```javascript
const chromedriver = new Chromedriver({	  const chromedriverOpts = {
    port: opts.chromedriverPort,	    port: opts.chromedriverPort,
    executable: opts.chromedriverExecutable,	    executable: opts.chromedriverExecutable,
    adb,	    adb,
    cmdArgs: opts.chromedriverArgs,	    cmdArgs: opts.chromedriverArgs,
    verbose: !!opts.showChromedriverLog,	    verbose: !!opts.showChromedriverLog,
    executableDir: opts.chromedriverExecutableDir,	    executableDir: opts.chromedriverExecutableDir,
    mappingPath: opts.chromedriverChromeMappingFile,	    mappingPath: opts.chromedriverChromeMappingFile,
    bundleId: opts.chromeBundleId,	    bundleId: opts.chromeBundleId,
    useSystemExecutable: opts.chromedriverUseSystemExecutable,	    useSystemExecutable: opts.chromedriverUseSystemExecutable,
    disableBuildCheck: opts.chromedriverDisableBuildCheck,	    disableBuildCheck: opts.chromedriverDisableBuildCheck,
    isAutodownloadEnabled : this. isChromedriverAutoDownloadEnabled()
  });
```

mykola-mokhnach: yes, easy

lib/commands/context.js
```diff
@@ -329,6 +332,18 @@ async function getChromedriverPort (portSpec) {
   return foundPort;
 }
 
+helpers.isChromedriverAutodownloadEnabled = function isChromedriverAutodownloadEnabled () {
```
umutuzgur: I think it should be `isChromedriverAutoDownloadEnabled` but it doesn't matter. I just wanted to comment in case it was a typo

lib/commands/context.js
```diff
@@ -353,6 +368,7 @@ helpers.setupNewChromedriver = async function setupNewChromedriver (opts, curDev
     bundleId: opts.chromeBundleId,
     useSystemExecutable: opts.chromedriverUseSystemExecutable,
     disableBuildCheck: opts.chromedriverDisableBuildCheck,
+    isAutodownloadEnabled: this.isChromedriverAutodownloadEnabled(),
```
umutuzgur: I actually have a concern now. How should this feature work when `mappingPath` and `executableDir` are set? Should we set `isAutodownloadEnabled` to `false` if they are passed?

mykola-mokhnach: they can work together: https://github.com/appium/appium-chromedriver/pull/131/files#diff-df03902a1655a73e18bdf5877e4a8de4R274

umutuzgur: Gotcha. Thanks for explaining

lib/commands/context.js
```diff
@@ -329,6 +332,18 @@ async function getChromedriverPort (portSpec) {
   return foundPort;
 }
 
+helpers.isChromedriverAutodownloadEnabled = function isChromedriverAutodownloadEnabled () {
+  try {
+    this.ensureFeatureEnabled(CHROMEDRIVER_AUTODOWNLOAD_FEATURE);
```
jlipps: you can use `isFeatureEnabled` if you just want the boolean value instead of using try/catch...

PR #553, opened 2019-07-10
feat: add 'ensureWebviewsHavePages' cap

jlipps:
This PR fixes an odd situation which I have observed with Chrome-backed webviews on Android. Currently, Appium's code assumes that if it can find a webview-type process in `/proc/net/unix`, then a webview exists for that process.

I've noticed that this assumption is not always true. Sometimes, the process exists, and a Chrome remote debugger is even running for the process, but there are no active webview pages. In this case, Chromedriver will always fail to attach to the webview, resulting in an error like "unable to discover open pages".

This PR adds a new capability `ensureWebviewsHavePages` which, if true, attempts to verify that a webview actually has active pages before adding it to the list of contexts. This serves to prevent users from trying to connect to a webview which is guaranteed to fail, and which will result in a cryptic error message.

I've added this behind a capability because the only way I can see to achieve this is to send a request to one of the [Chrome devtools HTTP endpoints](https://chromedevtools.github.io/devtools-protocol/). This involves running an adb command to forward a local port to the remote debugger port on the device, so we can make the necessary HTTP request, and then cleaning up the forwarded port afterwards. Thus it potentially adds overhead of time, and I'm also not certain that all webviews on all supported versions of Android support the Chrome devtools HTTP server. Once this feature is well-tested and proved useful, we could make this behavior opt-out rather than opt-in.

Because a local port is required for this feature, I've also added another capability, `webviewDevtoolsPort`, so that clients can specify a port to avoid port conflicts between sessions.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -33,7 +33,8 @@ commands.getContexts = async function getContexts () {
   } else {
     // otherwise we use ADB to figure out which webviews are available
     webviews = await webviewHelpers.getWebviews(this.adb,
-      this.opts.androidDeviceSocket);
+      this.opts.androidDeviceSocket, this.opts.ensureWebviewsHavePages,
+      this.opts.webviewDevtoolsPort);
   }
   this.contexts = _.union([NATIVE_WIN], webviews);
   log.debug(`Available contexts: ${JSON.stringify(this.contexts)}`);
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -32,9 +32,7 @@ commands.getContexts = async function getContexts () {
     webviews = [CHROMIUM_WIN];
   } else {
     // otherwise we use ADB to figure out which webviews are available
-    webviews = await webviewHelpers.getWebviews(this.adb,
-      this.opts.androidDeviceSocket, this.opts.ensureWebviewsHavePages,
-      this.opts.webviewDevtoolsPort);
+    webviews = await webviewHelpers.getWebviews(this.adb, this.opts);
   }
   this.contexts = _.union([NATIVE_WIN], webviews);
   log.debug(`Available contexts: ${JSON.stringify(this.contexts)}`);
--- a/lib/webview-helpers.js
+++ b/lib/webview-helpers.js
@@ -24,7 +24,7 @@ let helpers = {};
  *
  * @param {object} adb - an ADB instance
  *
- * @return {Array} - a list of webview process names (including the leading
+ * @return {Array.<string>} - a list of webview process names (including the leading
  * '@')
  */
 async function getPotentialWebviewProcs (adb) {
@@ -43,6 +43,12 @@ async function getPotentialWebviewProcs (adb) {
 }
 
 /**
+ * @typedef {Object} WebviewProc
+ * @property {string} proc - The webview process name (as returned by
+ * getPotentialWebviewProcs
+ * @property {string} webview - The actual webview context name
+ */
+/**
  * This function retrieves a list of system processes that look like webviews,
  * and returns them along with the webview context name appropriate for it.
  * If we pass in a deviceSocket, we only attempt to find webviews which match
@@ -52,8 +58,7 @@ async function getPotentialWebviewProcs (adb) {
  * @param {object} adb - an ADB instance
  * @param {string} deviceSocket - the explictly-named device socket to use
  *
- * @return {Array} - a list of objects of the form
- * {proc: <proc-name>, webview: <webview-name>}
+ * @return {Array.<WebviewProc>}
  */
 // TODO: some of this function probably belongs in appium-adb?
 async function webviewsFromProcs (adb, deviceSocket) {
@@ -92,8 +97,7 @@ async function webviewsFromProcs (adb, deviceSocket) {
  * pages, because such processes can't be automated by chromedriver anyway.
  *
  * @param {object} adb - an ADB instance
- * @param {object} wp - a 'webview-proc' object (as returned by
- * webviewsFromProcs), containing a 'proc' and 'webview' field.
+ * @param {WebviewProc} wp - the webview to check
  * @param {number} webviewDevtoolsPort - the local port to use for the check
  *
  * @return {boolean} - whether or not the webview has pages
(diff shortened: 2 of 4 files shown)


mykola-mokhnach:
Thanks @jlipps for looking into this.
Linking https://github.com/appium/appium/issues/11999

lib/commands/context.js
```diff
@@ -33,7 +33,8 @@ commands.getContexts = async function getContexts () {
   } else {
     // otherwise we use ADB to figure out which webviews are available
     webviews = await webviewHelpers.getWebviews(this.adb,
-      this.opts.androidDeviceSocket);
+      this.opts.androidDeviceSocket, this.opts.ensureWebviewsHavePages,
```
mykola-mokhnach: I'd rather put the new options into a dictionary

lib/webview-helpers.js
```diff
@@ -12,48 +13,97 @@ const CROSSWALK_SOCKET_SUFFIX = '_devtools_remote';
 const CROSSWALK_REGEXP_STRING = `(\\S*)${CROSSWALK_SOCKET_SUFFIX}`;
 const CROSSWALK_REGEXP = new RegExp(`@${CROSSWALK_REGEXP_STRING}`);
 const CROSSWALK_PROCESS_REGEXP = new RegExp(WEBVIEW_BASE + CROSSWALK_REGEXP_STRING);
+const DEFAULT_WEBVIEW_DEVTOOLS_PORT = 9222;
 
 
 let helpers = {};
 
 // This function gets a list of android system processes and returns ones
```
mykola-mokhnach: Can we instead add a proper docstring?

mykola-mokhnach: same below

lib/webview-helpers.js
```diff
@@ -12,48 +13,97 @@ const CROSSWALK_SOCKET_SUFFIX = '_devtools_remote';
...
+// webview-looking processes which have a remote debug port open but which
+// aren't actually running any pages, because such processes can't be automated
+// by chromedriver anyway.
+async function webviewHasPages (adb, wp, webviewDevtoolsPort) {
+  let hasPages = false;
+  const wvPort = webviewDevtoolsPort || DEFAULT_WEBVIEW_DEVTOOLS_PORT;
+
+  // proc names come with '@', but this should not be a part of the abstract
+  // remote port, so remove it
+  const remotePort = wp.proc.replace('@', '');
```
mykola-mokhnach: should we replace only `/^@/` match if the string starts with it or all matches must be replaced?

lib/webview-helpers.js
```diff
@@ -12,48 +13,97 @@ const CROSSWALK_SOCKET_SUFFIX = '_devtools_remote';
...
-  return _.uniq(webviews);
+  return webviews;
+}
+
+// given a 'webview-proc' object, check whether we can detect any active pages
+// corresponding to that webview. This is used by getWebviews to filter out any
+// webview-looking processes which have a remote debug port open but which
+// aren't actually running any pages, because such processes can't be automated
+// by chromedriver anyway.
+async function webviewHasPages (adb, wp, webviewDevtoolsPort) {
```
mykola-mokhnach: `wp` could be destructed rather directly here or in the function body to enhance readability

lib/webview-helpers.js
```diff
@@ -12,48 +13,97 @@ const CROSSWALK_SOCKET_SUFFIX = '_devtools_remote';
...
+
+  // proc names come with '@', but this should not be a part of the abstract
+  // remote port, so remove it
+  const remotePort = wp.proc.replace('@', '');
+
+  // forward the specified local port to the remote debugger port on the device
+  await adb.adbExec(['forward', `tcp:${wvPort}`, `localabstract:${remotePort}`]);
+  try {
+    const remoteDebugger = `http://localhost:${wvPort}/json/list`;
+    logger.info(`Attempting to get list of pages for webview '${wp.webview}' ` +
```
mykola-mokhnach: debug

lib/webview-helpers.js
```diff
@@ -12,48 +13,97 @@ const CROSSWALK_SOCKET_SUFFIX = '_devtools_remote';
...
+async function webviewHasPages (adb, wp, webviewDevtoolsPort) {
+  let hasPages = false;
+  const wvPort = webviewDevtoolsPort || DEFAULT_WEBVIEW_DEVTOOLS_PORT;
+
+  // proc names come with '@', but this should not be a part of the abstract
+  // remote port, so remove it
+  const remotePort = wp.proc.replace('@', '');
+
+  // forward the specified local port to the remote debugger port on the device
+  await adb.adbExec(['forward', `tcp:${wvPort}`, `localabstract:${remotePort}`]);
```
mykola-mokhnach: wouldn't this command fail if the port is already busy?

lib/webview-helpers.js
```diff
@@ -112,10 +162,34 @@ helpers.procFromWebview = async function procFromWebview (adb, webview) {
...
+  ensureWebviewsHavePages, webviewDevtoolsPort) {
+
   logger.debug('Getting a list of available webviews');
-  let webviews = await webviewsFromProcs(adb, deviceSocket);
+  let webviewProcs = await webviewsFromProcs(adb, deviceSocket);
+
+  if (ensureWebviewsHavePages) {
+    logger.info('Retrieved potential webviews; will filter out ones with no active pages');
+    webviewProcs = await asyncfilter(webviewProcs, async (wp) => {
+      return await webviewHasPages(adb, wp, webviewDevtoolsPort);
```
mykola-mokhnach: return is redundant

lib/webview-helpers.js
```diff
@@ -112,10 +162,34 @@ helpers.procFromWebview = async function procFromWebview (adb, webview) {
...
+    }, false /*ensure serial operation*/);
+  } else {
+    logger.info('Not checking whether webviews have active pages; use the ' +
+                "'ensureWebviewsHavePages' cap to turn this check on");
+  }
+
+  // webviewProcs contains procs, which we only care about for ensuring
+  // presence of pages above, so we can now discard them and rely on just the
+  // webview names
+  let webviews = webviewProcs.map((wp) => wp.webview);
```
mykola-mokhnach: const

lib/webview-helpers.js
jlipps: i don't think so, doesn't the filter method need to return true or false in order to do its job?

lib/webview-helpers.js
jlipps: this is reassigned below.

jlipps:
ok @mykola-mokhnach i endured the tedium of updating all the function comments to proper docstrings :-) and found an excuse to use `reduce` 🎉

lib/webview-helpers.js
mykola-mokhnach: I mean 

```
await asyncfilter(webviewProcs, async (wp) => await webviewHasPages(adb, wp, webviewDevtoolsPort))
```

should work just fine

mykola-mokhnach:
it would be useful to have the new cap documented after the PR is merged

jlipps:
published in 4.18.0. will put in docs PR soon.

PR #555, opened 2019-07-27
Add sensors support

vrunoa:
Prepare Android driver to support sensors set for Emulators implemented on ADB
https://github.com/appium/appium-adb/pull/456


lib/commands/actions.js
```diff
@@ -312,6 +312,13 @@ commands.networkSpeed = async function networkSpeed (networkSpeed) {
   await this.adb.networkSpeed(networkSpeed);
 };
 
+commands.sensorSet = async function sensorSet (sensorType, value) {
```
KazuCocoa: Can you add the docstring?

test/unit/commands/actions-specs.js
```diff
@@ -352,6 +352,22 @@ describe('Actions', function () {
...
   });
+  describe('sensorSet', function () {
+    it('should call sensor adb command for emulator', async function () {
+      sandbox.stub(driver.adb, 'sensorSet');
+      sandbox.stub(driver, 'isEmulator').returns(true);
+      await driver.sensorSet('light', 0);
+      driver.adb.sensorSet.calledWithExactly('light', 0)
+        .should.be.true;
+    });
+    it('should throw exception for real device', async function () {
```
KazuCocoa: 👍

lib/commands/actions.js
```diff
@@ -312,6 +312,20 @@ commands.networkSpeed = async function networkSpeed (networkSpeed) {
...
 };
 
+/**
+ * Emulate sensors values on the connected emulator.
+ *
+ * @param {string} sensorType - Sensor type declared in adb.SENSORS items.
+ * @param {number|string} value  - Number to set as the sensor value.
+ * @throws {Error} - If deviceType is not an emulator
+ */
+commands.sensorSet = async function sensorSet (sensorType, value) {
```
mykola-mokhnach: which external API endpoint is supposed to trigger this method?

vrunoa: @mykola-mokhnach this is something I want to discuss with the community. We have two options here; 
1. using one endpoint per type of sensor. For ex; W3C already has a method for light sensor in progress. https://www.w3.org/TR/ambient-light/
2. using one endpoint for all sensors. 
I'd like to hear you guys feedback. I'll probably start the conversation later today in slack

mykola-mokhnach: I would say it is always safe to have a "mobile:" endpoint for such purpose

vrunoa: Added a new mapping here: https://github.com/appium/appium-android-driver/pull/555/files#diff-2ac213fe13b12323b4767dd0bd037ee4R37

lib/commands/actions.js
```diff
@@ -312,6 +312,28 @@ commands.networkSpeed = async function networkSpeed (networkSpeed) {
   await this.adb.networkSpeed(networkSpeed);
 };
 
+/**
+ * Emulate sensors values on the connected emulator.
+ *
+ * @param {object} opts - Object containing sensor type declared in adb.SENSORS items and value
```
mykola-mokhnach: Please use appropriate format to describe this type. Check https://github.com/appium/appium-adb/blob/c00e48e9977f024ca3280e1ecf6b823997594779/lib/tools/system-calls.js#L141 as an example

KazuCocoa:
We should require `^7.9.0` in https://github.com/appium/appium-android-driver/blob/master/package.json#L39 for sensor

mykola-mokhnach:
LGTM. Do not forget to add the same mobile command to uia2 and espresso after this PR is published

PR #557, opened 2019-08-05
Fix time zone formatting for device time getter

mykola-mokhnach:
By default moment.js uses the local time zone. And we need to enforce the device time zone to be used while formatting the resulting stamp.

Addresses https://discuss.appium.io/t/driver-getdevicetime-works-incorrectly-for-appium-1-7-2/27369/3

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -34,22 +34,16 @@ commands.doSendKeys = async function doSendKeys (params) {
  *                          `YYYY-MM-DDTHH:mm:ssZ`, which complies to ISO-8601
  * @returns Formatted datetime string or the raw command output if formatting fails
  */
-const DEFAULT_DEVICE_TIME = 'YYYY-MM-DDTHH:mm:ssZ';
-commands.getDeviceTime = async function getDeviceTime (format = DEFAULT_DEVICE_TIME) {
-  log.info('Attempting to capture android device date and time');
-
-  if (!_.isString(format)) {
-    log.errorAndThrow(`The format specifier is expected to be a valid string specifier like '${DEFAULT_DEVICE_TIME}'. ` +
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -34,22 +34,17 @@ commands.doSendKeys = async function doSendKeys (params) {
...
-        `The current value is: ${format}`);
-  }
-
+commands.getDeviceTime = async function getDeviceTime (format = 'YYYY-MM-DDTHH:mm:ssZ') {
+  log.debug('Attempting to capture android device date and time. ' +
+    `The format specifier is '${format}'`);
   const deviceTimestamp = (await this.adb.shell(['date', '+%Y-%m-%dT%T%z'])).trim();
-  const parsedTimestamp = moment(deviceTimestamp, 'YYYY-MM-DDTHH:mm:ssZ');
+  log.debug(`Got device timestamp: ${deviceTimestamp}`);
+  const parsedTimestamp = moment.utc(deviceTimestamp, 'YYYY-MM-DDTHH:mm:ssZZ');
```
KazuCocoa: (kind of question)
Can we apply only one `Z` insteaf of `ZZ` in JS?
```
> m.utc(a, 'YYYY-MM-DDTHH:mm:ssZ').toString()
'Sat Jun 09 2018 07:21:54 GMT+0000'
> m.utc(a, 'YYYY-MM-DDTHH:mm:ssZZ').toString()
'Sat Jun 09 2018 07:21:54 GMT+0000'
```

mykola-mokhnach: ZZ means time zone in format `+0000` and Z equals to `+00:00`. Although, I assume moment parser is smart enough to properly parse the stamp even in both cases

PR #586, opened 2019-12-10
feat: Add a possibility to stream the device's screen using adb+gstreamer

mykola-mokhnach:
This feature does the same trick as we do in iOS, although it requires GStreamer to be installed on the host machine. As the result, it is possible now to broadcast the device screen to any MJPEG client (including Chrome browser), which might be extremely useful for further integration with Appium Desktop

### DIFF
--- a/lib/commands/streamscreen.js
+++ b/lib/commands/streamscreen.js
@@ -43,7 +43,7 @@ async function verifyStreamingRequirements (adb) {
     try {
       await fs.which(binaryName);
     } catch (e) {
-      throw new Error(`The ${binaryName} is not available in PATH on the host system. ` +
+      throw new Error(`The '${binaryName}' binary is not available in the PATH on the host system. ` +
         `See ${GST_TUTORIAL_URL} for more details on how to install it.`);
     }
   }
@@ -60,23 +60,19 @@ async function verifyStreamingRequirements (adb) {
 
 async function getDeviceInfo (adb) {
   const output = await adb.shell(['dumpsys', 'display']);
-  const widthMatch = /\bdeviceWidth=(\d+)/.exec(output);
-  if (!widthMatch) {
-    throw new Error(`Cannot parse the device width from ${output}`);
+  const result = {};
+  for (const [key, pattern] of [
+    ['width', /\bdeviceWidth=(\d+)/],
+    ['height', /\bdeviceHeight=(\d+)/],
+    ['fps', /\bfps=(\d+)/],
+  ]) {
+    const match = pattern.exec(output);
+    if (!match) {
+      throw new Error(`Cannot parse the device ${key} from the command output: ${output}`);
+    }
+    result[key] = parseInt(match[1], 10);
   }
-  const heightMatch = /\bdeviceHeight=(\d+)/.exec(output);
-  if (!heightMatch) {
-    throw new Error(`Cannot parse the device height from ${output}`);
-  }
-  const fpsMatch = /\bfps=(\d+)/.exec(output);
-  if (!fpsMatch) {
-    throw new Error(`Cannot parse the device FPS from ${output}`);
-  }
-  return {
-    width: parseInt(widthMatch[1], 10),
-    height: parseInt(heightMatch[1], 10),
-    fps: parseInt(fpsMatch[1], 10),
-  };
+  return result;
 }
 
 async function initDeviceStreamingProc (adb, deviceInfo, opts = {}) {
--- a/lib/commands/streamscreen.js
+++ b/lib/commands/streamscreen.js
@@ -20,7 +20,7 @@ const REQUIRED_GST_PLUGINS = {
   h264parse: 'gst-plugins-bad',
   jpegenc: 'gst-plugins-good',
   tcpserversink: 'gst-plugins-base',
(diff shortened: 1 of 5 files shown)


mykola-mokhnach:
The idea has been borrowed from https://stackoverflow.com/questions/27394916/android-native-opengl-openmax-screen-capture/27475298#27475298

lib/commands/streamscreen.js
```diff
@@ -0,0 +1,405 @@
...
+  if (!_.trim(await adb.shell(['which', SCREENRECORD_BINARY]))) {
+    throw new Error(
+      `The required '${SCREENRECORD_BINARY}' binary is not available on the device under test`);
+  }
+
+  for (const binaryName of [GSTREAMER_BINARY, GST_INSPECT_BINARY]) {
+    try {
+      await fs.which(binaryName);
+    } catch (e) {
+      throw new Error(`The ${binaryName} is not available in PATH on the host system. ` +
```
imurchie: This error message reads confusedly to me. I think it should say something like
```js
`The binary '${binaryName}' is not available in the PATH on the host system.`
```

lib/commands/streamscreen.js
```diff
@@ -0,0 +1,405 @@
...
+        `See ${GST_TUTORIAL_URL} for more details on how to install it.`);
+    }
+  }
+}
+
+async function getDeviceInfo (adb) {
+  const output = await adb.shell(['dumpsys', 'display']);
+  const widthMatch = /\bdeviceWidth=(\d+)/.exec(output);
+  if (!widthMatch) {
+    throw new Error(`Cannot parse the device width from ${output}`);
```
imurchie: If `output` is in single quotes it would be easier to see if there is oddities in whitespace in the output, when an error occurs.

lib/commands/streamscreen.js
```diff
@@ -0,0 +1,405 @@
...
+    screenRecordCmd += ` --bit-rate=${adjustedBitrate}`;
+  }
+  const adbArgs = [
+    ...adb.executable.defaultArgs,
+    'exec-out',
+    // The loop is required, because by default the maximum record duration
+    // for screenrecord is always limited
+    `while true; do ${screenRecordCmd} -; done`,
+  ];
+  const deviceStreaming = spawn(adb.executable.path, adbArgs);
```
imurchie: No way to do this command using `teen_process`?

mykola-mokhnach: I did it via teen_process first, but it looks like it adds too much overhead. Here we specifically need the process stdout to be piped with minimal delay.

lib/commands/streamscreen.js
mykola-mokhnach: the output here is quite long multiline string. How would you advice to print it?

lib/commands/streamscreen.js
imurchie: Ok.

lib/commands/streamscreen.js
imurchie: As it stands its all being added to the error. I just think it is good to have some sort of delimiter in case something is off. There have been cases where whitespace has mattered and it's been unclear in the logging.

mykola-mokhnach: As an option we could split the error message and the adb output itself to separate logging calls.

mykola-mokhnach:
@imurchie Do you have any more comments?

lib/commands/streamscreen.js
```diff
@@ -0,0 +1,408 @@
...
+    '!', 'jpegenc', `quality=${quality}`,
+    '!', 'multipartmux', `boundary=${BOUNDARY_STRING}`,
+    '!', 'tcpserversink', `host=${TCP_HOST}`, `port=${tcpPort}`,
+  ], {
+    stdio: [deviceStreamingProc.stdout, 'pipe', 'pipe']
+  });
+  gstreamerPipeline.on('exit', (code, signal) => {
+    log.debug(`Pipeline streaming process exited with code ${code}, signal ${signal}`);
+  });
+  const gstreamerLogger = logger.getLogger(`gstreamer@${this.opts.udid}`);
```
imurchie: This might create relatively unreadable lines, since the prefix will be really long, no?

mykola-mokhnach: the length of udid is about 12 chars. We can make it shorter if needed in the future

lib/commands/streamscreen.js
```diff
@@ -0,0 +1,408 @@
...
+    throw e;
+  }
+
+  let mjpegSocket;
+  let mjpegServer;
+  try {
+    await new B((resolve, reject) => {
+      mjpegSocket = net.createConnection(tcpPort, TCP_HOST, () => {
+        log.info(`Successfully connected to MJPEG stream at tcp://${TCP_HOST}:${tcpPort}`);
+        mjpegServer = http.createServer((req, res) => {
```
jlipps: is there no existing mjpeg server package we can use? or, should we extract this into its own utility?

mykola-mokhnach: I prefer to implement the stuff on my own if it is tiny rather than adding more dependencies. Nevertheless feel free to change that if you think the existing package would perform better or contains something the current solution does not have

PR #589, opened 2019-12-12
feat: Protect the screen streaming by feature setting

mykola-mokhnach:
I think it makes sense to hide the feature under the server switch, since opening connections on any given host/port is not a very secure practice. Also added an additional `pathname` argument to have a possibility to limit the broadcast to the given secret in the pathname

### DIFF
--- a/lib/commands/streamscreen.js
+++ b/lib/commands/streamscreen.js
@@ -223,10 +223,6 @@ function extractRemoteAddress (req) {
     || req.connection.socket.remoteAddress;
 }
 
-function extractUserAgent (req) {
-  return req.headers['user-agent'];
-}
-
 
 /**
  * @typedef {Object} StartScreenStreamingOptions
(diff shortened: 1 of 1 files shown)


lib/commands/streamscreen.js
```diff
@@ -249,12 +253,15 @@ async function initGstreamerPipeline (deviceStreamingProc, deviceInfo, opts = {}
  * @throws {Error} If screen streaming has failed to start or is not supported on the host system.
  */
 commands.mobileStartScreenStreaming = async function mobileStartScreenStreaming (options = {}) {
+  this.ensureFeatureEnabled(ADB_SCREEN_STREAMING_FEATURE);
```
KazuCocoa: Good idea 👍 
Need to update https://github.com/appium/appium-android-driver#opt-in-features-with-security-risk , too.

PR #604, opened 2020-02-19
fix: Fix logger creation inside streaming helpers

mykola-mokhnach:


lib/commands/streamscreen.js
```diff
@@ -119,7 +120,11 @@ async function initDeviceStreamingProc (adb, deviceInfo, opts = {}) {
   });
 
   let isStarted = false;
-  const deviceStreamingLogger = logger.getLogger(`${SCREENRECORD_BINARY}@${this.opts.udid}`);
+  const deviceStreamingLogger = logger.getLogger(`${SCREENRECORD_BINARY}@` +
+    _.truncate(deviceInfo.udid, {
```
imurchie: This seems like it can be pulled out, since it is used twice in the same way.

mykola-mokhnach: changed

jlipps:
Thanks!

PR #607, opened 2020-02-24
feat: Add a possibility to pass a letter as a platformVersion

mykola-mokhnach:
https://github.com/appium/appium/issues/13983

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -3,7 +3,7 @@ import path from 'path';
 import { exec } from 'teen_process';
 import { retry, retryInterval } from 'asyncbox';
 import logger from './logger';
-import { fs, util } from 'appium-support';
+import { fs } from 'appium-support';
 import { path as settingsApkPath } from 'io.appium.settings';
 import Bootstrap from './bootstrap';
 import B from 'bluebird';
@@ -12,6 +12,7 @@ import {
   default as unlocker, PIN_UNLOCK, PASSWORD_UNLOCK,
   PATTERN_UNLOCK, FINGERPRINT_UNLOCK } from './unlock-helpers';
 import { EOL } from 'os';
+import semver from 'semver';
 
 const PACKAGE_INSTALL_TIMEOUT = 90000; // milliseconds
 const CHROME_BROWSER_PACKAGE_ACTIVITY = {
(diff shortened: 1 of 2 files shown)


lib/android-helpers.js
```diff
@@ -203,49 +204,45 @@ helpers.getDeviceInfoFromCaps = async function getDeviceInfoFromCaps (opts = {})
...
-          || (majorPlatformVersion === majorDeviceVersion && minorPlatformVersion === minorDeviceVersion)) {
-          // Got a partial match - make sure we consider the most recent
-          // device version available on the host system
-          if (partialMatchCandidate
-              && util.compareVersions(deviceOS, '>', _.values(partialMatchCandidate)[0])
-              || !partialMatchCandidate) {
-            partialMatchCandidate = {[device.udid]: deviceOS};
-          }
+        if ((!_.includes(opts.platformVersion, '.') && platformVersion.major === deviceOS.major
+            || platformVersion.major === deviceOS.major && platformVersion.minor === deviceOS.minor)
```
imurchie: Should there be some extra parentheses in here? I'm trying to parse this condition and it is confusing.

lib/android-helpers.js
```diff
@@ -203,49 +204,45 @@ helpers.getDeviceInfoFromCaps = async function getDeviceInfoFromCaps (opts = {})
...
-        const deviceOS = util.coerceVersion(rawDeviceOS, false);
+        const deviceOS = semver.coerce(rawDeviceOS) || rawDeviceOS;
         if (!deviceOS) {
           continue;
         }
 
-        if (util.compareVersions(deviceOS, '==', platformVersion)) {
+        const bothVersionsCanBeCoerced = semver.valid(deviceOS) && semver.valid(platformVersion);
+        const bothVersionsAreStrings = _.isString(deviceOS) && _.isString(platformVersion);
+        if (bothVersionsCanBeCoerced && deviceOS.version === platformVersion.version
```
KazuCocoa: 👍

lib/android-helpers.js
mykola-mokhnach: no parentheses needed, AND has priority over OR. The meaning of this particular expression is:
- if we don't have a dot in platformVersion then only compare major numbers
- otherwise compare both major and minor version numbers
- if any of the conditions above is truthy then bump our candidate if needed ( e.g. the current candidate is not assigned yet or its version is older than the current deviceOS version)

PR #609, opened 2020-02-26
fix: Update emulator detection algorithm

mykola-mokhnach:


lib/driver.js
```diff
@@ -195,7 +195,8 @@ class AndroidDriver extends BaseDriver {
   }
 
   isEmulator () {
-    return !!(this.opts.avd || /emulator/.test(this.opts.udid));
+    const possibleNames = [this.opts.udid, this.adb?.curDeviceId];
+    return !!this.opts.avd || possibleNames.some((x) => /\bemulator\b/i.test(x));
```
imurchie: I think it would be better to have the regular expression as a constant outside, so it's not created a couple of times each time `isEmulator` is called.

Quick benchmark:
```
[08:51:48] internal regexp x 14,436,765 ops/sec ±2.40% (80 runs sampled)
[08:51:54] external regexp x 18,075,762 ops/sec ±1.77% (86 runs sampled)
```

PR #614, opened 2020-03-11
feat: Add ignore hidden api policy error

KazuCocoa:
Add `ignoreHiddenApiPolicyError` to ignore permission error when set/delete hidden api policy to relax the security. Some devices like a Chinese phone raises a permission error when the adb shell is called. Potentially, the security error blocks our change settings like animation scale, but it could be ignored.

Defaults to false. (Does not ignore the error)

https://github.com/appium/appium-adb/pull/507

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -247,7 +247,7 @@ class AndroidDriver extends BaseDriver {
         if (await this.adb.getApiLevel() >= 28) { // API level 28 is Android P
           // Don't forget to reset the relaxing in delete session
           log.warn('Relaxing hidden api policy to manage animation scale');
-          await this.adb.setHiddenApiPolicy('1');
+          await this.adb.setHiddenApiPolicy('1', this.opts.ignoreHiddenApiPolicyError || false);
         }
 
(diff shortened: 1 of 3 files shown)


lib/driver.js
```diff
@@ -247,7 +247,7 @@ class AndroidDriver extends BaseDriver {
         if (await this.adb.getApiLevel() >= 28) { // API level 28 is Android P
           // Don't forget to reset the relaxing in delete session
           log.warn('Relaxing hidden api policy to manage animation scale');
-          await this.adb.setHiddenApiPolicy('1');
+          await this.adb.setHiddenApiPolicy('1', this.opts.ignoreHiddenApiPolicyError || false);
```
mykola-mokhnach: `|| false` is not needed
you could simply do `!!this.opts.ignoreHiddenApiPolicyError` to make sure it is a boolean value

mykola-mokhnach: same comment is relevant for the two other linked PRs

KazuCocoa: Yes. Will fix them after this new version push

PR #622, opened 2020-04-01
Provide an option to set successful unlock timeout

a-en:
I'm facing the issue when 2 seconds is not enough in order to verify that screen was unlocked successfully. So in scope of this PR I'd like to propose the option to provide the value of this timeout as part of the desired capabilities. By default (if not provided) this timeout still will be 2 seconds.

If the name of proposed capability is fine (`unlockSuccessTimeout`), I could update the list of desired capabilities for Android.

### DIFF
--- a/lib/android-helpers.js
+++ b/lib/android-helpers.js
@@ -665,12 +665,13 @@ helpers.unlock = async function unlock (driver, adb, capabilities) {
     await helpers.unlockWithHelperApp(adb);
   } else {
     await helpers.unlockWithUIAutomation(driver, adb, {unlockType: capabilities.unlockType, unlockKey: capabilities.unlockKey});
-    await helpers.verifyUnlock(adb);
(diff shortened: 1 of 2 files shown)


lib/android-helpers.js
```diff
@@ -665,12 +665,13 @@ helpers.unlock = async function unlock (driver, adb, capabilities) {
     await helpers.unlockWithHelperApp(adb);
   } else {
     await helpers.unlockWithUIAutomation(driver, adb, {unlockType: capabilities.unlockType, unlockKey: capabilities.unlockKey});
-    await helpers.verifyUnlock(adb);
+    await helpers.verifyUnlock(adb, capabilities.unlockSuccessTimeout);
```
mykola-mokhnach, line 668: Please add the new capability to https://github.com/appium/appium-android-driver/blob/master/lib/desired-caps.js

PR #631, opened 2020-04-20
fix: Only provide element id to actions where it is needed

mykola-mokhnach:
The current touch actions algorithm used in android driver is quite counter-intuitive. The `parseTouch` already precalculates absolute gesture coordinates for the given elements, however it still passes element id with absolute coordinates to the downstream helpers. As far as I can understand the element id is used as a backup value if the coordinates action fails (even in that case the backend ignores the given x and y offsets and simply uses the central element point for the action). Also, since the coordinates precalculation is done on the Appium driver side (element location+size), there are many extra server calls, which slows down gestures in general. 

I don't want to change much there because of legacy reasons. Let's just align the action executor to only pass transformed coordinates without backup element ids to not confuse the backend and make the logic a bit more straightforward.

### DIFF
--- a/lib/commands/touch.js
+++ b/lib/commands/touch.js
@@ -7,23 +7,23 @@ import { asyncmap } from 'asyncbox';
 
 let commands = {}, helpers = {}, extensions = {};
 
-commands.doTouchAction = async function doTouchAction (action, opts) {
+commands.doTouchAction = async function doTouchAction (action, opts = {}) {
+  const { element, x, y, count, ms, duration } = opts;
+  // parseTouch precalculates absolute element positions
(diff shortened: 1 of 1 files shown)


lib/commands/touch.js
```diff
@@ -7,23 +7,23 @@ import { asyncmap } from 'asyncbox';
 
 let commands = {}, helpers = {}, extensions = {};
 
-commands.doTouchAction = async function doTouchAction (action, opts) {
+commands.doTouchAction = async function doTouchAction (action, opts = {}) {
+  const { element, x, y, count, ms, duration } = opts;
+  // parseTouch precalculates absolute element positions
```
KazuCocoa, line 12: 👍

PR #644, opened 2020-05-20
feat: Add support for logcat output customization

mykola-mokhnach:
Based on https://github.com/appium/appium-adb/pull/528

### DIFF
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -248,6 +248,12 @@ let commonCapConstraints = {
   },
   mockLocationApp: {
     isString: true
+  },
+  logcatFormat: {
+    isArray: true,
+  },
+  logcatFilterSpecs: {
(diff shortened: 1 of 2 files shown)


lib/desired-caps.js
```diff
@@ -248,6 +248,12 @@ let commonCapConstraints = {
   },
   mockLocationApp: {
     isString: true
+  },
+  logcatFormat: {
+    isArray: true,
```
KazuCocoa, line 253: Should here be isString?

mykola-mokhnach: [it is](https://github.com/appium/appium-android-driver/commit/82f25257df4eb48814e1b4f878f9c4ede6beb7b8) ;)

KazuCocoa: ah, yey 👍

PR #646, opened 2020-05-27
feat: Make it possible to retrieve native logs being in web context

mykola-mokhnach:
https://github.com/appium/appium/issues/3412


mykola-mokhnach:
This change itself does nothing for Android driver, but it will do as soon as we bump it to Espresso/UIA2 and add logs endpoints to chromedriver no proxy list

PR #662, opened 2020-09-21
feat: "mobile: getContexts" on Android

mwakizaka:
This PR adds an endpoint named `mobile: getContexts` by reusing a logic for `ensureWebviewsHavePages`

Related to https://github.com/appium/appium/issues/14745

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -62,9 +62,12 @@ commands.setContext = async function setContext (name) {
 };
 
 commands.mobileGetContexts = async function mobileGetContexts () {
-  const opts = _.cloneDeep(this.opts);
-  opts.ensureWebviewsHavePages = true;
-  opts.enableWebviewDetailsCollection = true;
+  const opts = {
+    androidDeviceSocket: this.opts.androidDeviceSocket,
+    ensureWebviewsHavePages: true,
+    webviewDevtoolsPort: this.opts.webviewDevtoolsPort,
+    enableWebviewDetailsCollection: true
+  };
   return await webviewHelpers.getWebViewsMapping(this.adb, opts);
 };
 
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -61,6 +61,13 @@ commands.setContext = async function setContext (name) {
   this.curContext = name;
 };
 
+commands.mobileGetContexts = async function mobileGetContexts () {
+  const opts = _.cloneDeep(this.opts);
+  opts.ensureWebviewsHavePages = true;
+  opts.enableWebviewDetailsCollection = true;
+  return await webviewHelpers.getWebViewsMapping(this.adb, opts);
+};
+
 helpers.switchContext = async function switchContext (name) {
   // We have some options when it comes to webviews. If we want a
   // Chromedriver webview, we can only control one at a time.
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -61,6 +61,52 @@ commands.setContext = async function setContext (name) {
   this.curContext = name;
 };
 
+/**
+ * @typedef {Object} WebviewsMapping
+ * @property {string} proc The name of the Devtools Unix socket
+ * @property {string} webview The web view alias. Looks like `WEBVIEW_`
+ * prefix plus PID or package name
+ * @property {?Object} info Webview information as it is retrieved by
+ * /json/version CDP endpoint
+ * @property {?Array<Object>} pages Webview pages list as it is retrieved by
+ * /json/list CDP endpoint
+ * @propery {?string} webviewName An actual webview name for switching context.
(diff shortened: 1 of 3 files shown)


lib/webview-helpers.js
```diff
@@ -370,6 +368,31 @@ helpers.getWebviews = async function getWebviews (adb, {
...
 
+/**
+ * Retrieves web views mapping for getContexts
+ *
+ * @param {object} adb - an ADB instance
+ * @param {GetWebviewOpts} opts (see note on getWebviews).
+ *
+ * @return {Array<WebviewProps>} webviewsMapping(see note no collectWebviewsDetails).
+ */
+helpers.getWebViewsMapping = async function getWebViewsMapping (adb, {
```
mykola-mokhnach, line 379: The names of web views returned by this method might differ from the ones returned in `getWebviews`. Please unify them

mwakizaka: Thank you for your comment. I think current implementation meets your comment. Now the object in returned value of  `getWebViewsMapping` will have a property named `webviewName` (e.g.  `WEBVIEW_<package_name>`). 

However, the value `webviewName` will be `undefined` when it reaches to https://github.com/appium/appium-android-driver/blob/master/lib/webview-helpers.js#L357 

Should `webviewName` have any dummy value other than `undefined`?

lib/webview-helpers.js
```diff
@@ -289,6 +289,31 @@ helpers.procFromWebview = async function procFromWebview (adb, webview) {
...
+  webviewDevtoolsPort = null,
+  enableWebviewDetailsCollection = null,
+} = {}) {
+  const webviewsMapping = await this.getWebViewsMapping(adb, {
+    androidDeviceSocket,
+    ensureWebviewsHavePages,
+    webviewDevtoolsPort,
+    enableWebviewDetailsCollection
+  });
+  const result = webviewsMapping.filter(function (w) {return w.webviewName;}).map(function (w) {return w.webviewName;});
```
mykola-mokhnach, line 312:
```
const result = webviewsMapping.reduce((acc, {webviewName}) => {
  if (webviewName) {
    acc.push(webviewName);
  }
  return acc;
}, []);
```
or 
```
const result = webviewsMapping.filter(({webviewName}) => !!webviewName).map(({webviewName}) => webviewName);
```

lib/webview-helpers.js
```diff
@@ -336,8 +371,8 @@ helpers.getWebviews = async function getWebviews (adb, {
   // webviewProcs contains procs, which we only care about for ensuring
   // presence of pages above, so we can now discard them and rely on just the
   // webview names
-  const result = [];
-  for (const {webview, pages, info, proc} of webviewsMapping) {
+  for (const webviewMapping of webviewsMapping) {
+    const {webview, pages, info, proc} = webviewMapping;
     if (ensureWebviewsHavePages && pages?.length === 0) {
```
mykola-mokhnach, line 376: this filtering should probably be done in the calling function

lib/commands/context.js
```diff
@@ -61,6 +61,13 @@ commands.setContext = async function setContext (name) {
   this.curContext = name;
 };
 
+commands.mobileGetContexts = async function mobileGetContexts () {
+  const opts = _.cloneDeep(this.opts);
```
mykola-mokhnach, line 65: should we really clone all opts here? we know which options are needed, so it would be fine to just create options objections here based on opts values

lib/webview-helpers.js
mykola-mokhnach: > Should webviewName have any dummy value other than undefined?

You could set the name to `null` for this case

mwakizaka: Thank you. It's done.

lib/webview-helpers.js
mwakizaka: I have addressed this along with https://github.com/appium/appium-android-driver/pull/662#discussion_r492533011

lib/commands/context.js
mwakizaka: Ah, no. So I just created a opt object there.

lib/commands/context.js
```diff
@@ -61,6 +61,16 @@ commands.setContext = async function setContext (name) {
   this.curContext = name;
 };
 
+commands.mobileGetContexts = async function mobileGetContexts () {
```
mykola-mokhnach, line 64: Please add a dosctring here that describes the returned result and its format (preferable with a JSON example)

mwakizaka: Thanks. I have added a doc string there. Could you check it?

mykola-mokhnach (approved):
perfect. Thanks for your work @mwakizaka

mykola-mokhnach:
Published in appium-android-driver@4.40.0

PR #670, opened 2020-11-18
feat: relax webview pid check with devtools port

KazuCocoa:
Related issue: https://github.com/appium/appium-desktop/issues/1618

According to the issue, WebView has a case the response by CDP (in the app under test) has webview(s) but the process id is not the app under test's one.

This PR tries to relax the process check.

### DIFF
--- a/lib/webview-helpers.js
+++ b/lib/webview-helpers.js
@@ -391,13 +391,9 @@ helpers.getWebViewsMapping = async function getWebViewsMapping (adb, {
         const pkg = pkgMatch ? pkgMatch[1] : await helpers.procFromWebview(adb, webview);
         wvName = `${WEBVIEW_BASE}${pkg}`;
       } catch (e) {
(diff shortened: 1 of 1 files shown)


ChaseSkywalker:
i try this but found out that the 'webviewDevtoolsPort' is null, and i can`t find any code that changes its value

image

image


i think if i don`t provide 'webviewDevtoolsPort' in capability, it remains null

image

mykola-mokhnach:
@ChaseSkywalker Are you sure you use the recent sources? There was a change in the recent Android SDKs while running `ps` without `-A` option was only showing processes created by the current user. This has been fixed by https://github.com/appium/appium-adb/pull/550

KazuCocoa:
Ah, the change is in 1.19.0. Commented it https://github.com/appium/appium-desktop/issues/1618#issuecomment-734460314

PR #672, opened 2020-12-07
fix: did not wait long enough for the app to come back after background

jlipps:
turns out a simple `await B.delay` can be off by a factor of minutes for long values. This looping strategy checks in with the actual system clock to make sure we wait long enough.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -1,10 +1,9 @@
 import _ from 'lodash';
 import androidHelpers from '../android-helpers';
 import { util } from 'appium-support';
-import B from 'bluebird';
 import log from '../logger';
 import moment from 'moment';
-import { waitForCondition } from 'asyncbox';
+import { waitForCondition, longSleep } from 'asyncbox';
 
 const MOMENT_FORMAT_ISO8601 = 'YYYY-MM-DDTHH:mm:ssZ';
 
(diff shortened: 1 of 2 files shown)


KazuCocoa (commented):
seems good, but i'm not so familiar with this change's appropriateness..

lib/commands/general.js
```diff
@@ -134,9 +134,20 @@ commands.background = async function background (seconds) {
...
     return true;
   }
+  const msToWait = seconds * 1000;
   let {appPackage, appActivity} = await this.adb.getFocusedPackageAndActivity();
   await this.adb.goToHome();
-  await B.delay(seconds * 1000);
+
+  // B.delay is not reliable for long periods. So if we have a short background period, just use it
+  // directly, otherwise spin in increments of 1s for as long as we need to.
+  if (seconds < 60) {
```
mykola-mokhnach, line 143: we could use waitForCondition instead from the module you wrote ;)

jlipps: LOL true, the ergonomics of that aren't great. Let me do something else.

jlipps:
ok what do you think about this @mykola-mokhnach , now we get some nice progress logging too

lib/commands/general.js
```diff
@@ -136,7 +135,19 @@ commands.background = async function background (seconds) {
...
+
+  // people can wait for a long time, so to be safe let's use the longSleep function and log
+  // progress periodically.
+  const sleepMs = seconds * 1000;
+  const thresholdMs = 30 * 1000; // use the spin-wait for anything over this threshold
+  // for our spin interval, use 1% of the total wait time, but nothing bigger than 30s
+  const intervalMs = _.min([30 * 1000, parseInt(sleepMs / 100, 10)]);
+  const progressCb = ({elapsedMs, progress}) => {
+    const waitSecs = (elapsedMs / 1000).toFixed(0);
+    const pct = (progress * 100).toFixed(2);
```
mykola-mokhnach, line 147: I would rather give the variable a full name

mykola-mokhnach (approved):

jlipps:
published in 4.41.1

PR #673, opened 2021-01-09
fix: not to set androidPackage as null for chromedriver

KazuCocoa:
I found a case in which `androidPackage` was `null` when `noReset` was `true` case.
Newer (Not sure which version..) chromedriver and Android OS version can ignore it, but older one (at least  4.x) raised an error invalid argument by chromedriver.
Without this `androidPackage`, the request worked.

### DIFF
--- a/lib/webview-helpers.js
+++ b/lib/webview-helpers.js
@@ -405,11 +405,12 @@ helpers.getWebviewDetails = function getWebviewDetails (adb, webview) {
  * See https://chromedriver.chromium.org/capabilities for more details.
  */
 helpers.createChromedriverCaps = function createChromedriverCaps (opts, deviceId) {
-  const caps = {
-    chromeOptions: {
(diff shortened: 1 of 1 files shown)


mykola-mokhnach (approved):

lib/webview-helpers.js
```diff
@@ -405,11 +405,12 @@ helpers.getWebviewDetails = function getWebviewDetails (adb, webview) {
...
-  const caps = {
-    chromeOptions: {
-      androidPackage: opts.chromeOptions?.androidPackage || opts.appPackage,
-    }
-  };
+  const caps = { chromeOptions: {} };
+
+  const androidPackage = opts.chromeOptions?.androidPackage || opts.appPackage;
+  if (androidPackage) {
+    caps.chromeOptions.androidPackage = androidPackage;
```
mykola-mokhnach, line 412: it probably makes sense to add a comment here to explain why androidPackage  should not be set to null

PR #682, opened 2021-03-08
ci: bump sdk version

KazuCocoa:


mykola-mokhnach:
We should not bump SDK version for this driver. It is expected that the maximum supported OS version there is 5.1. For higher versions UIA2/Espresso must be used

KazuCocoa:
oh, I see.
Emulators by the sdk version seemed unstable on Azure.
Then, probably we can disable some tests or relax the assertion like partial matching. Let me tweak as another pr

mykola-mokhnach:
yes, this was the initial ida.
We could set some predefined env variable for e2e tests, for example `CI=1` and then add `if (process.env.CI) return this.skip()` to flaky tests

PR #684, opened 2021-03-18
refactor: Move swipe helper methods to android-driver

mykola-mokhnach:
This is a shame we have all these methods in base driver while they are only needed in android driver. The next step would be to remove them from base driver for good


KazuCocoa (approved):

KazuCocoa:
agree 👍

PR #691, opened 2021-05-05
feat: add pinWithKeyEvent to put keys one by one for pin as key evnets

KazuCocoa:
Some devices like Samsung have customized keyboard in `pin`.
Then, it is reliable to send keys as keycode like `password` rather than finding numbers with ids. I named the strategy as `pinWithKeyEvent` for now. (Suggestions are welcome)

As backward compatibility, `pin` should be there, but the `pinUnlockWithKeyEvent` can be used as a fallback method when no system key ids in the pin view. Afaik, this fallback is pretty enough.

---

After this change:
- https://github.com/appium/appium-android-driver/blob/master/docs/UNLOCK.md
- readme in uia2 and espresso

### DIFF
--- a/lib/unlock-helpers.js
+++ b/lib/unlock-helpers.js
@@ -4,27 +4,41 @@ import _ from 'lodash';
 import { util } from 'appium-support';
 
 const PIN_UNLOCK = 'pin';
+const PIN_UNLOCK_KEY_EVENT = 'pinWithKeyEvent';
 const PASSWORD_UNLOCK = 'password';
 const PATTERN_UNLOCK = 'pattern';
 const FINGERPRINT_UNLOCK = 'fingerprint';
-const UNLOCK_TYPES = [PIN_UNLOCK, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGERPRINT_UNLOCK];
+const UNLOCK_TYPES = [PIN_UNLOCK, PIN_UNLOCK_KEY_EVENT, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGERPRINT_UNLOCK];
 const KEYCODE_NUMPAD_ENTER = 66;
 const KEYCODE_POWER = 26;
 const KEYCODE_WAKEUP = 224; // Can work over API Level 20
 const UNLOCK_WAIT_TIME = 100;
 const HIDE_KEYBOARD_WAIT_TIME = 100;
 const INPUT_KEYS_WAIT_TIME = 100;
+// mapping of Android keycode and numbers
+const KEY_CODE_NUMBER = {
+  '0': 7,
+  '1': 8,
+  '2': 9,
+  '3': 10,
+  '4': 11,
+  '5': 12,
+  '6': 13,
+  '7': 14,
+  '8': 15,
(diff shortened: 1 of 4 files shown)


lib/unlock-helpers.js
```diff
@@ -4,27 +4,41 @@ import _ from 'lodash';
...
+const UNLOCK_TYPES = [PIN_UNLOCK, PIN_UNLOCK_KEY_EVENT, PASSWORD_UNLOCK, PATTERN_UNLOCK, FINGERPRINT_UNLOCK];
 const KEYCODE_NUMPAD_ENTER = 66;
 const KEYCODE_POWER = 26;
 const KEYCODE_WAKEUP = 224; // Can work over API Level 20
 const UNLOCK_WAIT_TIME = 100;
 const HIDE_KEYBOARD_WAIT_TIME = 100;
 const INPUT_KEYS_WAIT_TIME = 100;
+// mapping of Android keycode and numbers
+const ANDROID_KEY_CODE_NUMBER = {
+  '0': 7,
```
mykola-mokhnach, line 20: I believe this mapping could be generated dynamically, there is no need to hardcode each item

lib/unlock-helpers.js
```diff
@@ -102,7 +116,8 @@ helpers.pinUnlock = async function pinUnlock (adb, driver, capabilities) {
   if (await adb.getApiLevel() >= 21) {
     let els = await driver.findElOrEls('id', 'com.android.systemui:id/digit_text', true);
     if (_.isEmpty(els)) {
-      throw new Error('Error finding unlock pin buttons!');
+      // fallback to pin with key event
```
mykola-mokhnach, line 119: it would make sense to add a log here

lib/unlock-helpers.js
mykola-mokhnach:
```
const ZERO_KEY_CODE = 7;
result = _.range(10).reduce((acc) => {
  acc[String.fromCharCode('0'.charCodeAt(0) + _.size(acc))] = ZERO_KEY_CODE + _.size(acc);
  return acc;
}, {});
```

lib/unlock-helpers.js
```diff
@@ -131,6 +135,28 @@ helpers.pinUnlock = async function pinUnlock (adb, driver, capabilities) {
...
+  // Then, sending keyevents are more reliable to unlock the screen.
+  for (const pin of keys) {
+    // 'pin' is number (0-9) in string.
+    // Number '0' is keycode '7'. number '9' is keycode '16'.
+    await adb.shell(['input', 'keyevent', parseInt(pin, 10) + NUMBER_ZERO_KEYCODE]);
+  }
+
+  // Some devices accept entering the code without pressing the Enter key
+  // When I rushed commands without this wait before pressKeyCode, rarely UI2 sever crashed
+  await sleep(UNLOCK_WAIT_TIME);
```
mykola-mokhnach, line 153: can we extract this duplicated block into a separate function to avoid duplication?

KazuCocoa: 👍

mykola-mokhnach (approved):

PR #693, opened 2021-05-21
feat: add server arg parsers

ayvnkhan:


lib/android-helpers.js
```diff
@@ -890,5 +890,37 @@ helpers.isEmulator = function isEmulator (adb, opts) {
 helpers.bootstrap = Bootstrap;
 helpers.unlocker = unlocker;
 
+helpers.parseReboot = function (driverArgs) {
```
mykola-mokhnach, line 893: I don't think we need these helpers. What is the possible usage scenario there?

mykola-mokhnach (changes requested):
See https://github.com/appium/appium-uiautomator2-driver/pull/459/files#r637359764

lib/android-helpers.js
```diff
@@ -890,5 +890,61 @@ helpers.isEmulator = function isEmulator (adb, opts) {
 helpers.bootstrap = Bootstrap;
 helpers.unlocker = unlocker;
 
-export { helpers, SETTINGS_HELPER_PKG_ID, prepareAvdArgs, ensureNetworkSpeed };
+const SERVER_ARGS = {
```
mykola-mokhnach, line 893: this code basically repeats capability validators from the base driver. I don't see a need in such duplication

mykola-mokhnach: See https://github.com/appium/appium-base-driver/blob/de7e41bbf7e7dc9c61749f094b997f95e73eef9b/lib/basedriver/capabilities.js#L24

jlipps: i think it's good to define the possible server args separately from the capabilities. but i agree, we could probably just use the existing caps validation logic on these. in addition, it's not necessary to put strings around key names in JS unless they contain a special character. pretty sure our lint will catch this if you run it.

jlipps (changes requested):

lib/android-helpers.js
```diff
@@ -890,5 +891,57 @@ helpers.isEmulator = function isEmulator (adb, opts) {
...
+    } else {
+      throw new Error(`"${argName}" is not a recognized key`);
+    }
+    return acc;
+  }, {});
+};
+
+helpers.parseArgs = function (opts, allArgs, knownArgNames) {
+  let acc = helpers.parseKnownArgs(opts, allArgs, knownArgNames);
+  _.keys(DEFAULT_SERVER_ARGS).forEach((arg) => {
```
mykola-mokhnach, line 933: a vanilla for loop would be more readable here

lib/android-helpers.js
```diff
@@ -890,5 +891,57 @@ helpers.isEmulator = function isEmulator (adb, opts) {
...
+    if (!_.has(acc, arg)) {
+      if (_.has(opts, arg)) {
```
mykola-mokhnach, line 935: acc[arg] = _.has(opts, arg) ? opts[arg] : DEFAULT_SERVER_ARGS[arg]

lib/android-helpers.js
```diff
@@ -890,5 +891,57 @@ helpers.isEmulator = function isEmulator (adb, opts) {
 helpers.bootstrap = Bootstrap;
 helpers.unlocker = unlocker;
 
-export { helpers, SETTINGS_HELPER_PKG_ID, prepareAvdArgs, ensureNetworkSpeed };
+const CONSTRAINTS = {
+  reboot: {
+    isBoolean: true,
+    presence: true
```
mykola-mokhnach, line 897: is it really mandatory that these values are present?

lib/android-helpers.js
```diff
@@ -890,5 +891,57 @@ helpers.isEmulator = function isEmulator (adb, opts) {
...
+  chromedriverExecutable: null,
+  suppressKillServer: false,
+};
+
+helpers.parseKnownArgs = function (opts, allArgs, knownArgNames) {
+  return _.toPairs(allArgs).reduce((acc, [argName, argValue]) => {
+    if (knownArgNames.includes(argName)) {
+      acc[argName] = argValue;
+    } else if (_.has(opts, argName)) {
+      acc[argName] = opts[argName];
```
mykola-mokhnach, line 923: is this really needed? why should we validate items of `opts`?

lib/android-helpers.js
```diff
@@ -890,5 +891,57 @@ helpers.isEmulator = function isEmulator (adb, opts) {
...
+    } else {
+      throw new Error(`"${argName}" is not a recognized key`);
```
mykola-mokhnach, line 925: I would also print the supported key names here

lib/android-helpers.js
```diff
@@ -13,6 +13,7 @@ import {
 } from './unlock-helpers';
 import { EOL } from 'os';
 import semver from 'semver';
+import { validateCaps } from 'appium-base-driver/build/lib/basedriver/capabilities';
```
mykola-mokhnach, line 16: this does not look like a valid import to me. The function should be present in package's index in order to be imported

lib/android-helpers.js
ayvnkhan: I'd assume so because usually appium passes in default arg values to the driver in the case none were given. Now in appium 2.0 we are trying to depreciate driver specific args. So once the parsing is done I'd assume we want validateCaps to ensure that the args are present(they could be either default or user passed in values)

ayvnkhan:
this branch depends on https://github.com/appium/appium/pull/15451/files

lib/android-helpers.js
```diff
@@ -890,5 +891,51 @@ helpers.isEmulator = function isEmulator (adb, opts) {
...
+  },
+  chromedriverExecutable: {
+    isString: true,
+  },
+  suppressKillServer: {
+    isBoolean: true,
+  },
+};
+
+const DEFAULT_SERVER_ARGS = {
```
mykola-mokhnach, line 909: I'd remove these defaults, since the default values for corresponding capabilities are anyway already set to these. Which means in case the corresponding arg is not set then we just skip it for good

lib/android-helpers.js
```diff
@@ -890,5 +891,51 @@ helpers.isEmulator = function isEmulator (adb, opts) {
 helpers.bootstrap = Bootstrap;
 helpers.unlocker = unlocker;
 
-export { helpers, SETTINGS_HELPER_PKG_ID, prepareAvdArgs, ensureNetworkSpeed };
+const CONSTRAINTS = {
```
mykola-mokhnach, line 894: ARGS_CONTRAINTS

lib/android-helpers.js
```diff
@@ -13,6 +13,7 @@ import {
 } from './unlock-helpers';
 import { EOL } from 'os';
 import semver from 'semver';
+import { validateCaps } from 'appium-base-driver';
```
mykola-mokhnach, line 16: The minimum base driver version in package.json should be updated for this to work

ayvnkhan: updated

lib/android-helpers.js
```diff
@@ -890,5 +891,46 @@ helpers.isEmulator = function isEmulator (adb, opts) {
...
+  },
+  chromedriverExecutable: {
+    isString: true,
+  },
+  suppressKillServer: {
+    isBoolean: true,
+  },
+};
+
+helpers.parseKnownArgs = function (allArgs, knownArgNames) {
```
jlipps, line 909: are we going to reimplement these functions in every driver? if so, we'd be better off putting this in appium-support or something like that (with a TODO that it should really be in basedriver once appium 2 is out). also, i don't see where these functions are used?

ayvnkhan: Would be better to put them in appium-support I agree, and they are used by the android drivers: uiautomator2, espresso.

PR #700, opened 2021-06-26
fix: Prefer file-based lock to acquire devtools port

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/15390#issuecomment-868864881

I still wouldn't rely on the automated port detection thing though and advice to set the `webviewDevtoolsPort` value manually for each parallel session though.


KazuCocoa (approved):
agree with that
> wouldn't rely on the automated port detection thing

PR #701, opened 2021-07-10
feat: Add mobile extension to start activities

mykola-mokhnach:
The current endpoint for activity startup, which is defined in base driver, is not very flexible and requires too much effort to add new args/features. It makes sense to move its functionality into an extension. I've also added the support for more modern options that am supports as of Android 11.


jlipps (approved):

lib/commands/intent.js
```diff
@@ -0,0 +1,274 @@
...
+    intent,
+    user,
+    wait,
+    stop,
+    windowingMode,
+    activityType,
+    display,
+  } = requireOptions(opts, ['intent']);
+  const cmd = [
+    'am', 'start-activity',
```
KazuCocoa, line 152: Should we keep `start` so far?

I should check which version, but an Android  5 emulator needed `start`. Newer ones still support `start`, so `start` is better so far.
(I noticed `startservice` vs `start-service` as well... let me check emulators.)

lib/commands/intent.js
```diff
@@ -0,0 +1,274 @@
...
+ * or required options are missing
+ */
+commands.mobileStartService = async function mobileStartService (opts = {}) {
+  const {
+    intent,
+    user,
+    foreground,
+  } = requireOptions(opts, ['intent']);
+  const cmd = [
+    'am', foreground ? 'start-foreground-service' : 'start-service',
```
KazuCocoa, line 230: It seems `start-service` should be `startservice` for lower than Android 8 as same as start-activity.

lib/commands/intent.js
```diff
@@ -0,0 +1,274 @@
...
+ * @throws {Error} If there was a failure while stopping the service
+ * or required options are missing
+ */
+commands.mobileStopService = async function mobileStopService (opts = {}) {
+  const {
+    intent,
+    user,
+  } = requireOptions(opts, ['intent']);
+  const cmd = [
+    'am', 'stop-service',
```
KazuCocoa, line 262: s/stop-service/stopservice/ as same as startservive.

lib/commands/intent.js
KazuCocoa: Android 7 was still `start`. 8 was `start-activity`.
We can use `start-activity` when we stop supporting Android 7 and lower, I think.
(Android 11 was able to recognize `start` as start-service)

I also think a comment about this with `-` or not depends on Android version will help out future work. (the below start/stop services as well)

mykola-mokhnach: Oh, I didn't know it was changed between Android versions. Thanks for the hint. Fixed now.

lib/commands/intent.js
mykola-mokhnach: Fixed

lib/commands/intent.js
mykola-mokhnach: fixed

KazuCocoa (approved):

PR #703, opened 2021-09-07
chore: remove autoLaunch

Andrewngabriel:
Remove `autoLaunch`

More info [#15807](https://github.com/appium/appium/issues/15807) and [#15815](https://github.com/appium/appium/pull/15815)

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -91,7 +91,6 @@ class AndroidDriver extends BaseDriver {
         disableAndroidWatchers: false,
         tmpDir: await tempDir.staticDir(),
         fullReset: false,
-        autoLaunch: true,
         adbPort: DEFAULT_ADB_PORT,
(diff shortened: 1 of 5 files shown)


mykola-mokhnach:
Same as https://github.com/appium/appium-xcuitest-driver/pull/1332#issuecomment-914038569

lib/driver.js
```diff
@@ -284,11 +283,6 @@ class AndroidDriver extends BaseDriver {
     if (this.isChromeSession) {
       // start a chromedriver session and proxy to it
       await this.startChromeSession();
-    } else {
-      if (this.opts.autoLaunch) {
```
jlipps, line 288: since `autoLaunch` defaults to true, we don't want to remove this block. We only want to remove behavior that would occur if `this.opts.autoLaunch` is *not* true.

PR #704, opened 2021-09-14
feat: Add a possibility to send an intent broadcast with a mobile command

mykola-mokhnach:
Also refactored existing intent-based helpers, so the intent name is not required anymore since `am` also allows to also provide the target name as a combination of `-a`/`-n` arguments

### DIFF
--- a/lib/commands/intent.js
+++ b/lib/commands/intent.js
@@ -10,16 +10,9 @@ const API_LEVEL_ANDROID_8 = 26;
 
 const commands = {};
 
-function requireOptions (opts, requiredKeys = []) {
-  const missingKeys = _.difference(requiredKeys, _.keys(opts));
-  if (!_.isEmpty(missingKeys)) {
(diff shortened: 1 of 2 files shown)


lib/commands/intent.js
```diff
@@ -176,54 +171,77 @@ commands.mobileStartActivity = async function mobileStartActivity (opts = {}) {
...
+
+/**
+ * Send a broadcast intent.
+ *
+ * @param {BroadcastOptions} opts
+ * @returns {string} The command output
+ * @throws {Error} If there was a failure while starting the activity
+ * or required options are missing
+ */
+commands.mobileBroadcast = async function mobileBroadcast (opts = {}) {
```
KazuCocoa, line 206: ah, good idea.
I remember I called this broadcast to automate push notification (via broadcast receiver in android app)

KazuCocoa (approved):

PR #708, opened 2021-09-24
feat: Implement fast unlock screen dismissal

mykola-mokhnach:
On android API 27+ it is possible to clear screen lock credential using adb, which is way faster and more stable in comparison to the current UIAutomator-based elements interaction approach.


lib/android-helpers.js
```diff
@@ -693,13 +695,20 @@ helpers.unlock = async function unlock (driver, adb, capabilities) {
...
 
+  const {
+    unlockType,
+    unlockKey,
+    unlockSuccessTimeout,
+  } = capabilities;
   logger.debug('Screen is locked, trying to unlock');
-  if (_.isUndefined(capabilities.unlockType)) {
+  if (unlockKey && await adb.isLockManagementSupported()) {
+    await unlocker.fastUnlock(unlockKey, unlockSuccessTimeout);
```
KazuCocoa, line 705: just a comment for now.

I see when `clear` the lock, the lock changes to "swipe".

I wonder some cases like VPN preference is set in the device under test.
Then, Android device requires to have secured lock like PIN. I think unlock by locksettings will fail then...
(I need to ensure this though, My memory could be old/wrong)

So, maybe we should consider a case the device must keep the current lock method.

mykola-mokhnach: yes, this is valid point. I'll add the necessary methods to restore the lock

KazuCocoa: mm, VPN did not require a secured lock... Let me see some other apps/examples for testing this.
I think some features required a secured lock...

KazuCocoa: I remember. For example, it was probably a customized security policy by a device admin.
Will check...

KazuCocoa: I cannot confirm the device admin stuff. Instead, I tried to lock by find my device. (not the same, but probably similar)

Then, the `adb shell locksettings clear` worked (succeeded to clear the lock) but it needed to off screen once, and it needed to on the screen.

1. calls `adb shell locksettings clear`
2. open the screen (still lock screen, but the cleared old password was wrong....)
3. turn off the screen
4. turn on the screen again (then, the lock was changed to "swipe")

I don't know exact behavior of lock by admin app, but I think it would be nice to have an option to keep "old" style for safe...

mykola-mokhnach: the implementation has been updated to restore the original lock credential after unlock operation is completed

KazuCocoa: It seems enough... but I still think if a device was managed by admin, then `locksettings` command fails like 
https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/core/java/com/android/server/locksettings/LockSettingsShellCommand.java;l=337-339?q=%22Lock%20credential%20cleared%22 ...
(I haven't read the code deeply yet, so this might not be related to such profile tho)

Unfortunately, I don't have such an environment, so I cannot test the case...

I heard some enterprise users who could not use devices that had no device policy as their test devices.
So, I hope such users get an opportunity to check this after our release. (Such enterprise users do not frequently update their test environment as well though...)

KazuCocoa (approved):
looks good to me (in patterns what I was able to test/consider...)

lib/android-helpers.js
KazuCocoa: ah.. or `COMMAND_GET_DISABLED` in https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/core/java/com/android/server/locksettings/LockSettingsShellCommand.java;l=132?q=%22Lock%20credential%20cleared%22 (after the credential check) may be an alternative as `isLockManagementSupported`...?
(just an idea)

mykola-mokhnach: These two commands have different purpose. COMMAND_GET_DISABLED returns true/false if we did or did not lock the device while `isLockManagementSupported` verifies if the corresponding adb functionality is available in general

mykola-mokhnach: > I heard some enterprise users who could not use devices that had no device policy as their test devices.
So, I hope such users get an opportunity to check this after our release. (Such enterprise users do not frequently update their test environment as well though...)

I'll add an additional capability to enforce UiAutomator-based unlock in the next PR

PR #709, opened 2021-09-26
feat: Make unlocking strategy configurable

mykola-mokhnach:
Just in case someone still needs to use the legacy approach. Also added a mobile helper for the stuff


KazuCocoa (approved):
Thank you!

PR #717, opened 2021-11-15
fix: Update parse memory data for Android 11(API 30)

YueChen-C:
getPerformanceData() get memoryinfo return : Error: Unable to parse memory data for Android 11(API 30) .

Android 11 (API 30) dumpsys meminfo adds "Rss Total" column before "Heap Size"

### DIFF
--- a/lib/commands/performance.js
+++ b/lib/commands/performance.js
@@ -3,19 +3,31 @@ import { retryInterval } from 'asyncbox';
 import log from '../logger';
 
 
-let commands = {}, helpers = {}, extensions = {};
+let commands = {}, helpers = {}, extensions = {}, memoryParsers = {};
 
 const NETWORK_KEYS = [['bucketStart', 'activeTime', 'rxBytes', 'rxPackets', 'txBytes', 'txPackets', 'operations', 'bucketDuration'], ['st', 'activeTime', 'rb', 'rp', 'tb', 'tp', 'op', 'bucketDuration']];
 const CPU_KEYS = ['user', 'kernel'];
 const BATTERY_KEYS = ['power'];
-const MEMORY_KEYS = ['totalPrivateDirty', 'nativePrivateDirty', 'dalvikPrivateDirty', 'eglPrivateDirty', 'glPrivateDirty', 'totalPss', 'nativePss', 'dalvikPss', 'eglPss', 'glPss', 'nativeHeapAllocatedSize', 'nativeHeapSize'];
-
+const MEMORY_KEYS = ['totalPrivateDirty', 'nativePrivateDirty', 'dalvikPrivateDirty', 'eglPrivateDirty', 'glPrivateDirty',
+  'totalPss', 'nativePss', 'dalvikPss', 'eglPss', 'glPss',
+  'nativeHeapAllocatedSize', 'nativeHeapSize',
+  'nativeRss', 'dalvikRss', 'totalRss'];
 const SUPPORTED_PERFORMANCE_DATA_TYPES = {
   cpuinfo: 'the amount of cpu by user and kernel process - cpu information for applications on real devices and simulators',
   memoryinfo: 'the amount of memory used by the process - memory information for applications on real devices and simulators',
   batteryinfo: 'the remaining battery power - battery power information for applications on real devices and simulators',
   networkinfo: 'the network statistics - network rx/tx information for applications on real devices and simulators'
 };
+const MEMORY_CONSTANT = {
+  NATIVE: 'Native',
+  DALVIK: 'Dalvik',
+  EGL: 'EGL',
+  GL: 'GL',
+  MTRACK: 'mtrack',
+  TOTAL: 'TOTAL',
+  HEAP: 'Heap'
+};
+
 
 const RETRY_PAUSE = 1000;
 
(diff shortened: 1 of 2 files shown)


lib/commands/performance.js
```diff
@@ -179,7 +211,7 @@ helpers.getMemoryInfo = async function getMemoryInfo (packageName, dataReadTimeo
 
     if (totalPrivateDirty && totalPrivateDirty !== 'nodex') {
       let headers = _.clone(MEMORY_KEYS);
-      let data = [totalPrivateDirty, nativePrivateDirty, dalvikPrivateDirty, eglPrivateDirty, glPrivateDirty, totalPss, nativePss, dalvikPss, eglPss, glPss, nativeHeapAllocatedSize, nativeHeapSize];
+      let data = [totalPrivateDirty, nativePrivateDirty, dalvikPrivateDirty, eglPrivateDirty, glPrivateDirty, totalPss, nativePss, dalvikPss, eglPss, glPss, nativeHeapAllocatedSize, nativeHeapSize, nativeRss, dalvikRss, totalRss];
```
mykola-mokhnach, line 214: Please split this long line

lib/commands/performance.js
```diff
@@ -154,6 +157,35 @@ helpers.getMemoryInfo = async function getMemoryInfo (packageName, dataReadTimeo
           totalPss = entries[1];
           totalPrivateDirty = entries[2];
         }
+      } else if (apilevel >= 30) {
+        let type = entries[0];
```
mykola-mokhnach, line 161: you could use array destruction for this purpose. For example:
```
const [type, subtype] = entries;
...
[,, nativePss, nativePrivateDirty,,, nativeRss, nativeHeapSize, nativeHeapAllocatedSize] = entries;
```

lib/commands/performance.js
```diff
@@ -8,7 +8,7 @@ let commands = {}, helpers = {}, extensions = {};
 const NETWORK_KEYS = [['bucketStart', 'activeTime', 'rxBytes', 'rxPackets', 'txBytes', 'txPackets', 'operations', 'bucketDuration'], ['st', 'activeTime', 'rb', 'rp', 'tb', 'tp', 'op', 'bucketDuration']];
 const CPU_KEYS = ['user', 'kernel'];
 const BATTERY_KEYS = ['power'];
-const MEMORY_KEYS = ['totalPrivateDirty', 'nativePrivateDirty', 'dalvikPrivateDirty', 'eglPrivateDirty', 'glPrivateDirty', 'totalPss', 'nativePss', 'dalvikPss', 'eglPss', 'glPss', 'nativeHeapAllocatedSize', 'nativeHeapSize'];
+const MEMORY_KEYS = ['totalPrivateDirty', 'nativePrivateDirty', 'dalvikPrivateDirty', 'eglPrivateDirty', 'glPrivateDirty', 'totalPss', 'nativePss', 'dalvikPss', 'eglPss', 'glPss', 'nativeHeapAllocatedSize', 'nativeHeapSize', 'nativeRss', 'dalvikRss', 'totalRss'];
```
mykola-mokhnach, line 11: Please avoid creating too long lines

YueChen-C:
I resubmitted the code and added a  API 30 unit case. please check it

test/unit/commands/performance-specs.js
```diff
@@ -158,10 +169,26 @@ describe('performance data', function () {
...
+          '120', '121', '127', // Rss        |native|dalvik|total
+        ],
+      ];
+      adb.getApiLevel.returns(30);
+      adb.shell.withArgs(shellArgs).returns(dumpsysDataAPI30);
+      (await driver.getMemoryInfo(PACKAGE_NAME)).should.be.deep
+        .equal(expectedResult30);
+      asyncbox.retryInterval.calledWith(RETRY_COUNT, RETRY_PAUSE).should.be.true;
+    });
+    it('should return memory info for 30>API>18', async function () {
```
mykola-mokhnach, line 191: the usual format for ranges is `18 < API < 30`

lib/commands/performance.js
```diff
@@ -113,73 +115,81 @@ helpers.getMemoryInfo = async function getMemoryInfo (packageName, dataReadTimeo
...
-          glPrivateDirty = entries[3];
+          [,, glPss, glPrivateDirty] = entries;
         } else if (type === 'TOTAL' && entries.length === 8) {
           // there are two totals, and we only want the full listing, which has 8 entries
-          totalPss = entries[1];
-          totalPrivateDirty = entries[2];
+          [, totalPss, totalPrivateDirty] = entries;
+        }
+      } else if (apilevel >= 30) {
+        const [type, subType] = entries;
```
mykola-mokhnach, line 155: it might be better readable if we move data parsers for different API versions to helper functions. Also, I would prefer to have magic strings like 'GL' or 'TOTAL' kept in constants.

lib/commands/performance.js
```diff
@@ -113,73 +115,81 @@ helpers.getMemoryInfo = async function getMemoryInfo (packageName, dataReadTimeo
...
-        nativePrivateDirty, nativePss, nativeHeapSize, nativeHeapAllocatedSize,
-        dalvikPrivateDirty, dalvikPss,
+    let totalPrivateDirty, totalPss, totalRss,
+        nativePrivateDirty, nativePss, nativeHeapSize, nativeHeapAllocatedSize, nativeRss,
+        dalvikPrivateDirty, dalvikPss, dalvikRss,
         eglPrivateDirty, eglPss,
         glPrivateDirty, glPss;
     let apilevel = await this.adb.getApiLevel();
     for (let line of data.split('\n')) {
       let entries = line.trim().split(' ').filter(Boolean);
```
mykola-mokhnach, line 125: does it make sense to rather split by `/\s+/`?

mykola-mokhnach: also filtering here by `Boolean` might cut zero values, which we probably don't want to happen

test/unit/commands/performance-specs.js
```diff
@@ -158,10 +169,26 @@ describe('performance data', function () {
       [
         '101', '102', '103', '104', '105', // private dirty total|native|dalvik|egl|gl
         '106', '107', '108', '109', '110', // pss           total|native|dalvik|egl|gl
-        '111', '112' // native        heap_alloc|heap_size
+        '111', '112', // native        heap_alloc|heap_size
+        undefined, undefined, undefined
       ],
     ];
-    it('should return memory info for API>18', async function () {
+    it('should return memory info for API>=30', async function () {
```
mykola-mokhnach, line 176: 👍

YueChen-C:
@mykola-mokhnach I submitted the code again. It looks better than last time

lib/commands/performance.js
```diff
@@ -61,6 +73,59 @@ commands.getPerformanceData = async function getPerformanceData (packageName, da
   return data;
 };
 
+// API level 30 and above
+//   ['<System Type>', '<Memory Type>', <pss total>, <private dirty>, <private clean>, <swapPss dirty>, <rss total>, <heap size>, <heap alloc>, <heap free>]
+memoryParsers.Api30AndAbove = function Api30AndAbove (entries, valDict) {
```
mykola-mokhnach, line 78: there is no need to create a separate namespace for this function. Simply define it being local to this module:

```
function parseMeminfoOutputForApi30Plus ...
```

lib/commands/performance.js
```diff
@@ -3,19 +3,31 @@ import { retryInterval } from 'asyncbox';
...
+  'totalPss', 'nativePss', 'dalvikPss', 'eglPss', 'glPss',
+  'nativeHeapAllocatedSize', 'nativeHeapSize',
+  'nativeRss', 'dalvikRss', 'totalRss'];
 const SUPPORTED_PERFORMANCE_DATA_TYPES = {
   cpuinfo: 'the amount of cpu by user and kernel process - cpu information for applications on real devices and simulators',
   memoryinfo: 'the amount of memory used by the process - memory information for applications on real devices and simulators',
   batteryinfo: 'the remaining battery power - battery power information for applications on real devices and simulators',
   networkinfo: 'the network statistics - network rx/tx information for applications on real devices and simulators'
 };
+const MEMORY_CONSTANT = {
```
mykola-mokhnach, line 21: I usually use Object.freeze on the value to make sure it won't be eventually modified

lib/commands/performance.js
```diff
@@ -61,6 +73,59 @@ commands.getPerformanceData = async function getPerformanceData (packageName, da
...
+  } else if (type === MEMORY_CONSTANT.TOTAL && entries.length === 9) {
+    // has 9 entries
+    [, valDict.totalPss, valDict.totalPrivateDirty,,, valDict.totalRss] = entries;
+  }
+};
+
+// API level between 18 and 30
+//   ['<System Type>', '<Memory Type>', <pss total>, <private dirty>, <private clean>, <swapPss dirty>, <heap size>, <heap alloc>, <heap free>]
+// except 'TOTAL', which skips the second type name
+memoryParsers.Api18And30 = function Api18And30 (entries, valDict) {
```
mykola-mokhnach, line 97: I'd rather return valDict than mutate it in-place

lib/commands/performance.js
mykola-mokhnach: Also, a general agreement is that method names should always contain verbs

lib/commands/performance.js
mykola-mokhnach: The current solution is also ok, but either the code or the method documentation must explicitly state that we intentionally mutate the incoming argument

YueChen-C:
@mykola-mokhnach I am trying to learn javascript, I want to fix this bug simply, but the result of the fix is not very satisfactory. So please  refactor and fix this code.

mykola-mokhnach:
> I am trying to learn javascript, I want to fix this bug simply, but the result of the fix is not very satisfactory. So please refactor and fix this code.

Unfortunately I don't have time this week. Probably, I can get back back to it closer to the weekend.
Altogether, nicely done as for a beginner.

YueChen-C:
I closed this PR and opened a new issue. Thank you for your reply

YueChen-C:
I closed this PR and opened a new issue. Thanks for your reply

PR #730, opened 2022-02-12
fix: Update the way system bars info is parsed

mykola-mokhnach:
It looks like dumpsys output is changing between different Android versions, so I've updated the parsing rules to make it work on Android 12+. Also refactored the code to make it a bit more easier to modify and update.

Addresses https://github.com/appium/appium/issues/16390

I've only tested on Android 11 and 12 emulators. I assume more testing is necessary on older system versions though.

### DIFF
--- a/lib/commands/system-bars.js
+++ b/lib/commands/system-bars.js
@@ -117,11 +117,13 @@ function parseWindows (lines) {
 }
 
 commands.getSystemBars = async function getSystemBars () {
-  const out = await this.adb.shell(['dumpsys', 'window', 'windows']);
-  if (!out) {
-    log.errorAndThrow('Did not get window manager output.');
+  let stdout;
+  try {
+    stdout = await this.adb.shell(['dumpsys', 'window', 'windows']);
+  } catch (e) {
+    throw new Error(`Cannot retrieve system bars details. Original error: ${e.message}`);
   }
-  return parseWindows(out);
+  return parseWindows(stdout);
 };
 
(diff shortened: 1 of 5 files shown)


jlipps (approved):
Code looks good if it works!

KazuCocoa:
Android 6 (tablet) was good, both landscape and portrait. Android 10 was as well.

KazuCocoa:
a unity game demo app also worked, each bar was invisible and visible in both cases.

lib/commands/system-bars.js
```diff
@@ -0,0 +1,129 @@
...
+    log.info(`No windows have been found whose title matches to ` +
+      `'${namePrefix}'. Assuming it is invisible. ` +
+      `Only the following windows are available: ${_.keys(windows)}`);
+    result[window] = _.cloneDeep(DEFAULT_WINDOW_PROPERTIES);
+  }
+  return result;
+}
+
+commands.getSystemBars = async function getSystemBars () {
+  const out = await this.adb.shell(['dumpsys', 'window', 'windows']);
```
KazuCocoa, line 120: need try/catch? e.g. when the device is disconnected at this time

KazuCocoa (approved):

lib/commands/system-bars.js
mykola-mokhnach: added

mykola-mokhnach:
Thanks for checking it @KazuCocoa

PR #732, opened 2022-03-02
fix: Read package name from web view info if not provided explicitly

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/16514


jlipps (approved):

KazuCocoa (approved):
ah... this case is needed when `appPackage` (and `app`) are not set. Then, a session starts a package via launching an activity api.
`autoLaunch: false` (that will be removed) and `app`/`appPackage` combination does not have this.

PR #745, opened 2022-05-18
fix: revert --pct-syskeys 0

KazuCocoa:
Revert the argument, `--pct-syskeys 0`.

`activateApp` does not use `am start` method for API level 24 with current implementation, so this revert also should not affect widely.

Btw, I remember we use `start-activity` instead of `start` for Android 8+, so this PR includes it.

### DIFF
--- a/lib/commands/app-management.js
+++ b/lib/commands/app-management.js
@@ -55,8 +55,9 @@ commands.queryAppState = async function queryAppState (appId) {
  */
 commands.activateApp = async function activateApp (appId) {
   this.log.debug(`Activating '${appId}'`);
+  const apiLevel = await this.adb.getApiLevel();
   // Fallback to Monkey in older APIs
-  if (await this.adb.getApiLevel() < 24) {
+  if (apiLevel < 24) {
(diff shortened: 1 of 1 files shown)


lib/commands/app-management.js
```diff
@@ -55,11 +55,13 @@ commands.queryAppState = async function queryAppState (appId) {
  */
 commands.activateApp = async function activateApp (appId) {
   this.log.debug(`Activating '${appId}'`);
+  const apiLevel = await this.adb.getApiLevel();
   // Fallback to Monkey in older APIs
-  if (await this.adb.getApiLevel() < 24) {
+  if (apiLevel < 24) {
+    // The monkey command could raise an issue as https://stackoverflow.com/questions/44860475/how-to-use-the-monkey-command-with-an-android-system-that-doesnt-have-physical
+    // but '--pct-syskeys 0' could cause another background process issue. https://github.com/appium/appium/issues/16941#issuecomment-1129837285
+    // We should not add it to avoid it.
```
mykola-mokhnach, line 63: this last sentence is redundant

mykola-mokhnach (approved):

PR #748, opened 2022-06-20
feat: Add mobile wrappers for media-projection-based screen recording

mykola-mokhnach:
Based on https://github.com/appium/io.appium.settings/pull/88

### DIFF
--- a/lib/commands/media-projection.js
+++ b/lib/commands/media-projection.js
@@ -4,6 +4,7 @@ import { util, fs, net, tempDir } from 'appium/support';
 import path from 'path';
 import B from 'bluebird';
 import { SETTINGS_HELPER_PKG_ID } from '../android-helpers';
+import moment from 'moment';
 
 
 const commands = {};
@@ -18,6 +19,7 @@ const RECORDING_ACTIVITY_NAME = `${SETTINGS_HELPER_PKG_ID}/io.appium.settings.Se
 const RECORDING_ACTION_START = `${SETTINGS_HELPER_PKG_ID}.recording.ACTION_START`;
 const RECORDING_ACTION_STOP = `${SETTINGS_HELPER_PKG_ID}.recording.ACTION_STOP`;
 const RECORDINGS_ROOT = `/storage/emulated/0/Android/data/${SETTINGS_HELPER_PKG_ID}/files`;
+const DEFAULT_FILENAME_FORMAT = 'YYYY-MM-DDTHH-mm-ss';
 
 
 async function uploadRecordedMedia (localFile, remotePath = null, uploadOptions = {}) {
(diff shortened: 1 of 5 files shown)


lib/commands/media-projection.js
```diff
@@ -0,0 +1,270 @@
...
+      '-a', RECORDING_ACTION_STOP,
+    ]);
+    try {
+      await waitForCondition(async () => !(await this.isRunning()), {
+        waitMs: RECORDING_STOP_TIMEOUT_MS,
+        intervalMs: 500,
+      });
+    } catch (e) {
+      throw new Error(
+        `The current media projection recording cannot stop within ${RECORDING_STOP_TIMEOUT_MS}ms timeout`
```
jlipps, line 140: `The attempt to stop the current media projection recording timed out after ${RECORDING_STOP_TIMEOUT_MS}ms`

lib/commands/media-projection.js
```diff
@@ -0,0 +1,270 @@
...
+ * If another recording has been already started then the command will exit silently.
+ * The previously recorded video file is deleted when a new recording session is started.
+ * Recording continues on driver session deletion if not stopped explicitly until the
+ * timeout happens.
+ *
+ * @param {?StartRecordingOptions} options Available options.
+ * @returns {boolean} True if a new recording has successfully started.
+ * @throws {Error} If recording has failed to start or is not supported on the device under test.
+ */
+commands.mobileStartMediaProjectionRecording = async function mobileStartMediaProjectionRecording (options = {}) {
```
jlipps, line 179: "mediaProjectionRecording" is an awful long string to have inside names everywhere :-(

jlipps (approved):

lib/commands/media-projection.js
mykola-mokhnach: I could use mpRecording, not sure if it would be readable enough though

jlipps: maybe for the internal code `mp` is fine, and keep the "mediaProjection" for the user-facing strings? not sure if that's easy to do.

it's not a huge deal, i just hate reading entire sentences as variable names :-)

lib/commands/media-projection.js
mykola-mokhnach: updated

lib/commands/media-projection.js
```diff
@@ -0,0 +1,271 @@
...
+}
+
+
+class MediaProjectionRecorder {
+  constructor (adb) {
+    this.adb = adb;
+  }
+
+  async isRunning () {
+    const stdout = await this.adb.shell(['dumpsys', 'activity', 'services']);
```
KazuCocoa, line 57: it seems like the result can be filtered by the package name/activity:
```
adb dumpsys activity services RECORDING_SERVICE_NAME
```

I haven't verified well, but potentially does this have slightly better performance rather than getting all services...?

Empty result:

```
ACTIVITY MANAGER SERVICES (dumpsys activity services)
  (nothing)
```

mykola-mokhnach: changed

KazuCocoa (approved):

PR #779, opened 2022-12-23
chore: Bump Appium Settings dependency

mykola-mokhnach:
Based on https://github.com/appium/io.appium.settings/pull/101


KazuCocoa:
My test will be late as my holidays plan, otherwise the change seems good

KazuCocoa (approved):

PR #782, opened 2023-01-11
fix: specify supported non-standard commands in newMethodMap

jlipps:
here is my first stab at this. i wrote a little script to basically generate the new method map we need. this script could be copied and changed in minor ways for every driver. it outputs a javascript object which i simply copied into the new method map declaration.

```js
const {METHOD_MAP} = require('@appium/base-driver');
const {AndroidDriver} = require('.');
const util = require('util');

const deprecatedCommandNames = {};
for (const [route, methodSpec] of Object.entries(METHOD_MAP)) {
  for (const [method, commandSpec] of Object.entries(methodSpec)) {
    if (commandSpec.deprecated && commandSpec.command) {
      deprecatedCommandNames[commandSpec.command] = {
        route,
        method,
        commandSpec,
      };
    }
  }
}

const newMethodMap = {};

for (const [command, details] of Object.entries(deprecatedCommandNames)) {
  const {route, method, commandSpec} = details;
  const instance = new AndroidDriver();
  if (instance[command]) {
    if (!newMethodMap[route]) {
      newMethodMap[route] = {};
    }
    newMethodMap[route][method] = {...commandSpec};
    delete newMethodMap[route][method].deprecated;
  }
}

console.log(util.inspect(newMethodMap, {depth: 5})); // eslint-disable-line no-console
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,6 @@ import * as commandIndex from './lib/commands/index';
 import * as webview from './lib/webview-helpers';
 import * as caps from './lib/desired-caps';
 
-
 const { AndroidDriver } = driver;
 const { helpers: webviewHelpers, NATIVE_WIN, WEBVIEW_WIN, WEBVIEW_BASE,
         CHROMIUM_WIN } = webview;
--- a/lib/method-map.js
+++ b/lib/method-map.js
@@ -1,0 +1,219 @@
+export const newMethodMap = /** @type {const} */ ({
+  '/session/:sessionId/timeouts/implicit_wait': {
+    POST: { command: 'implicitWait', payloadParams: { required: ['ms'] } }
+  },
+  '/session/:sessionId/ime/available_engines': { GET: { command: 'availableIMEEngines' } },
+  '/session/:sessionId/ime/active_engine': { GET: { command: 'getActiveIMEEngine' } },
+  '/session/:sessionId/ime/activated': { GET: { command: 'isIMEActivated' } },
+  '/session/:sessionId/ime/deactivate': { POST: { command: 'deactivateIMEEngine' } },
+  '/session/:sessionId/ime/activate': {
+    POST: {
+      command: 'activateIMEEngine',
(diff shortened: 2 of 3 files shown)


add-method-map.js
```diff
@@ -0,0 +1,32 @@
+const {METHOD_MAP} = require('@appium/base-driver');
```
jlipps, line 1: oops will delete this file before merging

jlipps:
we should probably also deprecate the commands in here that we plan to remove from the android drivers in the future. like the non-w3c compatible ones

index.js
```diff
@@ -17,7 +18,7 @@ const { helpers: androidHelpers, SETTINGS_HELPER_PKG_ID } = androidHelperIndex;
 
 export default AndroidDriver;
 export {
-  androidHelpers, androidCommands, AndroidDriver,
+  androidHelpers, androidCommands, AndroidDriver, newMethodMap,
```
mykola-mokhnach, line 21: do we need to export it here? I assume it should be enough to have the map being statically exposed by the driver class

jlipps: yeah, good point, i didn't know if it would work given how things are transpiled. but it seems to work.

mykola-mokhnach (approved):

lib/method-map.js
```diff
@@ -0,0 +1,219 @@
...
+      command: 'setValueImmediate',
+      payloadParams: { required: ['text'] }
+    }
+  },
+  '/session/:sessionId/appium/element/:elementId/replace_value': {
+    POST: {
+      command: 'replaceValue',
+      payloadParams: { required: ['text'] }
+    }
+  }
```
mykola-mokhnach, line 218: I assume we have forgotten about push/pull files/folder routes

jlipps: those seemed generically useful enough for any appium driver i decided not to deprecate them.

mykola-mokhnach: I thought we only keep endpoints that are in W3C spec

PR #785, opened 2023-02-22
fix: Increase interval for the helper app install

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/18233

I think for now it would be a simplest workaround we could apply there


rerorero (approved):
Thank you for so quickly fix!

PR #791, opened 2023-04-03
feat: Add mobile extensions to set and get connectivity

mykola-mokhnach:


KazuCocoa (approved):

lib/commands/network.js
```diff
@@ -37,6 +40,102 @@ commands.isWifiOn = async function isWifiOn () {
...
+    data: B.resolve(services.includes(DATA_KEY_NAME) ? undefined : this.adb.isDataOn()),
+    airplaneMode: B.resolve(
+      services.includes(AIRPLANE_MODE_KEY_NAME) ? undefined : this.adb.isAirplaneModeOn()
+    ),
+  };
+  await B.all(_.values(statePromises));
+  return _.fromPairs(services.map((k) => [k, statePromises[k].value()]));
+};
+
 commands.setNetworkConnection = async function setNetworkConnection (type) {
```
jlipps, line 139: should we rewrite the implementation of this method to use the new ADB methods?

mykola-mokhnach: maybe...
I don't like it much because it uses a non-standard endpoint and should be deprecated in favour of an extension command

jlipps: this endpoint is "standard" MJSONWP (https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#device-modes) and supported in all the clients i think. not sure if that changes your mind at all or if you intend to move every non-w3c endpoint to an extension command?

mykola-mokhnach: > if you intend to move every non-w3c endpoint to an extension command?

This is exactly what I wanted

jlipps: OK carry on then! We should follow a plan where the execute method is released, then clients are updated to simply wrap the execute method, then we can remove the additional route.

PR #813, opened 2023-05-24
chore: build with typescript instead of babel

boneskull:
prereq for getting it ready for automated docs.

PR #814, opened 2023-05-24
chore: add prettier, commitlint, swap husky

boneskull:
and reformat package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "appium-android-driver",
+  "version": "5.13.3",
   "description": "Android UiAutomator and Chrome support for Appium",
   "keywords": [
     "appium",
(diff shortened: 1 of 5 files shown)


boneskull:
Current dependencies on/for this PR:
* master
  * **PR #813** <a href="https://app.graphite.dev/github/pr/appium/appium-android-driver/813" target="_blank">Graphite</a> 
    * **PR #814** <a href="https://app.graphite.dev/github/pr/appium/appium-android-driver/814" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium-android-driver/814?utm_source=stack-comment).

package.json
```diff
@@ -116,14 +108,22 @@
...
     "lint-staged": "^13.0.3",
     "mocha": "^10.0.0",
-    "pre-commit": "^1.1.3",
+    "prettier": "^2.8.8",
     "rimraf": "^5.0.0",
     "semantic-release": "^20.0.2",
     "sinon": "^15.0.0",
     "ts-node": "^10.9.1",
-    "typescript": "^5.0.4",
+    "typescript": "^4.9.5",
```
mykola-mokhnach, line 119: why we don't support typescript v5?

package.json
```diff
@@ -116,14 +108,22 @@
...
     "xpath": "^0.x"
+  },
+  "peerDependencies": {
```
mykola-mokhnach, line 122: I was not quite sure about this. This is a library rather than standalone driver, so it makes more sense to import versioned server dependencies rather than use peer ones.

jlipps: hmm I think we still want to keep it as a peer since ultimately we want this package to still use whatever appium server the user is running, as with a driver.

jlipps (approved):

package.json
boneskull: updating this in a later PR

PR #819, opened 2023-06-14
feat: add all the types

boneskull:
**this PR requires both https://github.com/appium/appium-adb/pull/661 and https://github.com/appium/appium/pull/18721 to be published for the build to pass.**

_note: I did not type the entire codebase, but all of the commands are typed, as is the driver proper._

- moves all `@typedef`s which aren't aliases into `types.ts` and created many more
- each file now contains a single object literal, which is the mixin
- all methods in the mixin are defined in the object literal.  this allows us to use `ThisType` to avoid needing to use `@this` everywhere
- most everything remains js except the "find" mixin
- the mixin function and all mixin interfaces are defined in `mixins.ts`, in addition to the declaration merge with `AndroidDriver`
- the `coverage` mixin was removed entirely per a slack convo
- add missing `appWaitForLaunch` constraint
- update ESLint
- upgrade TS, use strict types


boneskull:
Current dependencies on/for this PR:
* master
  * **PR #813** <a href="https://app.graphite.dev/github/pr/appium/appium-android-driver/813" target="_blank">Graphite</a> 
    * **PR #814** <a href="https://app.graphite.dev/github/pr/appium/appium-android-driver/814" target="_blank">Graphite</a> 
      * **PR #819** <a href="https://app.graphite.dev/github/pr/appium/appium-android-driver/819" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium-android-driver/819?utm_source=stack-comment).

jlipps:
looks like appium-adb is published, and i'll run another publish of appium soon

boneskull:
@jlipps hold off on publishing appium; want to make sure xcuitest works ok with those changes

boneskull:
@jlipps actually, xcuitest is managed by renovate right, so I think it's okay to publish. was just worried loose deps would sink the ship

index.ts
```diff
@@ -0,0 +1,28 @@
+import {install} from 'source-map-support';
```
mykola-mokhnach, line 1: Didn't we already have the same issue in https://github.com/appium/node-teen_process/pull/253 ?

index.ts
```diff
@@ -0,0 +1,28 @@
...
+export {
+  CHROMIUM_WIN,
+  NATIVE_WIN,
+  WEBVIEW_BASE,
+  WEBVIEW_WIN,
+  default as webviewHelpers,
+} from './lib/helpers/webview';
+
+/**
+ * @privateRemarks `androidCommands` was previously exported as an object
```
mykola-mokhnach, line 20: we only need mixin commands there. Methods defined in the driver class itself must be ignored

lib/commands/app-management.js
```diff
@@ -1,337 +1,172 @@
...
+    if (/^error:/im.test(stdout)) {
+      throw new Error(`Cannot activate '${appId}'. Original error: ${stdout}`);
+    }
+  },
+
+  async mobileActivateApp(opts) {
+    const {appId} = requireArgs('appId', opts);
+    return await this.activateApp(appId);
+  },
+  async removeApp(appId, opts = {}) {
```
mykola-mokhnach, line 111: it is easier to read that if there is at least one line break between methods

mykola-mokhnach: also, why the docstrings are now gone?

lib/commands/context.js
```diff
@@ -1,479 +1,492 @@
...
+  },
+
+  isChromedriverContext(viewName) {
+    return _.includes(viewName, WEBVIEW_WIN) || viewName === CHROMIUM_WIN;
+  },
+
+  shouldDismissChromeWelcome() {
+    return (
+      !!this.opts.chromeOptions &&
+      _.isArray(this.opts.chromeOptions.args) &&
```
mykola-mokhnach, line 277: should we suppress the warning below?

lib/commands/mixins.ts
```diff
@@ -0,0 +1,962 @@
...
+   * {@link performMultiAction} is for reusing `performMultiAction` across android-drivers
+   * (like `appium-uiautomator2-driver`) and to avoid code duplication. Other
+   * android-drivers (like `appium-uiautomator2-driver`) need to override
+   * `doPerformMultiAction` to facilitate `performMultiAction`.
+   */
+  doPerformMultiAction(elementId: string, states: types.TouchState[]): Promise<void>;
+}
+
+declare module '../driver' {
+  interface AndroidDriver
```
mykola-mokhnach, line 924:
```
Interface 'AndroidDriver' cannot simultaneously extend types 'BaseDriver<{ readonly platformName: { readonly isString: true; readonly inclusionCaseInsensitive: readonly ["Android"]; readonly presence: true; }; readonly deviceName: { readonly isString: true; readonly presence: true; }; ... 93 more ...; readonly appWaitForLaunch: { ...; }; }, StringRecord, StringRecord, StringRe...' and 'FindMixin'.
```

lib/commands/actions.js
```diff
@@ -1,452 +1,348 @@
...
-  if (!util.hasValue(value)) {
-    this.log.errorAndThrow(`'value' argument is required`);
-  }
-  if (!this.isEmulator()) {
-    this.log.errorAndThrow('sensorSet method is only available for emulators');
-  }
-  await this.adb.sensorSet(sensorType, value);
-};
+const ActionsMixin = {
+  async keyevent(keycode, metastate) {
```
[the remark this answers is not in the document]
jlipps: ah. yeah that's it.

jlipps (commented):
this PR is impossible to really review. I think we should probably just merge/publish as a beta and then do a bunch of testing with uiauto2/espresso

index.ts
boneskull: ...what?

index.ts
boneskull: why?

lib/commands/app-management.js
boneskull: all docstrings are moved into the interfaces.

boneskull: did a search/replace: ` {2}\},\n {2}(?=.*?\()` with `  },\n\n  `  😄

lib/commands/context.js
boneskull: I don't see a warning?

lib/commands/mixins.ts
boneskull: see description

index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * containing _most_ of the methods within `AndroidDriver`.  It's unclear to me
+ * why we need to export this at all, or rather why it didn't include the
+ * prototype from `AndroidDriver` proper.  Hopefully, doing it this way won't
+ * break anything.
+ */
+export const androidCommands = {...AndroidDriver.prototype};
```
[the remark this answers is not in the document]
boneskull: I guess it remains to be seen.  I really have no idea why this is necessary at all

lib/commands/context.js
mykola-mokhnach: I assume it is a beta github feature.
It complains about `args` property is unknown for `chromeOptions`

index.ts
mykola-mokhnach: because AndroidDriver class itself contains stuff which is specific for the obsolete driver. It must not be used neither in uia2 nor in espresso

index.ts
mykola-mokhnach: @jlipps could you please explain better what the actual issue was?

boneskull:
@jlipps that sounds good. this will automatically be published if we merge it, so... maybe we should just publish it from the head branch?

mykola-mokhnach:
> @mykola-mokhnach why does the PR title have to be in conventional commit format? does it somehow affect the release or changelog?

yes, it affects the changelog and also the logic for automated versions bumping

jlipps:
>@jlipps that sounds good. this will automatically be published if we merge it, so... maybe we should just publish it from the head branch?

yeah I think so, maybe just publish a beta from this PR branch and we can work with it for a while. this repo doesn't see many commits so i don't anticipate much divergence. we could just freeze all work not related to this PR while we're working on it.

index.ts
boneskull: as far as I could tell the problem there was with `tsconfig.json`, not `source-map-support`

index.ts
boneskull: ok, I changed this to the way it was

boneskull: but also, ideally, uia2 and espresso would be subclasses of `AndroidDriver`, which renders it moot; they will get everything in `AndroidDriver` and would need to override any prop on the prototype they don't want

lib/commands/context.js
boneskull: oh..  yeah.  as I said, it won't work until we get those updates in.

index.ts
mykola-mokhnach: right now it is a collection of various mixins and helper methods which matter in this component. 
UIA2 and Espresso are too different to prefer subsclassing over composition.

boneskull:
as of right now, this needs new stuff in `@appium/types` to pass a build

jlipps:
does the release of appium I pushed a few hours ago do the trick?

boneskull:
@jlipps no it needs appium/appium#18817

boneskull:
assuming this build passes, I'll release another alpha off of it and deal w/ https://github.com/appium/appium-uiautomator2-driver/pull/635

jlipps:
👍

PR #824, opened 2023-06-16
chore: Make sharp to an optional dependency

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/18780

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -1,5 +1,5 @@
 import androidHelpers from '../android-helpers';
-import { fs, util, tempDir} from '@appium/support';
+import { fs, util, tempDir, imageUtil } from '@appium/support';
 import path from 'path';
 import B from 'bluebird';
 import { exec } from 'teen_process';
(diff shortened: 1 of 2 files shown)


lib/commands/actions.js
```diff
@@ -2,17 +2,24 @@ import androidHelpers from '../android-helpers';
...
 const dragStepsPerSec = 40;
+let _sharp;
 
 const commands = {};
 const helpers = {};
 const extensions = {};
 
+function requireSharp() {
+  if (!_sharp) {
+    _sharp = require('sharp');
```
jlipps, line 18: do we want to wrap this in error handling with a nice error message so the user knows what to do if this fails?

jlipps (approved):

lib/commands/actions.js
jlipps: or maybe this `requireSharp` function can even just be part of appium support? isn't sharp already an optional dep there?

jlipps (approved):

lib/commands/actions.js
mykola-mokhnach: it is already there

lib/commands/actions.js
```diff
@@ -409,7 +408,7 @@ helpers.getScreenshotDataWithAdbExecOut = async function getScreenshotDataWithAd
     throw new Error('Screenshot returned no data');
   }
 
-  return sharp(stdout);
+  return Buffer.from(stdout, 'ascii');
```
boneskull, line 411: this doesn't look right.  `stdout` is already a `Buffer`.

PR #850, opened 2023-09-14
ci: remove updating npm

KazuCocoa:


KazuCocoa:
to avoid node version error by the lastest npm

PR #860, opened 2023-10-05
feat: add waitForWebviewMs for mobile:getContexts

KazuCocoa:
To close https://github.com/appium/appium/issues/19251

https://github.com/appium/appium-uiautomator2-driver/pull/699 is the usage.


mykola-mokhnach (approved):
maybe we could also add some hint about usage of this option if no web views are found

PR #863, opened 2023-10-13
refactor!: Transform the package to a library

mykola-mokhnach:
BREAKING CHANGE: Removed the obsolete bootstrap-related stuff from the driver

I wanted to incorporate the recent version of this driver to Espresso, but this failed miserably. Now direct inheritance is used which brings a lot of unnecessary/deprecated stuff to UIA2 and Espresso with it.

It is time to put an end to that and delete the obsolete stuff from this package for good, so AndroidDriver becomes a simple collection of common procedures for UIA2 and Espresso drivers


KazuCocoa (approved):
lgtm, overall.

finally 👍

jlipps:
FINALLY!!!

PR #878, opened 2023-10-24
chore: apply auto commit hook lint only

KazuCocoa:
I'd like to avoid noisy lint fixes in https://github.com/appium/appium-android-driver/pull/860 so let me apply auto hook lint only as a separate pr. This is by commit hook lint change only

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -54,10 +54,7 @@ const ContextMixin = {
   },
 
   async getContexts() {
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
you can also remove prettier from the commit hook if you want to, we did the same for xcuitest driver
it could still be triggered manually when you want to.

mykola-mokhnach (approved):

PR #882, opened 2023-10-29
feat: Add 'hideKeyboard' capability

mykola-mokhnach:
Based on https://github.com/appium/io.appium.settings/pull/126

I would like to deprecate the `unicodeKeyboard` capability in the future as it is anyway not needed.
Also, the option has been removed from input APIs as it is not used there and is a leftover from the legacy bootstrap implementation.


KazuCocoa (approved):
sounds good about the unicodeKeyboard for me as well

PR #895, opened 2024-01-08
fix: io.appium.settings process check with isSettingsAppServiceRunningInForeground

KazuCocoa:
https://github.com/appium/appium-adb/pull/715 is required.

### DIFF
--- a/lib/helpers/android.ts
+++ b/lib/helpers/android.ts
@@ -809,48 +809,43 @@ const AndroidHelpers: AndroidHelpers = {
 
     // Reinstall would stop the settings helper process anyway, so
     // there is no need to continue if the application is still running
-    if (await adb.processExists(SETTINGS_HELPER_PKG_ID)) {
+    if (await adb.hasRunningSettingsAppForegroundService()) {
       logger.debug(
         `${SETTINGS_HELPER_PKG_ID} is already running. ` +
           `There is no need to reset its permissions.`
       );
-    } else {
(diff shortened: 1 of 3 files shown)


lib/helpers/android.ts
```diff
@@ -814,45 +814,51 @@ const AndroidHelpers: AndroidHelpers = {
         `${SETTINGS_HELPER_PKG_ID} is already running. ` +
           `There is no need to reset its permissions.`
       );
-      return;
```
mykola-mokhnach, line 817: I still don't understand why the `return` has been removed here. The below adjustments will only be done if the process is not running yes, so what is the point in restarting it afterwards? It might also take extra time on the session startup.

lib/helpers/android.ts
```diff
@@ -814,45 +814,51 @@ const AndroidHelpers: AndroidHelpers = {
         `${SETTINGS_HELPER_PKG_ID} is already running. ` +
```
KazuCocoa, line 814: `if (await adb.processExists(SETTINGS_HELPER_PKG_ID)) {` must be foreground process existing check as https://github.com/appium/appium-adb/pull/715

KazuCocoa:
Plus https://github.com/appium/appium-adb/pull/715

mykola-mokhnach:
Seems like unit tests need some updates

KazuCocoa:
Looks good for now

mykola-mokhnach (approved):

PR #896, opened 2024-01-09
feat: Add a helper method to simulate low-memory event

mykola-mokhnach:
Based on https://stackoverflow.com/questions/3656594/simulate-low-battery-low-memory-in-android


KazuCocoa (approved):
interesting, didn't know of this

PR #900, opened 2024-01-18
fix: stop parsing unreachable webview names

yaumu3:
## Intent

Currently, `getContexts` returns the webview names including the unreachable ones. This PR modifies the behavior not to return those.

## Confirmed

- ✅ unit test

With Android emulator,

1. Make an unreachable webview socket by the procedure described in https://issuetracker.google.com/issues/320819731 
2. Verify `getContexts` doesn't return the unreachable webview name

### DIFF
--- a/lib/commands/types.ts
+++ b/lib/commands/types.ts
@@ -191,7 +191,7 @@ export interface WebviewsMapping {
   /**
    * Webview pages list as it is retrieved by `/json/list` CDP endpoint
    */
-  pages: StringRecord[];
+  pages?: StringRecord[];
   /**
    * An actual webview name for switching context.
    *
--- a/lib/commands/types.ts
+++ b/lib/commands/types.ts
@@ -1081,7 +1081,7 @@ export type SwipeAction = [
   NonReleaseTouchAction,
   NonReleaseTouchAction,
   NonReleaseTouchAction,
-  ReleaseTouchAction
+  ReleaseTouchAction,
 ];
 
 export type TouchDragAction = [NonReleaseTouchAction, NonReleaseTouchAction, ReleaseTouchAction];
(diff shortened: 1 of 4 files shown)


lib/commands/types.ts
```diff
@@ -191,7 +191,7 @@ export interface WebviewsMapping {
   /**
    * Webview pages list as it is retrieved by `/json/list` CDP endpoint
    */
-  pages: StringRecord[];
+  pages?: StringRecord[];
```
yaumu3, line 194: Because of these lines, this can be `undefined`.

https://github.com/appium/appium-android-driver/blob/master/lib/helpers/webview.ts#L330-L335

lib/helpers/webview.ts
```diff
@@ -396,12 +396,18 @@ const WebviewHelpers: WebviewHelpers = {
 
     const result: string[] = [];
     for (const {webview, pages, proc, webviewName} of webviewsMapping) {
-      if (ensureWebviewsHavePages && pages?.length === 0) {
```
mykola-mokhnach, line 399: Simply change `pages?.length === 0` to `!pages?.length`

lib/commands/types.ts
mykola-mokhnach: Please add this comment into the docstring

yaumu3: How about something like this d0057995693acb4bd35aa124890593e06940f10d?

lib/helpers/webview.ts
yaumu3: Is it mentioned about the diff *after* the change `pages.length === 0`?

lib/helpers/webview.ts
```diff
@@ -396,12 +396,18 @@ const WebviewHelpers: WebviewHelpers = {
...
-        logger.info(
-          `Skipping the webview '${webview}' at '${proc}' ` +
-            `since it has reported having zero pages`,
-        );
-        continue;
+      if (ensureWebviewsHavePages) {
+        if (_.isUndefined(pages)) {
```
mykola-mokhnach, line 400: can pages be null or only undefined? would the `(!pages)` check be safer?

yaumu3: > can pages be null

It seems that this is the only place property `pages` is substituted.

https://github.com/appium/appium-android-driver/blob/7a9f6540ec010c73e08034fbbba73b8454bed5bf/lib/helpers/webview.ts#L333-L335

If the response body from `/json/list` is parsed as `null`, `pages` can be `null`.

---

IMO, such case is "no page retrieved" situation rather than "endpoint unreachable" one which I'm adding with this PR. Do you think it's rational to handle such case in the next conditional branch (`pages.length === 0`)?

mykola-mokhnach: if you say this is possible then the next branch is going to throw an exception as we cannot fetch any attribute from `null`.

yaumu3: > if you say this is possible then the next branch is going to throw an exception as we cannot fetch any attribute from null.

That's right, so I meant this change.

```diff
- if (pages.length === 0) {
+ if (_.isNull(pages) || pages.length === 0) {
```

mykola-mokhnach: it will work if you want to be explicit. May initial idea about `!pages?.length` was to make it simple and combine all cases into the one condition

yaumu3: Thank you so much for the elaboration. I've finally got the idea and fixed as  5e6b7ebe584eb0eb088cfd8ded6dda80f9399d24.

mykola-mokhnach (approved):

PR #901, opened 2024-01-18
refactor: Isolate base Android functionality

mykola-mokhnach:
BREAKING CHANGE: Removed androidHelpers export. Corresponding methods are now integrated into the driver
BREAKING CHANGE: Removed webviewHelpers export. Corresponding  methods are now integrated into the driver
BREAKING CHANGE: Removed redundant exports related to io.appium.settings.
BREAKING CHANGE: Removed deprecated APIs related to launchApp/closeApp/reset
BREAKING CHANGE: Removed  the unused `unlocker` property from the driver.

The mixin typesystem has been aligned. There are no explicit mixin definitions, which basically duplicate each method signature. Instead all types are now provided next to method definitions.

Improved grouping of different commands. Made it more explicit which of them are designed for export and which are just internal helpers

Improved helper methods design, so they don't need to retrieve driver-specific properties (like `log` or `adb`) via command line args, but use `this` context and explicit binding instead.

Marked legacy JWP touch-related methods as deprecated. They must be eventually replaced with proper W3C implementations.

Removed unused declarations for non-implemented methods.


mykola-mokhnach:
Unit tests obviously fail for now and need an update. Only the codebase has been updated to a buildable state.

KazuCocoa:
Took a look entirely. I don't have any objections to this

lib/commands/app-management.js
```diff
@@ -1,176 +1,497 @@
...
+ * @param {Omit<import('appium-adb').InstallOptions, 'appId'>} opts
+ * @returns {Promise<void>}
+ */
+export async function installApp(appPath, opts) {
+  const localPath = await this.helpers.configureApp(appPath, APP_EXTENSIONS);
+  await this.adb.install(localPath, opts);
+}
+
+/**
+ * @this {import('../driver').AndroidDriver}
```
jlipps, line 220: with so many `import`s it might be nice to just define the `AndroidDriver` type once below and refer to it elsewhere

jlipps (approved):
didn't review the code in detail but the big picture changes are OK to me. good to go when tests are in order.

lib/commands/app-management.js
mykola-mokhnach: done

PR #902, opened 2024-01-19
fix: set keepAlive false to parse webview pages correctly Node.js >=19

yaumu3:
Fix for https://github.com/appium/appium/issues/19678

### DIFF
--- a/lib/helpers/webview.ts
+++ b/lib/helpers/webview.ts
@@ -353,10 +353,10 @@ async function collectWebviewsDetails(
 }
 
 // https://chromedevtools.github.io/devtools-protocol/
-async function cdpList(host: string, port: number): Promise<object[]> {
+async function cdpGetRequest(host: string, port: number, endpoint: string): Promise<object[]> {
   return (
     await axios({
-      url: `http://${host}:${port}/json/list`,
+      url: `http://${host}:${port}${endpoint}`,
       timeout: CDP_REQ_TIMEOUT,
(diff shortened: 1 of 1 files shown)


lib/helpers/webview.ts
```diff
@@ -323,8 +324,8 @@ async function collectWebviewsDetails(
   for (const item of webviewsMapping) {
     detailCollectors.push(
       (async () => {
-        let port: number|undefined;
-        let host: string|undefined;
+        let port: number | undefined;
+        let host: string | undefined;
```
KazuCocoa, line 328: by auto-format or something?

KazuCocoa (approved):

lib/helpers/webview.ts
yaumu3: Yes, by the pre-commit hook triggered via husky.

lib/helpers/webview.ts
```diff
@@ -357,6 +358,11 @@ async function cdpList(host: string, port: number): Promise<object[]> {
     await axios({
```
mykola-mokhnach, line 358: this API can be extracted into a separate function call to avoid extra duplication

yaumu3: Thank you! Extracted as 3e1ea66eee409e56519e5d1f885ec048dc66c1fe

KazuCocoa (approved):

PR #913, opened 2024-02-05
Update activity constants for launching Chrome

ElCamino72:
Based on the findings presented in this article: https://automationchronicles.com/error-when-opening-chrome-on-android-13-via-adb/

### DIFF
--- a/lib/commands/context/helpers.js
+++ b/lib/commands/context/helpers.js
@@ -15,7 +15,7 @@ import {toDetailsCacheKey, getWebviewDetails, WEBVIEWS_DETAILS_CACHE} from './ca
 export const CHROME_BROWSER_PACKAGE_ACTIVITY = /** @type {const} */ ({
   chrome: {
     pkg: 'com.android.chrome',
-    activity: 'com.google.android.apps.chrome.Main',
+    activity: 'com.google.android.apps.chrome.IntentDispatcher',
   },
   chromium: {
     pkg: 'org.chromium.chrome.shell',
@@ -23,7 +23,7 @@ export const CHROME_BROWSER_PACKAGE_ACTIVITY = /** @type {const} */ ({
   },
   chromebeta: {
     pkg: 'com.chrome.beta',
-    activity: 'com.google.android.apps.chrome.Main',
+    activity: 'com.google.android.apps.chrome.IntentDispatcher',
   },
   browser: {
     pkg: 'com.android.browser',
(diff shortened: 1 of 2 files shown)


lib/commands/context/helpers.js
```diff
@@ -15,31 +15,31 @@ import {toDetailsCacheKey, getWebviewDetails, WEBVIEWS_DETAILS_CACHE} from './ca
 export const CHROME_BROWSER_PACKAGE_ACTIVITY = /** @type {const} */ ({
   chrome: {
     pkg: 'com.android.chrome',
-    activity: 'com.google.android.apps.chrome.Main',
+    activity: 'com.google.android.apps.chrome.IntentDispatcher',
```
mykola-mokhnach, line 18: not sure if it is a good idea to change records. Perhaps, it makes sense to just add new ones

mykola-mokhnach: maybe it also makes sense to autodetect the activity for the given package name using `cmd package resolve-activity --brief <pkg_name>` (only works on Android 7+)?

KazuCocoa:
https://github.com/appium/appium/issues/17492 is a relevant conversation we had before. 

The method where uses this const in uia2 just sets the appPackage info as a return value so maybe this change is not necessary as Appium's usage, but perhaps it makes sense to use the resolve-activity (personally, but not sure) to show what current active activity is as the response. (?)

Btw, in the past, a new session  with below caps succeeded in launching chrome process for Android 13 (if my memory was correct as https://github.com/appium/appium/issues/17492#issuecomment-1314708723 ) as well. It could depend on chrome/chromedriver version as well, but perpahs all activity do not need to be `IntentDispatcher`. Some, or they could depend on the chrome version/OS version etc

```
"appPackage": "com.android.chrome",
"appActivity": "com.google.android.apps.chrome.Main"
```

lib/commands/context/helpers.js
mykola-mokhnach: Maybe just try if https://github.com/appium/appium-uiautomator2-driver/pull/730 works for you

michapringle: 👍 Sure, we will try it.

mykola-mokhnach: you don't need to wait for a release. You can simply checkout my branch locally and then install the driver from the local filesystem using `--source=local` CLI argument.

michapringle: ok

PR #930, opened 2024-04-08
feat: Add a possibility to set device timezone

mykola-mokhnach:
https://sclarkso.com/tech/2021/10/05/quest-2-timezone-change-without-factory-reset.html

### DIFF
--- a/lib/commands/time.js
+++ b/lib/commands/time.js
@@ -44,6 +44,11 @@ export async function adjustTimeZone(zoneName) {
     );
   }
   this.log.info(`Setting the device time zone to '${zoneName}'`);
+  // The magic number '3' depends on the actual ordering of methods in
+  // the IAlarmManager interface and might be a subject of change between
+  // different Android API versions.
+  // See, for example,
+  // https://cs.android.com/android/platform/superproject/+/master:frameworks/base/apex/jobscheduler/framework/java/android/app/IAlarmManager.aidl;l=1?q=IAlarmManager
   await this.adb.shell(['service', 'call', 'alarm', '3', 's16', zoneName]);
 }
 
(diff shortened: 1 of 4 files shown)


lib/commands/time.js
```diff
@@ -23,14 +23,31 @@ export async function getDeviceTime(format = MOMENT_FORMAT_ISO8601) {
...
+ */
+export async function adjustTimeZone(zoneName) {
+  if (!moment.tz.names().includes(zoneName)) {
+    throw new Error(
+      `The provided time zone identifier '${zoneName}' is not known. ` +
+      `Please choose a valid TZ identifier from https://en.wikipedia.org/wiki/List_of_tz_database_time_zones`
+    );
+  }
+  this.log.info(`Setting the device time zone to '${zoneName}'`);
+  await this.adb.shell(['service', 'call', 'alarm', '3', 's16', zoneName]);
```
KazuCocoa, line 47: The `3` depends on the number of the method in `IAlarmManager.aidl` so the number could depends on Android OS versions. So I think it would be nice to give API version as an argument in case we need to modify the number.

https://cs.android.com/android/platform/superproject/+/master:frameworks/base/apex/jobscheduler/framework/java/android/app/IAlarmManager.aidl;l=1?q=IAlarmManager


I haven't checked any cases, but for example when https://android.googlesource.com/platform/frameworks/base/+/f76a50c/core/java/android/app/IAlarmManager.aidl case, the number might be `5`. (randomly googled)

KazuCocoa: never mined for the api level as an argument. we could call `const apiLevel = await this.adb.getApiLevel()`.

Maybe adding comment the `3` could be another number by depending on the aidl for the API level?

KazuCocoa (approved):
Lg for the change itself

lib/commands/time.js
mykola-mokhnach: added a comment

PR #938, opened 2024-06-07
fix: Add proper timestamps to server logs

mykola-mokhnach:
Similar to https://github.com/appium/appium-xcuitest-driver/pull/2406/

### DIFF
--- a/lib/commands/log.js
+++ b/lib/commands/log.js
@@ -37,12 +37,10 @@ export const supportedLogTypes = {
      */
     getter: (self) => {
       self.ensureFeatureEnabled(GET_SERVER_LOGS_FEATURE);
-      const timestamp = Date.now();
       return self.log
         .unwrap()
-        .record.map((x) =>
-          toLogRecord(
-            timestamp,
+        .record.map((x) => toLogRecord(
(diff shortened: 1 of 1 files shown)


lib/commands/log.js
```diff
@@ -37,12 +37,10 @@ export const supportedLogTypes = {
...
     getter: (self) => {
       self.ensureFeatureEnabled(GET_SERVER_LOGS_FEATURE);
-      const timestamp = Date.now();
       return self.log
         .unwrap()
-        .record.map((x) =>
-          toLogRecord(
-            timestamp,
+        .record.map((x) => toLogRecord(
+            /** @type {any} */ (x).timestamp ?? Date.now(),
```
KazuCocoa, line 43: btw, for old deps, the `timestamp` could be undefined so `Date.now()`  will be used, correct? (just a confirmation)

KazuCocoa (approved):

lib/commands/log.js
mykola-mokhnach: yes, this is done in order to keep the compatibility with older server versions

PR #955, opened 2024-09-05
feat: Emit logs over BiDi socket

mykola-mokhnach:
Similar to https://github.com/appium/appium-xcuitest-driver/pull/2458/

### DIFF
--- a/lib/commands/log.js
+++ b/lib/commands/log.js
@@ -1,5 +1,4 @@
 import {DEFAULT_WS_PATHNAME_PREFIX, BaseDriver} from 'appium/driver';
-import {util} from 'appium/support';
 import _ from 'lodash';
 import os from 'node:os';
 import WebSocket from 'ws';
@@ -179,7 +178,7 @@ export function assignBiDiLogListener (logEmitter, properties) {
         type,
         level: finalEntry.level,
         source: {
-          realm: util.uuidV4(),
+          realm: '',
         },
         text: finalEntry.message,
         timestamp: finalEntry.timestamp,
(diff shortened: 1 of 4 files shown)


lib/commands/log.js
```diff
@@ -155,6 +154,42 @@ export async function getLogTypes() {
...
+  const listener = (/** @type {import('../utils').LogEntry} */ logEntry) => {
+    const finalEntry = entryTransformer ? entryTransformer(logEntry) : logEntry;
+    this.eventEmitter.emit(BIDI_EVENT_NAME, {
+      context,
+      method: 'log.entryAdded',
+      params: {
+        type,
+        level: finalEntry.level,
+        source: {
+          realm: util.uuidV4(),
```
jlipps, line 182: this will be recalculated on every log emission?

jlipps: guess i'm not quite sure of the meaning of this `realm` attribute

jlipps (approved):

lib/commands/log.js
mykola-mokhnach: I've just copied the structure from the [spec](https://w3c.github.io/webdriver-bidi/#types-log-logentry). Maybe I understand it wrong

jlipps: yikes I have no idea what a realm is I guess. Could we just define it once in `assignBiDiLogListener` rather than on every log emission?

KazuCocoa (approved):

lib/commands/log.js
mykola-mokhnach: I've just set it to an empty string for now

PR #964, opened 2024-11-09
chore: use activateApp from adb

KazuCocoa:
It looks like... the same activateApp exists in https://github.com/appium/appium-adb/blob/74fb4f0a15f8e1c1db610aa633acf856877de209/lib/tools/apk-utils.js#L1180 so we can call the adb's one instead.

### DIFF
--- a/lib/commands/app-management.js
+++ b/lib/commands/app-management.js
@@ -6,7 +6,6 @@ import {EOL} from 'node:os';
 import B from 'bluebird';
 
 const APP_EXTENSIONS = ['.apk', '.apks'];
-const RESOLVER_ACTIVITY_NAME = 'android/com.android.internal.app.ResolverActivity';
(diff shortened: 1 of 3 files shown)


lib/commands/app-management.js
```diff
@@ -84,57 +84,7 @@ export async function mobileQueryAppState(opts) {
...
-      );
-    }
-    if (output.includes('monkey aborted')) {
-      this.log.errorAndThrow(`Cannot activate '${appId}'. Are you sure it is installed?`);
-    }
-    return;
-  }
-
-  let activityName = await this.adb.resolveLaunchableActivity(appId);
-  if (activityName === RESOLVER_ACTIVITY_NAME) {
```
mykola-mokhnach, line 110: Please also remove the unused constant

mykola-mokhnach (approved):

PR #988, opened 2025-03-17
fix: add backgroundApp

KazuCocoa:
Espresso driver missed this background app right now.

https://github.com/appium/appium-espresso-driver?tab=readme-ov-file#mobile-backgroundapp

Past implementation in espresso driver is the same as uia2's one, so this android driver can have it and both drivers refer to this.

### DIFF
--- a/lib/driver.ts
+++ b/lib/driver.ts
@@ -413,6 +413,7 @@ class AndroidDriver
   activateApp = activateApp;
   queryAppState = queryAppState;
   isAppInstalled = isAppInstalled;
+  mobileBackgroundApp = background;
 
   mobileGetUiMode = mobileGetUiMode;
   mobileSetUiMode = mobileSetUiMode;
--- a/lib/execute-method-map.ts
+++ b/lib/execute-method-map.ts
@@ -174,7 +174,12 @@ export const executeMethodMap = {
       required: ['appId'],
     },
   },
-
+  'mobile: backgroundApp': {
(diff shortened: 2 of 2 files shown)


lib/driver.ts
```diff
@@ -413,6 +413,7 @@ class AndroidDriver
   activateApp = activateApp;
   queryAppState = queryAppState;
   isAppInstalled = isAppInstalled;
+  mobileBackgroundApp = background;
```
mykola-mokhnach, line 416: for espresso driver the removal of backgroundApp was done deliberately https://github.com/appium/appium-espresso-driver/blob/6634b81d0014774039fd68a87cc141a4f67137ba/CHANGELOG.md?plain=1#L59

The reason for it is that espresso server, being integrated into the app under test, stops working if the app is not in foreground, so it makes no sense to have this API available there

mykola-mokhnach, line 416: `mobileBackgroundApp` and `background` definitions are not the same. `mobileBackgroundApp` sets the background argument to -1 by default, which changes the default behavior

KazuCocoa: That's right (default one)

KazuCocoa: added as a new mobile command: https://github.com/appium/appium-android-driver/pull/988/commits/2a847cfdba60b0a05a2d6270d3ed07269bc45df6

mykola-mokhnach (approved):

PR #994, opened 2025-04-16
feat: Add /session/:sessionId/network_connection to the method map

mykola-mokhnach:
I'd like to deprecate this endpoint in the base driver as it is anyway only used in Android

### DIFF
--- a/lib/method-map.js
+++ b/lib/method-map.js
@@ -164,4 +164,11 @@ export const newMethodMap = /** @type {const} */ ({
       payloadParams: {required: ['text']},
     },
   },
+  '/session/:sessionId/network_connection': {
+    GET: {command: 'getNetworkConnection'},
+    POST: {
+      command: 'setNetworkConnection',
+      payloadParams: {unwrap: 'parameters', required: ['type']},
+    },
+  },
 });


eglitise:
I think this is a good opportunity to change the endpoint URL so that it fits the W3C spec.
How about `/session/:sessionId/appium/networkconnection`, or `/session/:sessionId/appium/device/network_connection`?

mykola-mokhnach:
> I think this is a good opportunity to change the endpoint URL so that it fits the W3C spec. How about `/session/:sessionId/appium/networkconnection`, or `/session/:sessionId/appium/device/network_connection`?

We might do it, but these endpoints are hardcoded in multiple clients and they all would have to be updated. It might be an unnecessary extra work plus older clients might stop working completely, which would create inconveniences for customers

eglitise:
Sure, but eventually we would want to move everything only to W3C protocol-compliant endpoints. I assume that with Appium 3, we'll have major versions for all drivers anyway, so this endpoint change would fit right in with the removal of the base-driver endpoints.

mykola-mokhnach:
> Sure, but eventually we would want to move everything only to W3C protocol-compliant endpoints. I assume that with Appium 3, we'll have major versions for all drivers anyway, so this endpoint change would fit right in with the removal of the base-driver endpoints.

the idea is that removed endpoints are "transparent" to clients, which use either alternative "mobile" extensions or already have the same (old JWP) endpoint mapped "locally".  Such renaming might be a part of bigger initiative, where we'd like to put all non-standard endpoints under "appium" or any other subpath. Nevertheless in order to conduct such migration transparently without creating major inconveniences for users it would be necessary to:

- Add the new endpoint to the driver without removing the old one
- Mark the old one as deprecated
- Update all clients by adding support to both API endpoints, preferring the new one
- Let it be like that for a while until most of users migrate to new clients
- Remove old APIs from the driver (major version bump)
- Remove old endpoints from clients (major version bump)

eglitise:
Sure, that makes sense. Then for this PR, how about defining both the old endpoint (marking it as deprecated) and a new W3C compliant one, and point both to the same methods (`getNetworkConnection`/`setNetworkConnection`)?

mykola-mokhnach:
not sure if it makes sense to do it for a single endpoint. Like I said, this driver has much much more endpoints, that could be updated in the same way

eglitise (approved):

KazuCocoa:
`mobile: getConnectivity` / `mobile: setConnectivity` are alternative for these methods although this aims to keep these endpoints as "deprecated" for future removal, correct...?

KazuCocoa (approved):

mykola-mokhnach:
> `mobile: getConnectivity` / `mobile: setConnectivity` are alternative for these methods although this aims to keep these endpoints as "deprecated" for future removal, correct...?

Yes, this is the idea. Nevertheless we cannot simply remove old ones without ensuring that all clients are up to date.

PR #1005, opened 2025-08-06
fix: Make sure session id is set before cleaning up socket handlers

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/21464


KazuCocoa (approved):

KazuCocoa:
Let me merge this and bump uia2 to test this out

PR #1071, opened 2026-06-30
feat: grant runtime permissions to the Chrome package (chromedriverGrantPermissions)

AsijitM:
## Problem

On Android, Chrome interrupts an automated web session with a **native runtime-permission dialog** the first time a page uses geolocation, the camera (`getUserMedia`), or file download/upload. That dialog is a native OS prompt, not web content, so it can't be dismissed via the automation session and the session stalls.

## Fix

Add a `chromedriverGrantPermissions` capability. When set, `setupNewChromedriver` grants **all** runtime permissions to the **resolved** Chrome package (via `adb.grantAllPermissions`) **before** Chrome is launched, so they take effect for the session without a restart. Because the package is resolved by the driver, it covers the beta/dev/canary variants in `KNOWN_CHROME_PACKAGE_NAMES`, not just `com.android.chrome`.

The grant lives in appium-android-driver (which already owns app-permission handling and resolves the Chrome package) rather than in appium-chromedriver — per the review discussion on appium/appium-chromedriver#587, keeping chromedriver focused on browser interactions and Android-specific setup here. That chromedriver PR is now closed in favour of this approach; this PR has **no dependency** on it.

Since the grant is opt-in/explicit, a failure propagates rather than being swallowed.

## How tested

- New unit tests in `test/unit/commands/context-specs.ts`: grants to the resolved package when the cap is set; no grant when it isn't.
- `npm run build`, `npm run lint` pass; the context unit specs pass (44 passing).

## How it fixes the issue

With the permissions pre-granted before Chrome launches, the native permission dialog never appears, so geolocation/camera/file-access web tests run without manual interaction.

### DIFF
--- a/lib/commands/context/helpers.ts
+++ b/lib/commands/context/helpers.ts
@@ -320,7 +320,6 @@ export async function setupNewChromedriver(
           `resolved from the chromedriver capabilities, so runtime permissions cannot be granted`,
       );
     }
-    this.log.info(`Granting all runtime permissions to '${chromePkg}'`);
     await this.adb.grantAllPermissions(chromePkg);
   }
   const sessionCaps = await chromedriver.start(caps);
--- a/lib/commands/context/helpers.ts
+++ b/lib/commands/context/helpers.ts
@@ -274,10 +274,6 @@ export async function setupNewChromedriver(
     bundleId: (opts as any).chromeBundleId,
     useSystemExecutable: opts.chromedriverUseSystemExecutable,
     disableBuildCheck: opts.chromedriverDisableBuildCheck,
-    // Pre-grant Android runtime permissions to Chrome so the session is not blocked by a
-    // native permission dialog. Requires appium-chromedriver with the `grantPermissions`
-    // option (https://github.com/appium/appium-chromedriver/pull/587).
-    grantPermissions: opts.chromedriverGrantPermissions,
     details: details as any,
     isAutodownloadEnabled: isChromedriverAutodownloadEnabled.bind(this)(),
   };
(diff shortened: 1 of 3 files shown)


lib/commands/context/helpers.ts
```diff
@@ -309,6 +309,21 @@ export async function setupNewChromedriver(
...
+  // Optionally pre-grant all runtime permissions to the resolved Chrome package before Chrome
+  // is launched, so the session is not interrupted by a native permission dialog (geolocation,
+  // camera, file access, ...). Because the grant is requested explicitly, failures propagate.
+  if (opts.chromedriverGrantPermissions) {
+    const chromePkg = caps.chromeOptions?.androidPackage;
+    if (chromePkg) {
+      this.log.info(`Granting all runtime permissions to '${chromePkg}'`);
+      await this.adb.grantAllPermissions(chromePkg);
+    } else {
+      this.log.warn(
```
mykola-mokhnach: this should be an exception

mykola-mokhnach:
Please do not forget to document the new capability in UIA2 driver capability docs when this PR is merged

lib/commands/context/helpers.ts
AsijitM: Done in ceb01d0 — it now throws an `Error` when `chromedriverGrantPermissions` is set but the Chrome package name cannot be resolved from the chromedriver caps, instead of warning and silently skipping. Since the grant was explicitly requested, failing to honour it should propagate. Added a unit test for the throw path.

AsijitM:
Will do — I will add `chromedriverGrantPermissions` to the capability docs in the `appium-uiautomator2-driver` repo (and the espresso driver, since it inherits this) once this is merged and released. Thanks!
Raised here: https://github.com/appium/appium-uiautomator2-driver/pull/1018

lib/commands/context/helpers.ts
```diff
@@ -309,6 +309,20 @@ export async function setupNewChromedriver(
...
+  // camera, file access, ...). Because the grant is requested explicitly, failures propagate.
+  if (opts.chromedriverGrantPermissions) {
+    const chromePkg = caps.chromeOptions?.androidPackage;
+    if (!chromePkg) {
+      throw new Error(
+        `'chromedriverGrantPermissions' is enabled but the Chrome package name could not be ` +
+          `resolved from the chromedriver capabilities, so runtime permissions cannot be granted`,
+      );
+    }
+    this.log.info(`Granting all runtime permissions to '${chromePkg}'`);
```
mykola-mokhnach: this log line is redundant, already present in adb

AsijitM: Good catch — removed in 9577fa7. `adb.grantAllPermissions`/`grantPermissions` already log the package and permissions, so this line was just duplicating that.

mykola-mokhnach (approved):

KazuCocoa (approved):

